PR #17, opened 2021-04-24
feat: create project from existing contract

agraebe:
Adds a new command to generate a fresh project from an existing, deployed contract on the mainnet:

```
clarinet fork my-project SP162D87CY84QVVCMJKNKGHC7GGXFGA0TAR9D0XJW.puzzled-gold-stork
```


agraebe:
As discussed, this will be implemented differently

PR #20, opened 2021-05-05
fix: update cli command to add a new contract

michaelgold:
existing command `clarinet generate contract bbtc` results in a fatal error.


lgalabru (approved):
Good catch thank you @michaelgold!

PR #27, opened 2021-05-23
Feat/mine empty block until

### LINKED ISSUE #26: Feature: allow advance chain tip to secyfic block number.
As a user I would like to have a function that will let me advance chain tip not by X blocks like we can do with `chain.mineEmptyBlock(X)`, but by specifying the exact target block. That way I won't have to calculate exactly how many blocks I need to mine, to land in desired block.

I think only typescript code needs to be modified to achieve this.
We can add `block_height` field to `Chain` class and update it  to the most recent block height every single time we call `mineEmptyBlock()` or `mineBlock()`. And then add new function `mineEmptyBlocksTill(block_height: number)` that would calculate how many blocks we have to mine and calls `mineEmptyBlock()` in background.

LNow:
This PR introduced new public field to `Chain` class, that gives developers access to current chain tip height and new method that allows to advance chain up until specific block height. 

close #26

### DIFF
--- a/deno/index.ts
+++ b/deno/index.ts
@@ -104,6 +104,7 @@ export interface AssetsMaps {
 
 export class Chain {
   sessionId: number;
+  blockHeight: number = 1;
 
   constructor(sessionId: number) {
     this.sessionId = sessionId;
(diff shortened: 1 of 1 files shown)


lgalabru:
LGTM, thank you @LNow!

PR #57, opened 2021-06-25
add prereqs to readme for build from source

### LINKED ISSUE #7: Installing Clarinet on Linux
This is amazing, thank you!

When installing on Linux Mint (based on Ubuntu), I ran into a a few errors when I ran `cargo install clarinet`.

The main one was `linker 'cc' not found`.

To fix, I installed the same prereqs I used when building a miner, although I think `build-essential` may be the only one I needed.

https://github.com/AbsorbingChaos/stx-miner-script/blob/master/config-miner-krypton.sh#L51

Would it be worth adding a note to the readme?

whoabuddy:
This fixes #7 by adding a note about installing Rust and the required packages for Debian/Ubuntu Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,18 @@ provided in case Cargo is not available to you.
 
 ### Install from source using Cargo
 
+#### Prerequisites
+
+[Install Rust](https://www.rust-lang.org/tools/install) for access to `cargo`, the Rust package manager.
+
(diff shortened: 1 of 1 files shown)


lgalabru (approved):
LGTM, thank you so much @whoabuddy!

PR #65, opened 2021-06-29
ci: add new workflow to update develop branch on release

CharlieC3:
On release, the semantic-release plugin will create a commit to update the CHANGELOG and some other files, and then tag that commit with the new version. If someone then pushes a commit to the `develop` branch, semantic-release will create a new release candidate using the _previous_ version tag.

This is because when analyzing the commits in the `develop` branch, it does not see the recently-generated tag available on the `develop` branch. Ideally, a PR would be opened against the develop branch automatically after a release is performed to port over the new commit and tag from the `main` branch.

This would allow any further new commits to the `develop` branch to generate a release candidate with the most up-to-date tag version from `main`.

### DIFF
--- a/.github/workflows/open-develop-pr.yml
+++ b/.github/workflows/open-develop-pr.yml
@@ -7,7 +7,7 @@ name: Open Develop PR
 on:
   release:
     types:
-      - published
+      - released
   workflow_dispatch:
 
 jobs:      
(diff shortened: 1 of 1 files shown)


lgalabru (approved):
LGTM, thank you @CharlieC3!

PR #75, opened 2021-07-09
feat: add nft tutorial example code

pgray-hiro:
This PR adds the NFT tutorial code developed by Alex to the repo.

### DIFF
--- a/examples/nft/Clarinet.toml
+++ b/examples/nft/Clarinet.toml
@@ -1,0 +1,10 @@
+[project]
+name = "clarity-nft"
+requirements = []
+[contracts.my-nft]
+path = "contracts/my-nft.clar"
+depends_on = ["nft-trait"]
+
+[contracts.nft-trait]
+path = "contracts/nft-trait.clar"
+depends_on = []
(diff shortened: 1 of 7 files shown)


MarvinJanssen:
Looks fine but some of the structures in `my-nft.clar` seem a little over-complicated. The `(if ... (match ...))` structures are confusing. See this example implementation for comparison: https://github.com/MarvinJanssen/clarity-experiments/blob/main/contracts/sip009-nft.clar. 

General remark: I think it makes sense if we put example contracts in a separate repository. (And thus also move the existing ones.)

pgray-hiro:
Hey @MarvinJanssen, thanks for the feedback, Alex is actually the author of this contract, but I can see if I can rework the implementation.

I agree it would be nice to have a single repo of example contracts that we can use as a canonical example set. Is that something you'd like to setup from the Foundation side or should I set a separate repo up? Until we have that location I'd like to keep these examples here as they are referenced in the Stacks documentation.

As far as the reference contracts go, I would love to do write-ups/tutorials for each of them in the Stacks docs. Do you have them stored somewhere where I could look at them, or is it too early for that?

lgalabru:
What would be the next actions here @pgray-hiro?

lgalabru:
This PR has been staling for a few months, I'll close it for now but will keep the branch around in case we want to re-open and address feedbacks.

PR #80, opened 2021-07-14
feat: introducing devnet

lgalabru:
DevNet is a package of components that equip developers with an entire suite of Stacks utilities, directly on their local machine for rapid development and testing. Out of the box it provides a bitcoinid node for obtaining information from the bitcoin blockchain, a Stacks node for interacting with the network, the Stacks API for performing meaningful actions, and finally a Bitcoin and Stacks explorer for querying the historical record of transactions. It’s also super simple to integrate DevNet into our other recent developer tool, Clarinet. 

Dapp development on Stacks as a process is primarily composed of two big phases. The first being smart contract development. Clarinet already helpped developers with this phase by providing a testing framework and console for experimentation. Once devs have their smart contracts, the next step is connecting them with a frontend for user interfacing. That’s when the focus shifts to smart contract integration. At this point, devs need to test their contracts integrated with the whole configured environment: bitcoin-node, stacks-node, stacks-api, and the ability to explore the blocks and transactions.

This type of development workflow is what the `clarinet integrate` command is bringing to the table: a simple, local workflow where developers don't have to deal with infrastructure, deployments, or docker orchestrations. Clarinet is coordinating all these services, deploying contracts automatically, providing an environment for developers so they can focus on the development of their dapp. This DevNet beta should be interesting to any team that have some contracts already prototyped, and are maybe beginning to work on a frontend.

Screen Shot 2021-07-14 at 12 10 53 AM


diwakergupta:
@lgalabru we previously discussed, let's start following our standard dev process for Clarinet, in particular code reviews :) We can discuss specifics after the Clarinet meeting today.

CharlieC3:
@lgalabru Can we change the base branch to `develop` :)

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+        }
+
+
+        if let Some(ref bitcoin_blockchain_container_id) = self.bitcoin_blockchain_container_id {
+            println!("Terminating bitcoin-node...");
+            let _ = docker
+                .kill_container(bitcoin_blockchain_container_id, options.clone())
+                .await;
+            let _ = docker.remove_container(bitcoin_blockchain_container_id, None);
+        }
```
CharlieC3, line 1147: We should probably terminate this container last since it's a dependency of the `stacks-node` container, inversely of how it's the first container to be started. Doing it in this current order _shouldn't_ hurt anything, but it's a good practice to follow in general.

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+                Some(CreateImageOptions {
+                    from_image: devnet_config.bitcoin_node_image_url.clone(),
+                    ..Default::default()
+                }),
+                None,
+                None,
+            )
+            .try_collect::<Vec<_>>()
+            .await
+            .map_err(|_| "Unable to create image".to_string())?;
```
CharlieC3, line 303: Might be useful to print out the container ID in debug or verbose logs, most advanced docker users would find this useful - [example](https://github.com/fussybeaver/bollard/blob/master/examples/image_from_scratch.rs#L75-L80)

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+rpcserialversion=0
+rpcworkqueue=100
+disablewallet=0
+fallbackfee=0.00001
+
+[regtest]
+bind=0.0.0.0:{}
+rpcbind=0.0.0.0:{}
+rpcport={}
+"#,
```
CharlieC3, line 352: Future idea: Would be cool if `devnet` could do two things WRT container configs:

1. Optionally read/ingest configs on the local filesystem for any of the containers it runs. Otherwise use the default configs like this.
2. Have a subcommand under `devnet` like `clarinet devnet write-configs` to write the default container configs to the local filesystem for better debugging visibility, and to give devs an easy starting point to modify their devnet configs for a more custom deployment.

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+            name: format!("bitcoin.{}", self.network_name),
+        };
+
+        let container = docker
+            .create_container::<String, String>(Some(options), config)
+            .await
+            .map_err(|e| format!("Unable to create container: {}", e))?
+            .id;
+
+        self.bitcoin_blockchain_container_id = Some(container.clone());
```
CharlieC3, line 464: Similar to the above comment to print out the image ID, debug or verbose logs for printing this would be helpful for Docker users as well.

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+        );
+
+        let mut labels = HashMap::new();
+        labels.insert("project".to_string(), self.network_name.to_string()); 
+
+        let config = Config {
+            labels: Some(labels),
+            image: Some(devnet_config.bitcoin_node_image_url.clone()),
+            domainname: Some(self.network_name.to_string()),
+            tty: Some(true),
```
CharlieC3, line 439: Curious, what purpose is the TTY serving in `devnet`? Is it necessary for some operations?

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+            println!("{}", err);
+            return Err(err);
+        }
+
+        Ok(())
+    }
+
+    pub async fn restart(&mut self) {}
+
+    pub async fn terminate(&mut self) {
```
CharlieC3, line 1111: It might be useful to incorporate a `stop` function to allow devs to stop their devnet containers without terminating them and losing progress on their local cluster. This is possible if the dev uses the `docker` CLI directly, but would be a nice-to-have if devs can do this while in `devnet` land :)

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+count = 2
+block_time = 20000
+ignore_txs = false
+
+# Give more time to the first blocks
+[[blocks]]
+count = 3
+block_time = 45000
+ignore_txs = false
+"#,
```
CharlieC3, line 402: It might be a bit difficult to find these configs in the source code, would it be possible to separate them out into separate files that get included in the shipped binary?

CharlieC3 (commented):
General feedback! I only reviewed the orchestrator, but what you're doing seems solid for devnet launch 🥇

src/integrate/orchestrator.rs
lgalabru: this configs are being written in the working dir:

```
➜  stacks-devnet-1628076160 tree .
.
├── conf
│   ├── Config.toml
│   ├── bitcoin.conf
│   └── puppet-chain.toml
├── data
│   └── 1
│       ├── bitcoin
│       │   └── regtest
|       ...
│       └── stacks
│           └── krypton
|       ....
└── devnet.log
```

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+                .await;
+            let _ = docker.remove_container(bitcoin_explorer_container_id, None);
+        }
+    
+        if let Some(ref stacks_explorer_container_id) = self.stacks_explorer_container_id {
+            println!("Terminating stacks-explorer...");
+            let _ = docker
+                .kill_container(stacks_explorer_container_id, options.clone())
+                .await;
+        }
```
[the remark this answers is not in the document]
lgalabru: Addressed with latest

src/integrate/orchestrator.rs
lgalabru: Addressed with latest

src/integrate/orchestrator.rs
lgalabru: In this we're not trying to gracefully terminate services, just killing everything, so order doesn't really matter

src/integrate/orchestrator.rs
lgalabru: Addressed with latest

src/integrate/orchestrator.rs
lgalabru: Addressed with latest

src/integrate/orchestrator.rs
lgalabru: I'll keep this idea in mind for a future release :)

src/integrate/orchestrator.rs
```diff
@@ -0,0 +1,1204 @@
...
+            .expect("Unable to create working dir");
+        fs::create_dir(format!("{}/data/stacks", devnet_config.working_dir))
+            .expect("Unable to create working dir");
+
+        let bitcoin_explorer_port = devnet_config.bitcoin_explorer_port;
+        let stacks_explorer_port = devnet_config.stacks_explorer_port;
+        let stacks_api_port = devnet_config.stacks_api_port;
+
+        let _ = event_tx.send(DevnetEvent::ServiceStatus(ServiceStatusData {
+            order: 0,
```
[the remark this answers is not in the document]
lgalabru: Agreed. I'm thinking about approaching this component a bit differently, will address this feedback in a future refactor.

src/integrate/orchestrator.rs
```diff
@@ -731,16 +962,16 @@ events_keys = ["*"]
...
         let config = Config {
             labels: Some(labels),
             image: Some(devnet_config.stacks_api_image_url.clone()),
             domainname: Some(self.network_name.to_string()),
-            tty: Some(true),
+            tty: None,
             exposed_ports: Some(exposed_ports),
             env: Some(vec![
-                format!("STACKS_CORE_RPC_HOST=stacks.{}", self.network_name),
+                format!("STACKS_CORE_RPC_HOST=stacks-node.{}", self.network_name),
```
CharlieC3, line 974: Non-blocking: Looks like the container names, like `stacks-node`, is being referenced in a few places. If you decided to change the container name, it may break this ENV var accidentally if you don't also update it. Might be worth putting the container names in global vars to keep it consistent and easy to change.

CharlieC3 (changes requested):
Looks great! Only have one small change needed that I found. The other two are low priority and non-blocking -- they do not need to delay any kind of release.

src/integrate/orchestrator.rs
```diff
@@ -430,38 +540,128 @@ ignore_txs = false
...
+            "label".to_string(),
+            vec![format!("project={}", self.network_name)],
+        );
+        let options = Some(ListContainersOptions {
+            all: true,
+            filters,
+            ..Default::default()
+        });
+
+        let docker = match &self.docker_client {
```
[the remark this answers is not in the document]
lgalabru: Good idea, will follow up with a subsequent PR.

src/integrate/orchestrator.rs
lgalabru: Good idea, will follow up with a subsequent PR!

CharlieC3 (approved):
:shipit:

Issue #83, opened 2021-07-16
Automatically update winget version upon new release


lgalabru:
Related: https://github.com/hirosystems/clarinet/issues/73

Pointer: https://github.com/microsoft/winget-create

PR #87, opened 2021-07-21
ci: post release workflows

CharlieC3:
* Fixes Homebrew automatic submission by adding missing env var needed for Github authentication
* Adds winget automatic submission - [tested here](https://github.com/microsoft/winget-pkgs/pull/21914)

Closes https://github.com/hirosystems/clarinet/issues/83

### DIFF
--- a/.github/workflows/pkg-version-bump.yaml
+++ b/.github/workflows/pkg-version-bump.yaml
@@ -18,6 +18,10 @@ jobs:
           git config --global user.name "Hiro DevOps"
           git config --global user.email "45208873+blockstack-devops@users.noreply.github.com"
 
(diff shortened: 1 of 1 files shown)


lgalabru (approved):
LGTM! Should we release a `0.14.3` to confirm that the whole workflow works?

CharlieC3:
> LGTM! Should we release a `0.14.3` to confirm that the whole workflow works?

I _do_ feel more confident in the workflows now, but past experience tells me this could still fail, and I don't want to fall into a loop of creating releases to test post-release activities in other open-source projects.

I'm fine with waiting until the next real release and adjusting then if needed :)

Issue #97, opened 2021-08-11
Clarinet requires rust nightly to compile


CharlieC3:
In order for clarinet to stay compliant with homebrew requirements, it must be able to compile on a stable Rust version. [See discussion here for more info.](https://github.com/Homebrew/homebrew-core/pull/82839)

lgalabru:
Ha, I was not aware of this requirement.
To be addressed with https://github.com/hirosystems/clarinet/pull/100

Issue #94, opened 2021-08-08
`host.docker.internal` is not avaliable on Linux


LNow:
I've tried to use `clarinet integrate` command, and unfortunately it doesn't work on Linux machines.
The reason for that is on Linux `host.docker.internal` is not available. Linux equivalent of this host name is 172.17.0.1 (by default).

dantrevino:
Isn't this something that can be set conditionally based on os?

LNow:
On linux to have access to `host.docker.internal` inside container we have to add it explicitly to container. For example:
```sh
$ docker run --rm -it --add-host=host.docker.internal:host-gateway node:lts-alpine3.14
```

lgalabru:
To be addressed with https://github.com/hirosystems/clarinet/pull/100

zone117x:
The stacks-blockchain-api ran into this with github CI ubuntu runners. Was able to use this (quite old) docker entry script to create `host.docker.internal` when running in linux:
https://github.com/bufferings/docker-access-host/blob/master/docker-entrypoint.sh

I also had to disable the gh ubuntu firewall for this to work, but since then, it's been working well.

lgalabru:
Confirmed with @LNow - #100 is fixing the issue.

PR #100, opened 2021-08-16
Devnet adjustments

lgalabru:
This PR is upgrading to a version of Rocket (currently in RC) that does not require rust nightly, to address https://github.com/hirosystems/clarinet/issues/97 and fix distribution via brew.
It's also fixing some compatibility issues with Linux, raised in https://github.com/hirosystems/clarinet/issues/94.

### DIFF
--- a/src/integrate/orchestrator.rs
+++ b/src/integrate/orchestrator.rs
@@ -985,7 +985,7 @@ events_keys = ["*"]
                 ),
                 format!("STACKS_CORE_EVENT_HOST=0.0.0.0"),
                 format!("PG_HOST=postgres.{}", self.network_name),
-                format!("PG_PORT={}", devnet_config.postgres_port),
+                format!("PG_PORT=5432"),
                 format!("PG_USER={}", devnet_config.postgres_username),
                 format!("PG_PASSWORD={}", devnet_config.postgres_password),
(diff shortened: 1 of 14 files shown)


src/integrate/orchestrator.rs
```diff
@@ -552,14 +552,14 @@ ignore_txs = false
...
                 port_bindings: Some(port_bindings),
-
                 binds: Some(vec![
                     format!("{}/conf:/etc/bitcoin", devnet_config.working_dir),
                     format!(
                         "{}/data/{}/bitcoin:/root/.bitcoin",
                         devnet_config.working_dir, boot_index
                     ),
                 ]),
+                extra_hosts: Some(vec!["host.docker.internal:172.17.0.1".into()]),
```
zone117x, line 562: Last time I looked at this issue (over a year ago), I think it was possible for the docker gateway to be a different IP. Maybe needs to be detected, or configurable?

wileyj: it is possible, that was my thought as well - hardcoding the IP here will surface issues with users that use different subnets for docker

lgalabru: Yeah the hard coded IP address was an unsuccessful attempt to patch the issue, will revert.

lgalabru:
Was able to successfully run devnet on a linux box, with the latest change.

PR #102, opened 2021-08-16
Change default stacks_api_port to 3999

aviculturist:
Port 3999 is the default port used in Hiro Wallet for LocalNet
and this change allows clarinet integration to work out of the box


lgalabru (approved):
LGTM, thank you @aviculturist!

PR #106, opened 2021-08-23
Clarinet Extension #1

lgalabru:
This PR includes: 
- 1st functional Clarinet extension (hopefully useful!), 
- some wiring so that extensions can be automatically ran when devnet starts
- some fixes to make it so dependencies can be used with `clarinet check` and `clarinet test`.
- some in-memory caching optimizations that are boosting test execution speed


examples/counter/scripts/abi-generator.ts
```diff
@@ -0,0 +1,117 @@
...
+
+        const write = await Deno.writeTextFile("./artifacts/contracts.ts", code.flat().join("\n"));
+    }
+});
+
+function capitalize(source: string): string {
+    return source[0].toUpperCase() + source.slice(1);
+}
+
+function snakeToCamel(source: string): string {
```
zone117x, line 111: Unused function?

examples/counter/scripts/abi-generator.ts
```diff
@@ -0,0 +1,117 @@
...
+            }
+
+            if (public_funcs.length > 0) {
+                code.push([
+                    `    // Public functions`,
+                    `    export enum Public {`,
+                ]);
+                for (let f of public_funcs) {
+                    code.push([
+                        `        ${kebabToCamel(capitalize(f.name))} = "${f.name}",`,
```
zone117x, line 72: This normalizes hyphens, but what if `name` has other characters that are not valid js identifier syntax? E.g. question marks, exclamation marks?

examples/counter/scripts/abi-generator.ts
```diff
@@ -0,0 +1,117 @@
...
+// }
+//
+// By importing this file in their frontend code, developers can use constants, instead 
+// of hard coding principals and strings:
+// 
+// await makeContractCall({
+//     contractAddress: CounterContract.ADDRESS,
+//     contractName: CounterContract.NAME,
+//     functionName: CounterContract.Public.INCREMENT,
+//     functionArgs: [CounterContract.Args.INCREMENT(namespace)], // or so
```
zone117x, line 31: Is this `functionArgs` example a TODO / WIP?

examples/counter/scripts/abi-generator.ts
```diff
@@ -0,0 +1,117 @@
...
+        code.push([
+            `// Code generated with the clarinet-abi-generator extension`,
+            `// Manual edits will be overwritten`,
+            ``,
+        ]);
+
+        for (let [contractId, contract] of contracts) {
+            let [address, name] = contractId.split(".");
+            code.push([
+                `export namespace ${kebabToCamel(capitalize(name))}Contract {`,
```
zone117x, line 48: I'm curious about the usage of typescript `namespace`, any specific reason for it? I don't think I've seen it used outside of .d.ts files before.

lgalabru: I'm using a namespace to create some of scope, to avoid collisions when you have multiple contracts. How would you approach this?

examples/counter/scripts/abi-generator.ts
lgalabru: Addressed with latest

examples/counter/scripts/abi-generator.ts
lgalabru: Addressed with latest

examples/counter/scripts/abi-generator.ts
lgalabru: Addressed with latest

examples/counter/scripts/abi-generator.ts
lgalabru: We can revisit the approach if it turns out being a disagreement for developers using this extension.
The goal of this PR is to demonstrate how to use extensions and how they can be used for automating tasks.

zone117x (approved):
LGTM!

PR #148, opened 2021-10-27
Add 128bit numbers support

### LINKED ISSUE #82: Use native js `bigint` rather than `number`
Looks like 128-bit Clarity integers are being represented as js `number`'s (64-bit floating point). This causes problems with numbers that take ~53 or more bits to represent. I think the best solution here is to use native js `bigint` to represent these values. 

IMO, helper constructors like 
https://github.com/hirosystems/clarinet/blob/9beb61a538be055c220886300705eac552f6bf28/deno/index.ts#L318-L320
should continue to accept `number` as well as `bigint` (principal of being most permissive for inputs, but most explicit for outputs):
```ts
function uint(val: number | bigint) {
  return `u${val}`;
}
```

But response values (that represent 128-bit ints/uints) should return a `bigint`. That would result in this example code
https://github.com/hirosystems/clarinet/blob/9beb61a538be055c220886300705eac552f6bf28/examples/billboard/tests/billboard_test.ts#L34-L35
becoming
```ts
        assetMaps = chain.getAssetsMaps();
        assertEquals(assetMaps.assets['STX'][wallet_1.address], balance - 210n);
```

Obviously this isn't much of a concern with something like a STX balance that is unlikely to be higher than a 53-bit integer, but FTs and many other use cases use larger numbers where this becomes an issue.

LNow:
This PR adds 128bit numbers support.

Close #82

### DIFF
--- a/deno/index.ts
+++ b/deno/index.ts
@@ -360,8 +360,8 @@ declare global {
     expectSome(): String;
     expectNone(): void;
     expectBool(value: boolean): boolean;
-    expectUint(value: number|bigint): number|bigint;
-    expectInt(value: number|bigint): number|bigint;
+    expectUint(value: number|bigint): bigint;
+    expectInt(value: number|bigint): bigint;
     expectBuff(value: ArrayBuffer): ArrayBuffer;
     expectAscii(value: String): String;
     expectUtf8(value: String): String;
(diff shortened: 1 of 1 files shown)


lgalabru (approved):
Thank you for this contribution @LNow, LGTM!

lgalabru:
@CharlieC3 what would be the next steps, concerning the windows build error?

Screen Shot 2021-11-02 at 12 26 37 PM

CharlieC3:
@lgalabru The build is failing because a GH secret is required, and the PR is from a fork which isn't allowed to use GH secrets. Unfortunately there is no way around this besides re-opening the PR from an internal branch or skipping the step (and others) entirely.

lgalabru:
Oh ok that's fine then, I thought that we were looking at an expired certificate or something.
Thanks!

PR #155, opened 2021-11-06
Feat: add assertions for NFT events

LNow:
This PR adds working assertions for all 3 (mint, transfer, burn) NFT events.

I decided to detect events not only based on `assetId` but also `assetAddress`. Same contract can be deployed using different addresses and in such cases using only `assetId` is not selective enough and can give false positive results.

I added example project with simple SIP-009 compliant contract and tests that shows how these new assertions could be used.

### DIFF
--- a/deno/index.ts
+++ b/deno/index.ts
@@ -404,7 +404,12 @@ declare global {
       assetAddress: String,
       assetId: String
     ): Object;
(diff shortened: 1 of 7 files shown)


lgalabru (approved):
Thank you for this contribution @lnow and implementing these new assertions - LGTM!

PR #156, opened 2021-11-06
Improve argument types

friedger:
This PR
* set the argument type to a specific type instead of `ClarityValue`

Still to do: The generated import should reflect the used types. What is the best approach here?


lgalabru (approved):
Neat addition, thank you @friedger!

Issue #168, opened 2021-12-07
Updating pass on dependencies


lgalabru:
A bunch of warnings are being emitted by `cargo audit`.
RUSTSEC-2021-0076
RUSTSEC-2021-0079
RUSTSEC-2021-0078
RUSTSEC-2021-0093
RUSTSEC-2020-0159
RUSTSEC-2021-0119
RUSTSEC-2020-0071
RUSTSEC-2021-0124
They do not seem critical in the context of Clarinet, but most of them could be addressed with crate updates at this point.

Issue #136, opened 2021-10-15
[Feature] Opt-in Telemetry


agraebe:
It would be great to integrate opt-in telemetry to collect crashes/errors and general usage data in an anonymous fashion. Here are a few ideas for events to track:

## General
- [ ] Number of installations, including version
- [ ] Number of commands run per installation/user

## Errors
- [ ] Unexpected errors (input command, response/error trace)
- [ ] Unrecognized commands run (input command)

## Usage
- [ ] Most used commands
- [ ] How many users get to run tests
- [ ] How many users get to run the DevNet
- [ ] How many users get to publish contracts

agraebe:
Quick update on this - we've chatted about the telemetry requirements and the idea is to leverage Segment as the aggregator. From within Segment, we have the flexibility to set destinations without changing the implementation.

@lgalabru does that sound good to you?

PR #176, opened 2021-12-15
Introducing telemetry

lgalabru:
Address https://github.com/hirosystems/clarinet/issues/136, https://github.com/hirosystems/clarinet/issues/168

Telemetry is being introduced, as a gated feature.
The version of clarinet being distributed on Brew, Winget and Docker are being compiled with this feature. If developers want to use a telemetry free version, they can compile locally Clarinet without passing the `--features=telemetry` flag.
When developers are generating a new project, the following prompt is being displayed:
```
Send usage data to Hiro.
Help Hiro improve its products and services by automatically sending diagnostics and usage data [Y/n]:
```
Depending on developers input, a setting `telemetry` is enabled / disabled in the project's manifest: 
```
[project]
...
telemetry = true
```
This setting can be adjusted manually anytime. If the `telemetry` flag is absent from the config, telemetry is disabled by default.
If Clarinet has telemetry available and the developers opted in for telemetry in the context of the ongoing project, events will be raised when:
- New project is generated
- Test suite is executed
- Devnet is spawned
- Console session is started
- Contracts are published

Events properties (project name, authors, fingerprint) are being hashed locally - PII are not being sent over the wire.


src/frontend/telemetry.rs
```diff
@@ -0,0 +1,146 @@
...
+        .enable_time()
+        .max_blocking_threads(32)
+        .build()
+        .unwrap();
+
+    rt.block_on(send_event(event));
+}
+
+async fn send_event(event: DeveloperUsageEvent) {
+    let segment_api_key = "<REPLACE_SEGMENT_API_KEY>";
```
obycode, line 48: Is this secure? I would expect we should only have an API key embedded if this was running server-side for security reasons.

obycode (changes requested):
Looking good. A few comments.

src/frontend/telemetry.rs
obycode: Oh, I see... is this just the write key, which is intended to be public?

src/frontend/cli.rs
```diff
@@ -209,11 +230,33 @@ pub fn main() {
                 current_dir.to_str().unwrap().to_owned()
             };
 
-            let changes = generate::get_changes_for_new_project(current_path, project_opts.name);
+            let telemetry_enabled = if cfg!(feature = "telemetry") {
```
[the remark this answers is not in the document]
lgalabru: Thank you for your review and your help on the wording. Will address!

src/frontend/telemetry.rs
lgalabru: This is a placeholder that the CI should be seeding (via a Github Secret) before building releases. I might have to summon @CharlieC3 for the last mile on this one.

obycode: Right, I figured that the CI job would be filling it in. I was just worried about putting an API key into the deployed binary, since if it is in there, it could be extracted. After I looked at the Segment docs I realized that it looks like the "write key" is expected to be public, so that should not be a problem. 

Actually, since that is the case, I guess it should be okay to just include it in the repo, right?

src/frontend/cli.rs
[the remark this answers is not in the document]
lgalabru: Addressed with https://github.com/hirosystems/clarinet/pull/176/commits/a35bcb3e755529430796dbd5ecd4914e18c1b8ad

src/frontend/cli.rs
```diff
@@ -209,11 +230,33 @@ pub fn main() {
...
+                if let Some(disable_telemetry) = project_opts.disable_telemetry {
+                    !disable_telemetry
+                } else {
+                    println!("{}", yellow!("Send usage data to Hiro."));
```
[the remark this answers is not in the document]
lgalabru: Addressed with https://github.com/hirosystems/clarinet/pull/176/commits/a35bcb3e755529430796dbd5ecd4914e18c1b8ad

src/frontend/telemetry.rs
CharlieC3: @obycode Can you point out where it says the API key is expected to be public? My understanding is that anyone that has the API key can write to our Segment account.

obycode: They have to be public because they can be used on a frontend to send telemetry to Segment, so there is no choice but to expose them in the JS code. 

Here's something from an incident report:
> ... including customer write keys for Segment (which are considered public),...
https://segment.com/blog/bulletins-incident090519/

I don't see it in the main docs, but it makes sense if they're telling you to use them on the client side, they have to be exposed.

obycode: Yes, anyone with that write key could send fake data to our Segment account. The hope is just that there is no motivation to do so. I guess it doesn't hurt if we leave it out of the repo, so an attacker would at least have to do a little work to extract the key from the binary.

obycode: The only problem will be if someone builds from source and enables telemetry, they'll have to fill in that field or it will get errors all the time.

CharlieC3: Interesting, thanks for the info! There may still be some value in not making it easy for anyone to obtain (e.g. plaintext in the repo), so it could still be injected during CI. But we'd need to be sure people can still build clarinet locally without needing to supply  a Segment API key. Is that something we can ensure?

obycode: Looks like we had the same thought there 😄. We could just add a note in the README warning about enabling telemetry when building from source.

CharlieC3: ha you beat me to it by seconds! 😆 
Yeah that sounds good. Is telemetry enabled by default? Maybe it should be disabled by default so users don't `git pull` the latest changes and have to do something to mitigate them.

CharlieC3 (approved):
LGTM!

Once this is merged to `main` and a new version is created, we'll need to update the [homebrew Forumla](https://github.com/Homebrew/homebrew-core/blob/master/Formula/clarinet.rb) to include the `telemetry` feature flag.

obycode (approved):
Looks great!

PR #178, opened 2021-12-16
Compatibility with clarity-repl v0.17.0+

lgalabru:
In clarity-repl v0.17.0, we updated the type of `result` in the `ExecutionResult` data structure - from String to ClarityValue.
The interface Deno/Rust in Clarinet is still expecting a Clarity value formatted as a string, and was acting strangely in presence of a ClarityValue. This PR is fixing this issue.


obycode (approved):
LGTM

PR #182, opened 2021-12-20
Migrate LSP into clarinet

lgalabru:
Something that has been in the back of my mind for sometimes now.
With this PR, the idea is to move the Clarity LSP Rust implementation to Clarinet. There are several good reasons to make that move:
- The LSP is clearly a second class citizen today. By moving this feature to this repo, it would become top of mind,
- We want to leverage the amazing pipeline that we have in place in this repo. Apple M1 users are struggling with the LSP, which is something that we solved with this repo, thanks to brew.
- https://github.com/hirosystems/clarity-lsp is going to be re-purposed as a web extension soon, that can be loaded in vscode.dev / github.dev. However, we want to keep this feature available for Vim / Emacs / etc, and Clarinet is a good candidate for hosting this feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,9 +14,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.16.0"
+version = "0.15.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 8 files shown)


obycode (approved):
I agree. This makes a lot of sense. Looks good to me!

PR #195, opened 2022-01-04
feat: added ignore and only options to clarinet test

hugocaillard:
pass the option `only: true` to `Clarinet.test({})` to run only this test
pass the option `ignore: true` to `Clarinet.test({})` to ignore the test

### DIFF
--- a/deno/index.ts
+++ b/deno/index.ts
@@ -186,6 +186,8 @@ type PreSetupFunction = () => Array<Tx>;
 
 interface UnitTestOptions {
   name: string;
+  only?: true;
+  ignore? :true;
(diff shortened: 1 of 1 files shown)


obycode:
Thanks @hugocaillard! Could you just add some usage information to the body of the commit message? Soon, we'll have a better place to document stuff like this, but for now, an explanation in the commit message should suffice.

hugocaillard:
@obycode Like so or does it need to be more detailed?
Couldn't find an exemple in the past commits

obycode (approved):
👍 looks good, thanks!

Issue #213, opened 2022-01-12
Backslash not removed when escaping quotes in strings


hugocaillard:
On clarinet the escaping backslashes are not removed in strings (ascii or utf8).
On the testnet it works as intended

Code to reproduce:
```lisp
(define-public (say-hello)
  ;; returns `Hello \"World\"` instead of `Hello "World"`
  (ok "Hello \"World\"")
)
```

I also realised that backslashes are added before single quotes. Please tell me if I should open a second issue about that
```lisp
(define-public (say-hello2)
  ;; returns `Hello \'World\'` instead of `Hello 'World'`
  (ok "Hello 'World'")
)
```

lgalabru:
Interesting, thanks for opening this issue.
Clarinet should probably be dumping .clar files in vector of bytes that we would pass to the clarity-repl.
Thoughts @obycode?

obycode:
Interesting... the string is un-escaped after the parser. I'll dig into it a bit more.

obycode:
I believe the problem is that the result of mine block is being processed by a `format!` (deno.rs:1218)
```rust
    Some(output) => format!("{}", output),
```

And the implementation for the String clarity value uses this function from clarity/types/mod.rs:415:
```rust
impl fmt::Display for ASCIIData {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let mut escaped_str = String::new();
        for c in self.data.iter() {
            let escaped_char = format!("{}", std::ascii::escape_default(*c));
            escaped_str.push_str(&escaped_char);
        }
        write!(f, "{}", format!("\"{}\"", escaped_str))
    }
}
```

I should be able to fix it tomorrow.

obycode:
Ok, I've got a fix in 5505ad83. I'm going to add some unit tests before opening a PR. I've basically just added a recursive function to print the `Value`s instead of using the `Display` trait as defined in the clarity library, since that is where the character encoding happens. If either of you have a better name for this function, let me know (currently `value_to_string`).

hugocaillard:
Tested locally and it's fixed 🙌
Thx

PR #217, opened 2022-01-16
fix: generate proper strings from contract results

obycode:
Fixes: #213


lgalabru (approved):
LGTM, thank you @obycode!

PR #222, opened 2022-01-18
Don't attempt to sign Windows executable on merge to develop

obycode:
Testing a change in the CI script to avoid an error from code-signing in Windows from outside developers.


CharlieC3 (approved):
:shipit:

PR #230, opened 2022-01-27
feat: update to work with new parser

obycode:
This corresponds to changes in hirosystems/clarity-repl#85.

### DIFF
--- a/src/generate/project.rs
+++ b/src/generate/project.rs
@@ -175,9 +175,9 @@ history.txt
 name = "{}"
 authors = []
 telemetry = {}
-[project.repl.analysis]
+[repl.analysis]
 passes = ["check_checker"]
-[project.repl.analysis.check_checker]
+[repl.analysis.check_checker]
 # If true, inputs are trusted after tx_sender has been checked.
 trusted_sender = false
 # If true, inputs are trusted after contract-caller has been checked.
--- a/src/generate/project.rs
+++ b/src/generate/project.rs
@@ -177,6 +177,15 @@ authors = []
 telemetry = {}
 [project.repl.analysis]
 passes = ["check_checker"]
+[project.repl.analysis.check_checker]
+# If true, inputs are trusted after tx_sender has been checked.
+trusted_sender = false
+# If true, inputs are trusted after contract-caller has been checked.
+trusted_caller = false
+# If true, untrusted data may be passed into a private function without a
+# warning, if it gets checked inside. This check will also propagate up to the
+# caller.
+callee_filter = false
 
 # [contracts.counter]
 # path = "contracts/counter.clar"
--- a/src/generate/project.rs
+++ b/src/generate/project.rs
@@ -175,7 +175,8 @@ history.txt
 name = "{}"
 authors = []
 telemetry = {}
-analysis = ["check_checker"]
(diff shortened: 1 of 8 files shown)


obycode:
These tests are expected to fail until the corresponding changes are merged into the REPL (hirosystems/clarity-repl#85).

src/types/project_manifest.rs
```diff
@@ -21,6 +21,7 @@ pub struct ProjectConfigFile {
     requirements: Option<Value>,
     analysis: Option<Vec<String>>,
     costs_version: Option<u32>,
+    parser_version: Option<u32>
```
lgalabru, line 24: What's your take on spawning a new section in the schema of the manifest, dedicated to tweaking the VM / REPL internals?

obycode: Yes, that's probably a good idea. My check-checker changes (PR coming soon) is also tweaking this structure, to pass options directly to analysis passes:
```toml
[project.analysis_settings.check_checker]
trusted_sender = true
```

Do you think something like this would make sense:
```toml
[project]
name = "my-contract"
authors = []
description = ""
telemetry = false
requirements = []

[contracts.nft-trait]
path = "contracts/nft-trait.clar"
depends_on = []

[contracts.test]
path = "contracts/test.clar"
depends_on = ["nft-trait"]

[repl]
parser_version = 1
analysis = ["check_checker"]

[repl.analysis_settings.check_checker]
trusted_sender = true
```

obycode:
Note - we also need to push a release of clarity-repl and update this Cargo.toml file accordingly before pushing this.

src/types/project_manifest.rs
obycode: Now I'm thinking the `analysis_settings` struct is not necessary and it could just be `repl.check_checker`.

obycode: Ok, actually I have a much nicer fix to this in my check-checker branch. Mind if we hold off on this change until that one?

lgalabru (approved):
This PR looks good to me, thank you @obycode!
Before merging, could you do a sanity pass and make sure that the LSP is still functional with these changes?
Thanks!

src/types/project_manifest.rs
obycode: This is implemented in the latest commits.

lgalabru (dismissed):

src/generate/project.rs
```diff
@@ -175,7 +175,17 @@ history.txt
 name = "{}"
 authors = []
 telemetry = {}
-analysis = ["check_checker"]
+[project.repl.analysis]
```
lgalabru, line 178: What's your take on flattening a bit the schema here? `repl.analysis`, instead of `project.repl.analysis`?

obycode: Sure, that makes sense. I will quickly make that change and push.

obycode: Changed in a569502

lgalabru (approved):
Looks great, thank you @obycode!

unclemantis:
We stable yet?

obycode:
> We stable yet?
> […](#)

😄 Yup, merging and releasing now!

Issue #231, opened 2022-01-28
Failed to select version for 'tokio' on `cargo install`


whoabuddy:
I was upgrading to the latest version today (`v0.24.0`) with `cargo install --path . --locked` and received the following error:

```
  Installing clarinet v0.24.0 (/home/whoabuddy/Dev/stacks/clarinet)
    Updating crates.io index
error: failed to select a version for `tokio`.
    ... required by package `deno v1.11.0 (/home/whoabuddy/Dev/stacks/clarinet/vendor/deno/cli)`
    ... which is depended on by `clarinet v0.24.0 (/home/whoabuddy/Dev/stacks/clarinet)`
versions that meet the requirements `=1.8.1` are: 1.8.1

all possible versions conflict with previously selected packages.

  previously selected package `tokio v1.14.0`
    ... which is depended on by `clarinet v0.24.0 (/home/whoabuddy/Dev/stacks/clarinet)`

failed to select a version for `tokio` which could resolve this conflict
```

This was with `clarinet 0.20.0` already installed, previously I've had no issue updating this way, is there an extra step I need to take?

obycode:
Hi @whoabuddy. I haven't seen that before. Could you try a clean install and let me know if you still get this error?
```
cargo clean
cargo install --path . --locked
```

whoabuddy:
Hi @obycode - tried that and still the same error. Looking at my rust/cargo versions installed I'm a little behind:

```
rustc 1.55.0 (c8dfcfe04 2021-09-06)
cargo 1.55.0 (32da73ab1 2021-08-23)
```

I'm updating those now and will try a clean install after then report back.

whoabuddy:
With the latest versions I still saw the same error:

```
rustc 1.58.1 (db9d1b20b 2022-01-20)
cargo 1.58.0 (f01b232bc 2022-01-19)
```

Based on [this article](https://catto.io/2021/10/fix-failed-to-select-a-version-cargo/) I tried the following:

```bash
rm -rf ~/.cargo/registry
cargo clean
cargo install --path . --locked
```

Unfortunately, I'm still running into a similar error:

```
  Installing clarinet v0.24.0 (/home/whoabuddy/Dev/stacks/clarinet)
    Updating crates.io index
error: failed to select a version for `tokio`.
    ... required by package `deno v1.11.0 (/home/whoabuddy/Dev/stacks/clarinet/vendor/deno/cli)`
[...]

PR #232, opened 2022-01-29
docs: clarify develop vs main branch

obycode:
Since the default branch is `develop`, when users clone the repo, it
will be in this branch by default. This change clarifies in the README
that if they want the latest stable code, they should switch to the
`main` branch.

Add information about updated the deno submodule when pulling into a
previously checked out repo.

Resolves: #231


lgalabru (approved):
LGTM!

PR #238, opened 2022-02-04
Add support for multiple deployers.

fariedt:
This adds an optional deployer string field to contract sections
in Clarinet.toml.  If the key is not present, clarinet defaults to
the deployer account in the settings file.

Tested in the console and by publishing to the testnet.


lgalabru:
Thank you @fariedt!
Your work might be conflicting with https://github.com/hirosystems/clarinet/pull/240 but I'll do my best to port your feature over.

PR #261, opened 2022-02-15
docs: improve billboard example

obycode:
The addresses were hard-coded into the test, and it is not obvious to
new developers where those are coming from. This change uses the account
names to get the addresses instead.

See hirosystems/docs#67


lgalabru (approved):
LGTM!

Issue #263, opened 2022-02-17
Rely on rustls instead of openssl-sys


lgalabru:
We should use `rustls` when possible, and try to avoid depending on openssl-sys - causing some challenges in our CI/CD pipeline.
 
```
% cargo tree --target x86_64-unknown-linux-musl -i openssl-sys
openssl-sys v0.9.65
├── native-tls v0.2.7
│   ├── hyper-tls v0.5.0
│   │   └── reqwest v0.11.4
│   │       ├── clarinet v0.25.0 (/Users/matt/Projects/clarinet)
│   │       ├── clarity-repl v0.22.0
│   │       │   └── clarinet v0.25.0 (/Users/matt/Projects/clarinet)
│   │       ├── deno_fetch v0.30.0 (/Users/matt/Projects/clarinet/vendor/deno/extensions/fetch)
│   │       │   └── deno_runtime v0.16.0 (/Users/matt/Projects/clarinet/vendor/deno/runtime)
│   │       │       ├── clarinet v0.25.0 (/Users/matt/Projects/clarinet)
│   │       │       └── deno v1.11.0 (/Users/matt/Projects/clarinet/vendor/deno/cli)
│   │       │           └── clarinet v0.25.0 (/Users/matt/Projects/clarinet)
│   │       └── segment v0.1.2
│   │           └── clarinet v0.25.0 (/Users/matt/Projects/clarinet)
│   ├── reqwest v0.11.4 (*)
│   └── tokio-native-tls v0.3.0
│       ├── hyper-tls v0.5.0 (*)
│       └── reqwest v0.11.4 (*)
└── openssl v0.10.35
    └── native-tls v0.2.7 (*)

```

PR #264, opened 2022-02-18
feat: stacks-devnet-js support for linux-musl (non-glibc, e.g. alpine)

zone117x:
Build `stacks-devnet-js` for linux-x64-musl.

Also fixes https://github.com/hirosystems/clarinet/issues/263

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -134,45 +134,24 @@ jobs:
           - os: ubuntu-latest
             platform: linux
             target: x86_64-unknown-linux-gnu
-            linker_package:
-            cc: ""
             architecture: x64
-          # To be uncommented later when adding new distros
-          # - os: ubuntu-latest
-          #   platform: linux
-          #   target: aarch64-unknown-linux-gnu
-          #   linker_package: gcc-aarch64-linux-gnu
-          #   cc: aarch64-linux-gnu-gcc
-          #   architecture: arm64
-          # - os: ubuntu-latest
-          #   platform: linux
-          #   target: armv7-unknown-linux-gnueabihf
-          #   linker_package: gcc-arm-linux-gnueabihf
-          #   cc: arm-linux-gnueabihf-gcc
-          #   architecture: armv7l
-          # - os: ubuntu-latest
-          #   platform: linux
-          #   target: x86_64-unknown-linux-musl
-          #   linker_package:
-          #   cc: musl-gcc
-          #   architecture: x64
+            libc: glibc
+          - os: ubuntu-latest
+            platform: linux
+            target: x86_64-unknown-linux-musl
+            architecture: x64
+            libc: musl
           - os: windows-latest
             platform: windows
             target: x86_64-pc-windows-msvc
-            linker_package:
-            cc: ""
             architecture: x64
           - os: macos-latest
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -248,10 +238,24 @@ jobs:
...
+        uses: docker://rust:alpine3.15
+        env:
+          RUSTFLAGS: -C target-feature=-crt-static
+        with:
+          entrypoint: /bin/sh
+          args: -c "
+            cd node-bindings &&
+            apk add alpine-sdk nodejs npm git &&
+            npm install --ignore-scripts &&
+            npm run build-${{ env.SHORT_TARGET_NAME }}"
```
zone117x, line 258: @CharlieC3 this looks appalling, but was the only approach I could get working. Let me know if you have any suggestions on how to improve this.

Cargo.toml
```diff
@@ -26,7 +26,8 @@ deno_core = { path = "./vendor/deno/core", optional = true }
 deno_runtime = { path = "./vendor/deno/runtime", optional = true }
 deno = { path = "./vendor/deno/cli", optional = true }
 # clarity_repl = { package = "clarity-repl", path = "../../clarity-repl", features = ["cli"] }
-clarity_repl = { package = "clarity-repl", version = "=0.22.1" }
+# clarity_repl = { package = "clarity-repl", version = "=0.22.0" }
+clarity-repl = { git = "https://github.com/hirosystems/clarity-repl", branch = "develop" }
```
zone117x, line 30: Waiting on https://github.com/hirosystems/clarity-repl/pull/99 to be released

.github/workflows/ci.yaml
CharlieC3: Curious what happens if you use the `cross` tool to compile on the `musl` target in the normal `Build - Node` step? [It would look something like this](https://gist.github.com/Thearas/8a656c2e3783ec1a623146f672e21ee3).

zone117x: That tool is wrapper for a script pretty similar to the above, it wraps some docker commands. In this case, it wouldn't really help much because we'd still need run scripts within cross's docker image to install nodejs. 

Also, I don't think it would work even after that, because their image is still doing cross-compilation from an ubuntu base image (https://github.com/cross-rs/cross/blob/master/docker/Dockerfile.x86_64-unknown-linux-musl) using musl-cross tooling, which runs into various problems outputting the type of library we're building for nodejs integration. The only way I was able to get it to work was building from a musl environment (like alpine).

lgalabru (dismissed):
thank you @zone117x, i'm in favor of testing this new CI steps in the next release!
Can we mark your PR as "Ready for review" ?

obycode:
The repl release in progress right now should be 0.23.0 - https://github.com/hirosystems/clarity-repl/actions/runs/1889532602.

Cargo.toml
obycode: Realized my other comment should have been here... You can update this to 0.23.0, which was just released. Thanks!

obycode (approved):
Thanks @zone117x!

CharlieC3 (approved):
LGTM! :shipit:

PR #267, opened 2022-02-21
Docs for the check checker

obycode:


README.md
```diff
@@ -175,6 +174,151 @@ $ clarinet check
...
+```clarity
+(define-public (dangerous (amount uint))
+    (let ((sender tx-sender))
+        ;; #[allow(unchecked_data)]
+        (as-contract (stx-transfer? amount tx-sender sender))
+    )
+)
+```
+
+**`#[filter(var1, var2)]`**
```
lgalabru, line 308: What's your take on tagging this annotation as the **recommended** way to deal with warnings?

lgalabru (approved):

README.md
obycode: I like that. How about I add this line below the description?

*This is the safest and preferred way to silence warnings that you consider false positives.*

lgalabru: 👍

lgalabru (approved):
Thanks @obycode!

PR #270, opened 2022-02-23
CLI improvements

obycode:
* Cleanup commands
* Improve help documentation
* Add a command to generate shell completion scripts
* Commit the current version of shell completion scripts


completions/README.md
```diff
@@ -0,0 +1,14 @@
...
+- **bash**: *clarinet.bash*
+- **elvish**: *clarinet.elv*
+- **fish**: *clarinet.fish*
+- **powershell**: *_clarinet.ps1*
+- **zsh**: *_clarinet*
+
+These files are automatically generated by `clap` based on the arguments structs in *src/frontend/cli.rs*. To regenerate them after making changes, use:
+
+```sh
+clarinet completions (bash|elvish|fish|powershell|zsh)
```
lgalabru, line 13: Should we be writing this documentation assuming that most of developers will be installing clarinet through a package manager? When this is the case, the `clarinet.*` completion files won't be provided during the installation, developers will have to generate them locally, right?

lgalabru (approved):
This is awesome, thanks @obycode!

completions/README.md
obycode: I think this specific README can stay as-is, since it will only be seen by users that have downloaded the source. Then I'll add a note in the main README for users installing through package managers, describing how they should generate these files.

obycode: Added in 5ffddff3d9ea40559c125de26a821e84edc5c809. 

@lgalabru can you please review that change as well and let me know if you think that solves the problem you were thinking of?

lgalabru (approved):
Looks great, thanks!

PR #274, opened 2022-02-24
docs: add documentation on clarinet test

### LINKED ISSUE #132: `clarinet test` requires test file to include `_test` in name
When adding a new test file to your `clarinet` project, if the test file name does not end with `_test` then `clarinet test` will fail with error `No matching test modules found`. Normally test frameworks target all files in a folder like "tests" for inclusion and ignore filename conventions. Assuming there is good reason to force the `_test` naming heuristic, could we at least update the documentation to caution users that this convention is required?

obycode:
This commit adds more information about unit testing with clarinet into
the README.

Resolves #132


lgalabru (approved):
LGTM!

PR #294, opened 2022-03-24
Doc: Use tar.gz in the pre-built binaries install options

empeje:
I think in the releases page only tar.gz available for the binaries

image

So I create a snippets for people to easily install it

e.g. https://github.com/StackerDAOs/backend/blob/dev/bin/install_clarinet.sh


obycode (approved):
Good catch. Thanks @empeje!

lgalabru (approved):

empeje:
Thanks for accepting the PR @lgalabru and @obycode

Issue #297, opened 2022-03-29
Re-enable code coverage in CI


lgalabru:
Code coverage through `grcov` recently break. After quick convo with @obycode, we will disable this for v0.28.0 and revisit the approach after the release.

lgalabru:
Addressed in v0.28.0

PR #301, opened 2022-03-30
fix: repair test CI workflow

lgalabru:
Address https://github.com/hirosystems/clarinet/issues/297, by installing grcov on top of a stable toolchain, then switch to nightly toolchain for producing the coverage informations. Add 13s to the workflow on a non critical path, which sounds reasonable.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -71,11 +71,10 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Install Rust toolchain
+      - name: Install Rust toolchain stable
         uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
-          target: ${{ matrix.target }}
           profile: minimal
           override: true
 
@@ -95,9 +94,17 @@ jobs:
         if: steps.cache-cargo.outputs.cache-hit != 'true'
         run: RUSTC_BOOTSTRAP=1 cargo install grcov
 
+      - name: Install Rust toolchain nightly
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly-2022-03-28
+          profile: minimal
+          components: llvm-tools-preview
+          override: true
+
       - name: Unit Tests
         env:
-          RUSTFLAGS: "-Cinstrument-coverage"
+          RUSTFLAGS: "-Zinstrument-coverage"
           LLVM_PROFILE_FILE: "clarinet-%p-%m.profraw"
         run: cargo build --locked && cargo test
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -97,7 +97,7 @@ jobs:
       - name: Install Rust toolchain nightly
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: nightly-2022-03-28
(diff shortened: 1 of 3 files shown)


obycode (dismissed):
LGTM. Nice to have that fixed so quickly. Thanks @lgalabru!

.github/workflows/ci.yaml
```diff
@@ -95,9 +94,17 @@ jobs:
         if: steps.cache-cargo.outputs.cache-hit != 'true'
         run: RUSTC_BOOTSTRAP=1 cargo install grcov
 
+      - name: Install Rust toolchain nightly
```
CharlieC3, line 97: Why install Rust stable if we then install Rust nightly over it? Wouldn't it be better to just revert to installing Rust nightly at the beginning?

CharlieC3 (commented):
For posterity, Rust stable fails to generate code coverage reports because it doesn't ship with the `profiler` runtime by default (but nightly does).
https://github.com/rust-lang/rust/blob/master/config.toml.example#L296

If we want to get around using nightly, we'd likely need to create a custom toolchain of Rust ourselves.
https://rust-lang.github.io/rustup/concepts/toolchains.html#custom-toolchains

.github/workflows/ci.yaml
lgalabru: We install stable, compile grcov with this toolchain, then install nightly to build clarinet with profiling enabled.
I tried what you suggested, using the profiler enabled in the Cargo.toml with stable but ended up with an empty lcov file, so ended up keeping the nightly approach.

obycode (approved):
👍

.github/workflows/ci.yaml
CharlieC3: @lgalabru But it doesn't look like that's what's happening here:
https://github.com/hirosystems/clarinet/blob/48e8f62bdb67518328725b43912eccd334e22f92/.github/workflows/ci.yaml#L74-L103

Rust stable is being installed, cache restored, grcov installed, then Rust nightly is being installed over Rust stable, _then_ we compile and test. I think this would be equivalent to installing Rust nightly in the beginning instead of Rust stable, wouldn't it?

lgalabru: when we invoke the command `cargo install grcov` we're not pulling a pre-built binary, grcov is being compiled with the stable toolchain that just installed.

CharlieC3: @lgalabru Correct, so why not install just `grcov` onto Rust nightly? What are we gaining by having this circuitous approach?

lgalabru: Feel free to revisit this flow! Getting the CI back on track ASAP was important for increasing our confidence in the upcoming release, if we could just stop relying on nightly that'd be great.

CharlieC3: > we recently switched to rust `edition2021` and had to upgrade our version of rust + rust nightly and installing `grcov` using more recent nightlies was failing (https://github.com/hirosystems/clarinet/runs/5726898694?check_suite_focus=true).

Oh gotcha, yeah if it fails to install grcov using Rust nightly I can totally see why this is necessary.

PR #305, opened 2022-03-30
fix: adjust some devnet env variables

lgalabru:
Some last minute tweaks: 
- fixing an issue with the explorer not displaying blocks / transactions (devnet is definitely not a 1st class citizen on the explorer side - would be great to get a round of QA there - cc @He1DAr)
- displaying the /doc endpoint in the console, and enabling it (thanks @zone117x @rafaelcr!)
Screen Shot 2022-03-30 at 5 11 45 PM


obycode (approved):
LGTM

PR #309, opened 2022-04-01
fix: incorrect costs synthesis

### LINKED ISSUE #133: [Feature] Costs report by default should take into account only success function calls
`clarinet test --costs` is great, but it takes into account all function calls including failed ones. If we have tests that verify all execution paths we might get a false impression that our contracts are super cheap and efficient while in reality they are not.

Example:
**Report generated with only 1 test that verify green path:**
```
+-----------------------------------+------------------+------------+---------------------+-------------+----------------------+--------------+
|                                   | Runtime (units)  | Read Count | Read Length (bytes) | Write Count | Write Length (bytes) | Tx per Block |
+-----------------------------------+------------------+------------+---------------------+-------------+----------------------+--------------+
| main::stack-tokens                | 28962000 (0.58%) | 47 (0.61%) |        5981 (0.01%) |  22 (0.28%) |         3443 (0.02%) |          164 |
+-----------------------------------+------------------+------------+---------------------+-------------+----------------------+--------------+
|                                                                                                                                             |
+-----------------------------------+------------------+------------+---------------------+-------------+----------------------+--------------+
| Mainnet Block Limits (Stacks 2.0) |       5000000000 |       7750 |           100000000 |        7750 |             15000000 |            / |
+-----------------------------------+------------------+------------+---------------------+-------------+----------------------+--------------+
```
**Report generated using 3 tests: 1 green path and 2 red path:**
```
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
|                                   | Runtime (units) | Read Count | Read Length (bytes) | Write Count | Write Length (bytes) | Tx per Block |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
| main::stack-tokens                | 2623000 (0.05%) |  3 (0.04%) |        2538 (0.00%) |           0 |                    0 |         1906 |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
|                                                                                                                                            |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
| Mainnet Block Limits (Stacks 2.0) |      5000000000 |       7750 |           100000000 |        7750 |             15000000 |            / |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
```

I think `clarinet test --costs` should generate report only for positive function calls. 
And when we add flag `--all` or replace `--costs` with `--costs-all` it will generate report using all.

### LINKED ISSUE #250: [BUG] `clarinet test --costs` generates 2 or more different execution costs when function is tested by 2 or more tests
When function is covered by 2 or more tests, execution costs calculated by `clarinet test --costs` are calculated only by one test and randomly printed to stdout.

**Steps To Reproduce**
1. `clarinet new vrf-test`
2. `cd vrf-test`
3. `clarinet contract new vrf`
4. paste following code into contract file:
```clojure
(define-constant HEX_BUFF 0x000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f202122232425262728292a2b2c2d2e2f303132333435363738393a3b3c3d3e3f404142434445464748494a4b4c4d4e4f505152535455565758595a5b5c5d5e5f606162636465666768696a6b6c6d6e6f707172737475767778797a7b7c7d7e7f808182838485868788898a8b8c8d8e8f909192939495969798999a9b9c9d9e9fa0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebfc0c1c2c3c4c5c6c7c8c9cacbcccdcecfd0d1d2d3d4d5d6d7d8d9dadbdcdddedfe0e1e2e3e4e5e6e7e8e9eaebecedeeeff0f1f2f3f4f5f6f7f8f9fafbfcfdfeff)

(define-constant ERR_FAIL (err 1000))

;; VRF functions
(define-map BlockVRF uint uint)

;; #[allow(unchecked_data)]
(define-public (get-and-save-rnd (block uint))
  (ok (unwrap! (get-or-create-rnd-uint block true) ERR_FAIL))
)

(define-private (get-or-create-rnd-uint (block uint) (save bool))
  (match (map-get? BlockVRF block) 
    rnd (some rnd)  
    (match (match (get-block-info? vrf-seed block)
    vrfSeed (some (get-rnd-from-vrf-seed vrfSeed))
    none
  )
      rnd
      (begin
        (and save (map-set BlockVRF block rnd))
        (some rnd)
      )
      none
    )
  )
)


(define-read-only (get-rnd (block uint))
  (match (get-block-info? vrf-seed block)
    vrfSeed (some (get-rnd-from-vrf-seed vrfSeed))
    none
  )
)

(define-read-only (get-rnd-from-vrf-seed (vrfSeed (buff 32)))
  (+
    (convert-to-le (element-at vrfSeed u16) u15)
    (convert-to-le (element-at vrfSeed u17) u14)
    (convert-to-le (element-at vrfSeed u18) u13)
    (convert-to-le (element-at vrfSeed u19) u12)
    (convert-to-le (element-at vrfSeed u20) u11)
    (convert-to-le (element-at vrfSeed u21) u10)
    (convert-to-le (element-at vrfSeed u22) u9)
    (convert-to-le (element-at vrfSeed u23) u8)
    (convert-to-le (element-at vrfSeed u24) u7)
    (convert-to-le (element-at vrfSeed u25) u6)
    (convert-to-le (element-at vrfSeed u26) u5)
    (convert-to-le (element-at vrfSeed u27) u4)
    (convert-to-le (element-at vrfSeed u28) u3)
    (convert-to-le (element-at vrfSeed u29) u2)
    (convert-to-le (element-at vrfSeed u30) u1)
    (convert-to-le (element-at vrfSeed u31) u0)
  )
)

(define-private (convert-to-le (byte (optional (buff 1))) (pos uint))
  (* (buff-to-u8 (unwrap-panic byte)) (pow u2 (* u8 pos)))
)

(define-read-only (buff-to-u8 (byte (buff 1)))
  (unwrap-panic (index-of HEX_BUFF byte))
)
```
5. paste following code into test file
```ts

import { Clarinet, Tx, Chain, Account, Contract, types } from 'https://deno.land/x/clarinet@v0.25.0/index.ts';
import { assertEquals } from 'https://deno.land/std@0.90.0/testing/asserts.ts';

Clarinet.test({
    name: "[VRF] get-rnd() returns (some uint) for blocks from the past",
    async fn(chain: Chain, accounts: Map<string, Account>, contracts: Map<string, Contract>) {
      const deployer = accounts.get("deployer")!;
  
      chain.mineEmptyBlock(200);
      for (let b = 100; b > 80; b--) {
        const response = chain.callReadOnlyFn("vrf", "get-rnd", [types.uint(b)], deployer.address);
  
        response.result.expectSome();
      }
    },
  });
  
  Clarinet.test({
    name: "[VRF] get-rnd() returns none for current block and blocks from the future",
    async fn(chain: Chain, accounts: Map<string, Account>, contracts: Map<string, Contract>) {
      const deployer = accounts.get("deployer")!;
  
      for (let b = 200; b < 220; b++) {
        const response = chain.callReadOnlyFn("vrf", "get-rnd", [types.uint(b)], deployer.address);
  
        response.result.expectNone();
      }
    },
  });
```
7. execute `clarinet test --costs` wait 2-3sec, and execute again (do it multiple times)

Most of the time it prints costs calculated based on 1st test:
```
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
|                                   | Runtime (units) | Read Count | Read Length (bytes) | Write Count | Write Length (bytes) | Tx per Block |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
| vrf::get-rnd                      |   62911 (0.00%) |  4 (0.05%) |        2355 (0.00%) |           0 |                    0 |         1937 |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
|                                                                                                                                            |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
| Mainnet Block Limits (Stacks 2.0) |      5000000000 |       7750 |           100000000 |        7750 |             15000000 |            / |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
```
but sometimes it prints costs calculated only using 2nd test:
```
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
|                                   | Runtime (units) | Read Count | Read Length (bytes) | Write Count | Write Length (bytes) | Tx per Block |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
| vrf::get-rnd                      |    9404 (0.00%) |  4 (0.05%) |        2355 (0.00%) |           0 |                    0 |         1937 |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
|                                                                                                                                            |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
| Mainnet Block Limits (Stacks 2.0) |      5000000000 |       7750 |           100000000 |        7750 |             15000000 |            / |
+-----------------------------------+-----------------+------------+---------------------+-------------+----------------------+--------------+
```

**Expected behavior**
`clarinet test --costs` should produce coherent output. Either one per test, or one calculated base on multiple tests.

**Environment**
I tested it using Clarinet v0.24.0, v0.25.0 and v0.25.1

**Additional context**
It is not the same as #133 but combination of these 2 issues makes the `clarinet test--costs' completely unreliable.

LNow:
This PR introduces new way of presenting execution costs summary.

Cost of each contract is presented in separate table.

image

fix #133, fix #250

### DIFF
--- a/src/runnner/deno.rs
+++ b/src/runnner/deno.rs
@@ -1,9 +1,10 @@
 use clarity_repl::clarity::coverage::CoverageReporter;
 use clarity_repl::clarity::types;
 use clarity_repl::clarity::util::hash;
+use clarity_repl::prettytable::format::{FormatBuilder, LinePosition, LineSeparator};
 use clarity_repl::prettytable::{color, format, Attr, Cell, Row, Table};
 use clarity_repl::repl::session::CostsReport;
-use clarity_repl::repl::Session;
+use clarity_repl::repl::{CostSynthesis, Session};
 use deno::ast;
 use deno::colors;
 use deno::create_main_worker;
@@ -551,12 +552,66 @@ struct FunctionStats {
     pub write_length: Vec<(u64, u64)>,
 }
 
+#[derive(Debug)]
+struct CallStats {
+    pub runtime: u64,
+    pub runtime_limit: u64,
+    pub read_count: u64,
+    pub read_count_limit: u64,
+    pub read_length: u64,
+    pub read_length_limit: u64,
+    pub write_count: u64,
+    pub write_count_limit: u64,
+    pub write_length: u64,
+    pub write_length_limit: u64,
+    pub tx_per_block: u64,
+}
+
+fn safe_div(limit: u64, total: u64) -> u64 {
+    limit.checked_div(total).unwrap_or_else(|| limit)
+}
+
+impl CallStats {
+    #[must_use]
+    fn from_cost_report(costs: CostSynthesis) -> Self {
+        let limit = costs.limit;
+        let total = costs.total;
+
+        let mut tx_per_block_limits = vec![
+            safe_div(limit.runtime, total.runtime),
+            safe_div(limit.read_count, total.read_count),
+            safe_div(limit.read_length, total.read_length),
+            safe_div(limit.write_count, total.write_count),
+            safe_div(limit.write_length, total.write_length),
+        ];
+
+        tx_per_block_limits.sort_by(|a, b| a.cmp(&b));
+        let tx_per_block = tx_per_block_limits.first().unwrap();
+
+        Self {
+            runtime: total.runtime,
+            runtime_limit: limit.runtime,
+            read_count: total.read_count,
+            read_count_limit: limit.read_count,
+            read_length: total.read_length,
+            read_length_limit: limit.read_length,
+            write_count: total.write_count,
+            write_count_limit: limit.write_count,
+            write_length: total.write_length,
+            write_length_limit: limit.write_length,
+            tx_per_block: *tx_per_block,
+        }
+    }
+}
+
 fn display_costs_report() {
     let mut consolidated: BTreeMap<String, BTreeMap<String, Vec<CostsReport>>> = BTreeMap::new();
     let sessions = sessions::SESSIONS.lock().unwrap();
     let mut mins: BTreeMap<(&String, &String), (f32, CostsReport, Bottleneck)> = BTreeMap::new();
     let mut maxs: BTreeMap<(&String, &String), (f32, CostsReport, Bottleneck)> = BTreeMap::new();
     let mut stats: BTreeMap<(&String, &String), FunctionStats> = BTreeMap::new();
+
+    let mut contracts_stats: BTreeMap<&String, BTreeMap<&String, Vec<CallStats>>> = BTreeMap::new();
 
     for (session_id, (name, session)) in sessions.iter() {
         for report in session.costs_reports.iter() {
@@ -627,6 +682,13 @@ fn display_costs_report() {
             );
 
             let key = (&report.contract_id, &report.method);
+
+            let contract_stats = contracts_stats
+                .entry(&report.contract_id)
+                .or_insert(BTreeMap::new());
+            let call_stats = contract_stats.entry(&report.method).or_insert(Vec::new());
+
+            call_stats.push(CallStats::from_cost_report(report.cost_result.clone()));
 
             let st = stats.entry(key).or_insert(FunctionStats::default());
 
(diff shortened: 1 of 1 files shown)


LNow:
For each function call I calculate how many TX with exactly the same call you can fit into one block:https://github.com/hirosystems/clarinet/blob/62d3c14f2d0ec99323aacaf5df6c489e7ab7b65a/src/runnner/deno.rs#L594-L603

Then I sort these costs fore each function in order to extract the least and most expensive call:
https://github.com/hirosystems/clarinet/blob/62d3c14f2d0ec99323aacaf5df6c489e7ab7b65a/src/runnner/deno.rs#L685-L688

So I think bottleneck identification is still preserved, as most expensive call is your bottleneck.

With regards to average values, personally I find them very useful if you have function that can have different execution costs based on the inputs you provide to it. An example of such function can be `uint-to-ascii`: https://github.com/LNow/clarity-projects/blob/main/contracts/utility/uint-to-ascii.clar
With unit tests that exhaust whole range of inputs: https://github.com/LNow/clarity-projects/blob/main/tests/utility/uint-to-ascii.test.ts I can find out what is the average costs of using it:
image
That way I don't have to calculate it manually.

lgalabru:
I think average would be useful if you were including only one instance of a test in that average:
As in, if in your test suite, you have 8 contract-call `(sum 1 1)` and 1 contract-call `(sum 1832901238901238102381029380129038102939123810293 1832901238901238102381029380129038102939123810293)` you probably to average this 2 distinct cases once, instead of diluting test case 2 with 8 instances of test case 1.
Is that already the case in your implementation?

LNow:
No, unfortunately it is a very naive implementation.

lgalabru:
We could keep that implementation naive, and just introduce a hash map for collecting the results, where the key being used involves the arguments being used?

LNow:
Interesting idea 🤔 
I'll try to implement it, but also I'd like to keep current solution as well.

Some functions can have different execution costs based on who's called them, at what block, what was the state of the contract storage, what values are returned by other contracts etc. Even when you call them with exactly the same arguments. Because of that I want to keep original solution.

LNow:
@lgalabru unfortunately I failed to implement your idea, so it has to stay the way it is right now or someone else have to add it.

lgalabru:
Sounds good. Per your comment in the PR description, your work would fix https://github.com/hirosystems/clarinet/issues/133.
Do you mind writing a quick test, or at least explain what was the problem? Also were you able to confirm that this new PR is improving the consistency with the expectations that @whoabuddy unpacked here https://github.com/hirosystems/clarinet/issues/133#issuecomment-1024453703? 
thank you for your help on that front!

LNow:
Probably changing this line was the root cause:
https://github.com/hirosystems/clarinet/blob/6a83d9dda263aa20c8d88dc58cf6b675e4d90f03/src/runnner/deno.rs#L658
And changing it to:
```rust
let (max, max_report, max_bottleneck) = maxs.get(&(contract_id, method)).unwrap();
```
would solve the problem.
I say "probably", because I couldn't understand significant portion of your code, and that is why I wrote this almost from scratch.

> Also were you able to confirm that this new PR is improving the consistency with the expectations that @whoabuddy unpacked here https://github.com/hirosystems/clarinet/issues/133#issuecomment-1024453703?

Yes, this PR provides correct results. 
They are not exactly the same (see example below for just one call), but I think it is because in console costs of `(contract-call? )` is added to the summary.
```
+----------------------+----------+------------+
|                      | Consumed | Limit      |
+----------------------+----------+------------+
| Runtime              | 13824023 | 5000000000 |
+----------------------+----------+------------+
| Read count           | 2415     | 7750       |
+----------------------+----------+------------+
| Read length (bytes)  | 175844   | 100000000  |
+----------------------+----------+------------+
| Write count          | 604      | 7750       |
+----------------------+----------+------------+
| Write length (bytes) | 91948    | 15000000   |
+----------------------+----------+------------+

vs

┌────────────────────────────────────────┬──────────┬────────────────────────────┬─────────────────────────┬──────────────────────────┬───────────────────────┬─────────────────────────┬──────────────┐
│                                                                                                                                                                                                      │
│ ✨  ST1HTBVD3JG9C05J7HBJTHGR0GGW7KXW28M5JS8QE.citycoin-core-v1                                                                                                                                       │
│                                                                                                                                                                                                      │
├────────────────────────────────────────┼──────────┼────────────────────────────┼─────────────────────────┼──────────────────────────┼───────────────────────┼─────────────────────────┼──────────────┤
│                                        │ # Calls  │ Runtime (units)            │ Read Count              │ Read Length (bytes)      │ Write Count           │ Write Length (bytes)    │ Tx per Block │
├────────────────────────────────────────┼──────────┼────────────────────────────┼─────────────────────────┼──────────────────────────┼───────────────────────┼─────────────────────────┼──────────────┤
│ mine-many                              │       16 │  min:     38728  (0.001%)  │  min:     4   (0.052%)  │  min:   29905  (0.030%)  │  min:    0  (0.000%)  │  min:      0  (0.000%)  │         1937 │
│                                        │          │  max:  13814911  (0.276%)  │  max:  2411  (31.110%)  │  max:  175827  (0.176%)  │  max:  601  (7.755%)  │  max:  91601  (0.611%)  │            3 │
│                                        │          │  avg:   1044307  (0.021%)  │  avg:   181   (2.335%)  │  avg:   40614  (0.041%)  │  avg:   43  (0.555%)  │  avg:   6552  (0.044%)  │          504 │
├────────────────────────────────────────┼──────────┼────────────────────────────┼─────────────────────────┼──────────────────────────┼───────────────────────┼─────────────────────────┼──────────────┤

lgalabru (approved):
LGTM

PR #316, opened 2022-04-08
docs: change clarinet deploy to clarinet contract publish

mcintyre94:
- `clarinet deploy` doesn't seem to exist any more, and seems to have been replaced by `clarinet contract publish`
- this PR updates the README + some CLI output to use the updated command

### DIFF
--- a/README.md
+++ b/README.md
@@ -420,7 +420,7 @@ You can use Clarinet to deploy your contracts to the public Testnet environment 
 evaluation on a blockchain. Use the following command:
 
 ```bash
-$ clarinet deploy --testnet
+$ clarinet contract publish --testnet
 ```
 
 ### Use Clarinet in your CI workflow as a GitHub Action
--- a/src/frontend/cli.rs
+++ b/src/frontend/cli.rs
@@ -985,13 +985,13 @@ fn display_deploy_hint() {
         yellow!("Once your contracts are ready to be deployed, you can run the following:")
     );
 
-    println!("{}", blue!("  $ clarinet deploy --testnet"));
+    println!("{}", blue!("  $ clarinet contract publish --testnet"));
     println!(
         "{}",
         yellow!("    Deploy all contracts to the testnet network.\n")
     );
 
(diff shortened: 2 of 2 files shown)


sabbyanandan:
@mcintyre94: Thanks for bringing this up! 

Depending on the version of Clarinet, there's a bit of UX difference. In the 0.27.x line, you'd see a `clarinet contract <subcommand>`. However, in the latest release (0.28.x), this is now refactored to `clarinet contracts <subcommand>`. Only in the latest release, the `contracts` (in plural) command includes an option to `publish` the contract to different chains.

@lgalabru @obycode: I couldn't quickly find from the commit history when the `clarinet deploy ..` was deprecated/removed. Could one of you please point to the proper use of the command? We could take this opportunity also to update any other references in README/docs.

README.md
```diff
@@ -420,7 +420,7 @@ You can use Clarinet to deploy your contracts to the public Testnet environment
 evaluation on a blockchain. Use the following command:
 
 ```bash
-$ clarinet deploy --testnet
+$ clarinet contract publish --testnet
```
sabbyanandan, line 423: In 0.28.x, this should be:

> $ clarinet contracts publish ...

lgalabru: Good catch, thanks.
Both `clarinet contract publish --network` and `clarinet contracts publish --network` should work for deploying contracts. Depending on the action (new, publish, etc) being sent to the subcommand contract/contracts, using the singular or plural might be more or less intuitive, which is why I created an alias `contract` for the `contracts` subcommand.
We moved `clarinet publish` to `clarinet contracts publish` because we are about to introduce a new type of offchain components that will also be published / deploy.   
This change went live with 0.28.

lgalabru: https://github.com/hirosystems/clarinet/blob/7e57cd2e0202b90a0d49b006c6407f815b4c24c9/src/frontend/cli.rs#L124

lgalabru (approved):
LGTM, thank you @mcintyre94!

PR #325, opened 2022-04-20
Update to clarity-repl 0.26.0

obycode:
Removes contract ordering from clarinet and fixes a bug in the LSP.

Note - this will not work as-is until clarity-repl 0.26.0 release is completed.


lgalabru (approved):
🚀

PR #330, opened 2022-04-22
fix: fixed problem with contract ordering in lsp

obycode:
This adds the new contract ordering logic to the lsp, which fixes a bug exposed by the previous changes.


lgalabru (approved):
LGTM, thanks @obycode!

PR #340, opened 2022-05-04
feat: introducing hooks

lgalabru:


AlgorithmGG:
Looks you guys are having fun, what is good?

PR #343, opened 2022-05-09
Update README.md

jhammond2012:
Seems that "contracts" should only be "contract" here.


lgalabru (approved):
LGTM! Thanks @jhammond2012

PR #345, opened 2022-05-09
WIP: Add support for DAP

obycode:
This is a WIP until a version of clarity-repl is released with DAP support (see hirosystems/clarity-repl#142).

PR #347, opened 2022-05-11
fix: deployments / unordered contracts

lgalabru:


obycode:
The errors from CI on this are fixed in #345, so let's ignore those.

src/publish/mod.rs
```diff
@@ -210,21 +211,39 @@ pub fn publish_all_contracts(
...
+                .interpreter
+                .build_ast(contract_id.clone(), initial_contract.code.clone(), 2);
+        contract_asts.insert(contract_id.clone(), ast);
+    }
+
+    let result = ASTDependencyDetector::detect_dependencies(&contract_asts, &BTreeMap::new());
+    let dependencies = match result {
+        Ok(deps) => deps,
+        Err(_) => {
+            println!("Deployments in presence of requirements is not supported yet.");
```
obycode, line 235: You should still be able to deploy to mainnet in the presence of requirements, just not to devnet, right? Although since this will be fixed shortly with #333, maybe we can leave this simplification.

obycode (approved):
LGTM. Thanks for the quick fix!

obycode:
Confirmed: everything works when rebased on top of `feat/dap` branch.

src/publish/mod.rs
obycode: Could we check here if `network` is `Devnet` and only report this error and exit in that case? Or am I missing something?

lgalabru: I think even a Testnet / Mainnet deployment, that involves some requirements would require a special effort for handling the dependencies checking, no?

PR #352, opened 2022-05-13
fix: use `PathBuf` for paths

obycode:
This change corresponds with hirosystems/clarity-repl@2f1b676 to
switch from using a `String` to using a `PathBuf` for storing the
contract path so that it can be properly compared on Windows.


obycode:
Test failures are due to dependency on clarity-repl release with corresponding changes.

obycode:
This change is already integrated into #333.

PR #368, opened 2022-05-23
feat: support for hyperchain on devnet

lgalabru:
In this PR, we're exploring support for Hyperchain on Devnet, and trying to remove all the friction from developers plates.
With the API, the Explorer and Stacks.js not supporting this new layer, there's not much that can be done, but shipping this first version could be used for testing these other components.


obycode (approved):
LGTM, just one clarification in the documentation.

moodmosaic:
🚀

PR #376, opened 2022-05-25
fix: code coverage not including initial executions

lgalabru:
Regression introduced in v0.31.0
Screen Shot 2022-05-25 at 14 10 59

To be fixed in v0.31.1
Screen Shot 2022-05-25 at 14 11 05


obycode (approved):
LGTM!

PR #377, opened 2022-05-25
fix: display errors from parsing

lgalabru:


obycode (approved):
Thanks for the quick fix!

PR #386, opened 2022-05-26
fix: update repl, better handling of default plans

lgalabru:


src/deployment/mod.rs
```diff
@@ -1055,7 +1078,10 @@ pub fn generate_default_deployment(
         transactions.push(tx);
     }
 
-    let tx_chain_limit = 25;
+    let tx_chain_limit = match no_batch {
+        true => 100_000,
```
obycode, line 1082: Curious: is this number meaningful, or is it just some big number?

obycode (approved):
LGTM!

PR #390, opened 2022-05-31
Fix alias for clarinet deployment

mcintyre94:
When I run the commands in README I get an error eg:

```
$ clarinet deployment generate --mainnet
error: The subcommand 'deployment' wasn't recognized

        Did you mean 'deployments' or 'deployments'?
...
```

This commit updates the alias for `deployments` to `deployment` so that these commands work

### DIFF
--- a/src/frontend/cli.rs
+++ b/src/frontend/cli.rs
@@ -105,7 +105,7 @@ enum Requirements {
 }
 
 #[derive(Subcommand, PartialEq, Clone, Debug)]
-#[clap(bin_name = "deployment", aliases = &["deployments"])]
+#[clap(bin_name = "deployment", aliases = &["deployment"])]
 enum Deployments {
     /// Check deployments format
     #[clap(name = "check", bin_name = "check")]


README.md
```diff
@@ -449,13 +449,13 @@ You can use Clarinet to publish your contracts to Devnet / Testnet / Mainnet env
...
 ```bash
-$ clarinet deployment generate --mainnet
+$ clarinet deployments generate --mainnet
 ```
 
 After **cautiously** reviewing (and updating if needed) the generated plan, you can use the command:
 
 ```bash
-$ clarinet deployment apply -p <path-to-plan.yaml>
+$ clarinet deployments apply -p <path-to-plan.yaml>
```
lgalabru, line 458: Thanks @mcintyre94!
How would you feel about fixing the line
https://github.com/hirosystems/clarinet/blob/a8e746096dfda411d96627f82c500c737ddf149b/src/frontend/cli.rs#L108
instead?
In `aliases`, we should have `deployment` instead of `deployments`. Supporting both singular and plural is just more convenient in terms of DX.

mcintyre94: Ah, that's what was going on! The "Did you mean 'deployments' or 'deployments'?" message should have clued me in :) Much better fix, updated the PR - TY!

lgalabru: Awesome, thank you @mcintyre94!

lgalabru (approved):

Issue #396, opened 2022-06-01
Error message is too general when having a typo on contract name 0.31.0


FriendsFerdinand:
When I execute a `contract-call?`  on Clarinet 0.31.0 and the contract name is incorrect, I get the following error after running `clarinet check`:

```
error: unable to detect dependencies
```

I feel as though this is too general. I remember in previous versions this being much more specific. Here's an example repo (it's nothing big, I just make a `contract-call?` from first-contract to second-contract: https://github.com/FriendsFerdinand/dependency-test

first-contract.clar:
```
(define-public (first-call)
  (contract-call? .second-contrac call-this)
)
```

second-contract.clar:
```
(define-public (call-this)
  (begin
    (asserts! true (err u1))
    (ok true)
  )
)
```

PR #400, opened 2022-06-03
fix: improve error messages about dependencies

obycode:
When a dependency cannot be found, just ignore it, because the later
analysis will detect it and report a better error message, complete with
proper location information. Also improve the printing of errors from
the ordering of contracts.

See also: hirosystems/clarity-repl#188
Fixes: #396


obycode:
@lgalabru, I had to manually rebase and force push. please re-approve. Thanks!

lgalabru (approved):

PR #410, opened 2022-06-13
ci: rework CI and release process

CharlieC3:
Reworks the CI pipeline to implement a number of changes.

# Release Changes
* Removes semantic-release and all related files from repo & release flow
* Removes the workflow which opens a PR from `main` to `develop` after a release.
* The new release flow will extract the Clarinet `version` number in the root `Cargo.toml` file, and compare that version against the latest release tag in Github.
    * If they **do not** match, it will perform a release.
    * If they **do** match, it will not release since the `Cargo.toml` version was not updated since the last GH release.
* When performing a release, the workflow will...
    * Tag the created Docker images with the new version
    * Publish Stacks Devnet JS to NPM if the `package.json` is a version higher than what's in NPM
    * Create a new git tag matching the version found in `Cargo.toml` with a prefix of "v" (e.g. v1.2.3)
    * Create a new Github release and upload all the pre-built dists of Clarinet, Stacks Devnet JS, and Oreo
    * Trigger the pkg-version-bump workflow to release on Homebrew and Winget package managers
* The workflow will no longer...
    * Commit back to the repo
    * Write the changelog in the new Github release (This _can_ be automated in the future if you have an expected layout to the `Changelog.md` it can copy from)

# CI Changes
* Oreo distributable and Docker image is now built as part of the CI pipeline
    * Only one Oreo dist is currently being built, but more can easily be added
* CI workflow has more cardinality now, splitting up builds for Clarinet, Stacks Devnet JS, and Oreo into separate jobs.
    * This uses more GH runners, but parallelizes the Windows builds and **reduces total runtime by 15 - 25 minutes on average.**
    * Despite duplicating some code, it's far easier this way to make changes to the workflow and know what pieces of the pipeline will be impacted.
* Pre-built distributables for Clarinet and Oreo are re-used in Docker images
* Clarinet and Oreo Dockerfiles were rewritten to...
    * Use pre-built distributables instead of building in the Docker build process.
    * Use `debian:bullseye` instead of `debian:stretch` base images, since stretch is quite old and is missing Rust dependencies needed to run the pre-built dists.

Closes https://github.com/hirosystems/devops/issues/974
Closes https://github.com/hirosystems/devops/issues/951


.github/workflows/ci.yaml
```diff
@@ -368,116 +336,363 @@ jobs:
...
+        if: matrix.target == 'x86_64-unknown-linux-gnu' && startsWith(github.ref, 'refs/heads/main') && needs.get_release_info.outputs.tag != ''
+        with:
+          package: node-bindings/package.json
+          access: public
+          greater-version-only: true
+          token: ${{ secrets.NPM_TOKEN }}
+
+  dist_oreo:
+    name: Build Oreo Distributions
+    runs-on: ${{ matrix.os }}
```
lgalabru, line 510: I'd be perfectly fine with the idea of just considering unix for Oreo (instead of currently macOS/windows/unix - if my reading is correct)

dockerfiles/Oreo.dockerfile
```diff
@@ -1,33 +1,11 @@
...
-WORKDIR /src/components/orchestra-event-observer
-
-RUN mkdir /out
-
-RUN cargo build --release
-
-RUN cp target/release/orchestra-event-observer /out
-
-FROM debian:stretch-slim
+COPY orchestra-event-observer /bin/orchestra-event-observer
```
lgalabru, line 3: Should we be keeping a dockerfile that would let developers building their own docker image? With the current changes, the only way to build a docker image would be via GitHub, right?
I was thinking about moving all the dockerfiles at play in this repo in the `dockerfiles` folder (including `./Dockerfile`, that would be renamed to `./dockerfiles/Clarinet.dockerfile`.
So we could add a `dockerfiles/ci/Oreo.dockerfile` or something?

.github/workflows/ci.yaml
CharlieC3: I only put down one dist variant for Oreo because I wasn't sure if you wanted to create more like we do for Clarinet. But if you only need one variant, I'll simplify this section a bit.

dockerfiles/Oreo.dockerfile
CharlieC3: > With the current changes, the only way to build a docker image would be via GitHub, right? 

No, it would still be possible to build one locally, but it would not be as straight forward. It would take a few commands now instead of one.

Having two separate Dockerfiles would certainly maintain ease of access to create one locally, but it's good to be aware that this would mean more overhead for anyone updating the build process. Any changes made to CI would likely have to propagate down to these developer images as well to avoid diverging build processes.

Despite this, I still think it's well worth it to maintain both sets of Dockerfiles.

CharlieC3:
> could we shave some more CI time by only considering ubuntu for Build Stacks Devnet JS Distributions, in the context of a normal PR? We could enable the full OS matrix when the PR is a rc/next or a release action?

This _could_ theoretically help speed up builds under specific scenarios where there are no GH runners available for a long time, but the bottleneck is always the Windows build for Clarinet by a large margin. Plus, we have 60 GH runners available between hirosystems repos. So practically you wouldn't experience this benefit except for fringe scenarios.

That being said, using as few GH runners as possible in PR builds is good for the hirosystems org as a whole, since it leaves more GH runners free to be used concurrently elsewhere. So I'll look into making this change anyways if that's ok.

> In our last release, we merged https://github.com/hirosystems/clarity-repl in this repo, and we would need to get the github action being ported in this repo. Could you help us with that?

Sure!

lgalabru (approved):
LGTM, thank you @CharlieC3! Some steps have been stuck for a few hours, is that something we need to look into?

CharlieC3:
@lgalabru Those steps are hard-coded [here](https://github.com/hirosystems/clarinet/settings/branch_protection_rules/21455419). Once this is merged, the branch rules should be updated to use the new steps/jobs.

PR #412, opened 2022-06-14
docs: clarify the usage of `clarinet check`

obycode:
When a file is passed to `clarinet check`, then a syntax-only check is
performed on that one file. A complete check cannot be done when limited
to only one file, since type information for called contracts is not
available.


lgalabru (approved):
LGTM, thanks @obycode!

Issue #418, opened 2022-06-20
Using `try!` as variable name in `let`  should not pass a clarinet check


FriendsFerdinand:
### Description 
Consider the following contract where I use `try!` as a variable name:
```clarity
(define-public (call-this (fail bool))
  (let (
    (try! (try-me fail))
  )
    (ok true)
  )
)

(define-public (try-me (fail bool))
  (begin
    (asserts! fail (err u999))
    (ok true)
  )
)
```
This passes a `clarinet check`, but when executed on `clarinet console` returns a RuntimeError:
`Runtime Error: Runtime error while interpreting ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.contract-7: Unchecked(NameAlreadyUsed("try!"))`

If you use an `asserts!` instead, it will not pass `clarinet check`, naming a syntax error.
If I write: `(define-data-var try! uint u0)`, then it will return a Runtime error `NameAlreadyUsed` after running `clarinet check`.

### Conclusion

I think running `clarinet check` should be able to detect this kind of error before trying to run the contract code on `clarinet console/test`.

obycode:
Thanks for reporting this @FriendsFerdinand.

I see that the type-checker does not include the built-in reserved keywords when it does its checks, but during execution, this is checked. We should be able to just add this check into `ContractContext::check_name_used`.

obycode:
We should also ensure that the unexpected behavior from https://github.com/stacks-network/stacks-blockchain/issues/2696 is unaffected by this change.

lgalabru:
This will be fixed at the canonical vm level, will close this issue.
Feel free to reopen if you see another path forward.

obycode:
This change is not included in 2.1. For the time being, let's report this as an error in Clarinet, and then push to get this included in the next network fork.

smcclellan:
@hugocaillard Has this been resolved upstream?

PR #420, opened 2022-06-21
fix: check for reserved names in let bindings

obycode:
This change checks for reserved names in addition to checking for names
previously defined by the user.

Fixes: #418


obycode:
- [ ] I should also add unit tests for this before merging

moodmosaic:
> * [ ]  I should also add unit tests for this before merging

Yep! And depending on how the test(s) will look like, it might make sense to turn the test(s) into (a single?) [proptest](https://github.com/AltSysrq/proptest#getting-started). (May be tackled on a separate PR.)

lgalabru:
With the VM being pulled from the stacks-blockchain repo this approach does not seem to be viable anymore, so I'll close this PR.

PR #421, opened 2022-06-21
Release Next

lgalabru:


components/clarinet-deployments/src/lib.rs
```diff
@@ -180,7 +180,7 @@ pub async fn generate_default_deployment(
                 }
                 None => {
                     let stacks_node = format!("http://localhost:20443");
-                    let bitcoin_node = format!("http://devnet:devnet@localhost:18443");
+                    let bitcoin_node = format!("http://devnet:devnet@localhost18443");
```
obycode, line 183: This looks like a typo

obycode (changes requested):
Looks good! One possible typo?

components/clarinet-deployments/src/lib.rs
lgalabru: Good catch, thanks. addressed with latest.

obycode (approved):
🚀 :shipit:

PR #434, opened 2022-07-08
Introduce VSCode Bridge for Web LSP

hugocaillard:
- Required for https://github.com/hirosystems/clarinet-vscode/issues/4


components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -282,31 +282,20 @@ impl ClarityInterpreter {
...
+                //         Ok(res) => (res, vec![], true),
+                //         Err(error) => (
+                //             ContractAST::new(contract_identifier, vec![]),
+                //             vec![error.diagnostic],
+                //             false,
+                //         ),
+                //     };
+
+                // switch to v2 for now for performances
+                ast::build_ast(&contract_identifier, &snippet, &mut ())
```
hugocaillard, line 298: Using exclusively v2 for now since it speeds up the testing process

hugocaillard:
@lgalabru @obycode 
Still some work to do here but definitely ready for an first review, especially for potential obvious improvements on the rust logic / syntax

components/clarity-lsp/src/utils/vscode_vfs.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    fn read_contract_content(
+        &self,
+        manifest_location: FileLocation,
+        relative_path: String,
+    ) -> PerformFileAccess {
+        log!("reading contract");
+        let mut contract_location = manifest_location.get_parent_location().unwrap();
+        let _ = contract_location.append_path(&relative_path);
```
obycode, line 62: Should this error be handled?

components/clarity-lsp/src/utils/vscode_vfs.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    fn write_file(
+        &self,
+        manifest_location: FileLocation,
+        relative_path: String,
+        content: &[u8],
+    ) -> PerformFileAccess {
+        log!("writting contract");
+        let mut contract_location = manifest_location.get_parent_location().unwrap();
+        let _ = contract_location.append_path(&relative_path);
```
obycode, line 93: Should this error be handled?

components/clarity-repl/src/repl/interpreter.rs
obycode: I don't think we should merge this to develop though, right?

hugocaillard: Yep sure, I'll remove it when it's time to merge. Or maybe add a condition to have this behavior on wasm? At least during the "alpha" phase

obycode: On second thought, maybe it is fine to just delete this comparison with v1. We are confident enough in v2 that it will be merged upstream, so it should be fine to remove this check now.

obycode: Wait, no but there are still some things that v2 allows but will fail in v1... we need clarinet to catch those.

components/clarinet-deployments/src/lib.rs
```diff
@@ -453,9 +471,19 @@ pub async fn generate_default_deployment(
             }
         };
 
-        let mut contract_location = manifest.location.get_project_root_location()?;
-        contract_location.append_path(&contract_config.path)?;
-        let source = contract_location.read_content_as_utf8()?;
+        let (contract_location, source) = match file_accessor {
```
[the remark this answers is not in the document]
hugocaillard: It was this initial idea, but the wasm FileAccessor needs to return Future while the native one wouldn't. So since they have different signature, it's not possible. 
Do you confirm @lgalabru?

I agree that this logic is quite redundant

components/clarinet-deployments/src/lib.rs
```diff
@@ -619,3 +647,93 @@ pub async fn generate_default_deployment(
...
+    let dependencies = ASTDependencyDetector::detect_dependencies(&contract_asts, &BTreeMap::new());
+
+    let dependencies = match dependencies {
+        Ok(dependencies) => dependencies,
+        Err((dependencies, _)) => {
+            // No need to report an error here, it will be caught and reported
+            // with proper location information by the later analyses.
+            dependencies
+        }
+    };
```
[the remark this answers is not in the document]
hugocaillard: I have no idea 👀 cc @lgalabru

components/clarinet-deployments/src/requirements.rs
```diff
@@ -30,8 +45,24 @@ pub async fn retrieve_contract(
...
+            match file_accessor
+                .write_file(
+                    manifest_location,
+                    contract_location.to_string(),
+                    code.as_bytes(),
+                )
+                .await
+            {
+                Ok(_) => Ok(()),
+                Err(err) => Err(err),
```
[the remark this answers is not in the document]
hugocaillard: Not necessarily, what do you think would be better?

lgalabru:
Note to myself: let's ping @yourmercury whenever this PR merges

components/clarinet-deployments/src/lib.rs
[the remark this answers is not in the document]
obycode: The local version could return a future even if it doesn't need to. I guess the idea is just to avoid that unnecessary overhead if possible?

components/clarinet-deployments/src/requirements.rs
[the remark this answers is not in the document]
obycode: I think it would make sense to allow the caller to handle the error instead of just ignoring it. This function already returns a Result, so it should be simple.

components/clarinet-files/src/lib.rs
```diff
@@ -18,11 +18,30 @@ pub use project_manifest::{
...
 };
 use serde::ser::{Serialize, SerializeMap, Serializer};
+use std::future::Future;
+use std::pin::Pin;
 use std::{borrow::BorrowMut, path::PathBuf, str::FromStr};
 use url::Url;
 
 pub const DEFAULT_DEVNET_BALANCE: u64 = 100_000_000_000_000;
 
+pub type PerformFileAccess = Pin<Box<dyn Future<Output = Result<(FileLocation, String), String>>>>;
```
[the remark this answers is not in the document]
lgalabru: `PerformFileAccess` is not a trait, iirc it's a future that will be converted to a js function at some point?

components/clarinet-files/src/project_manifest.rs
```diff
@@ -99,7 +118,7 @@ impl ProjectManifest {
         }
 
         let project_name = project_manifest_file.project.name;
-        let mut project_root_location = manifest_location.get_project_root_location()?;
+        let mut project_root_location = manifest_location.get_parent_location().unwrap();
```
lgalabru, line 121: Should we switch back to the `try`, instead of `unwrap`?

components/clarity-lsp/src/common/state.rs
```diff
@@ -297,10 +294,19 @@ pub async fn build_state(
     // expect contracts to be created, edited, removed.
     // A on-disk deployment could quickly lead to an outdated
     // view of the repo.
-    let manifest = ProjectManifest::from_location(manifest_location)?;
+    let manifest = match file_accessor {
+        None => ProjectManifest::from_location(manifest_location)?,
+        Some(file_accessor) => {
+            // We should create a new async function in `ProjectManifest` that can take a `file_accessor: &Box<dyn FileAccessor>`
+            // as an argument, and that will use:
+            // file_accessor.read_manifest_content()
```
lgalabru, line 302: Is this what we ended up doing? Should we remove / update this comment? Explaining the rationale could be helpful for maintenance / knowledge sharing, as a lot of thoughts and experiments went into the final approach.

components/clarity-lsp/src/utils/mod.rs
```diff
@@ -243,6 +261,9 @@ pub fn build_default_native_keywords_list() -> Vec<CompletionItem> {
     items
 }
 
+use clarinet_files::FileLocation;
+use lsp_types::Url;
```
lgalabru, line 265: Should we move these imports back to the top?

components/clarity-lsp/src/vscode_bridge.rs
```diff
@@ -0,0 +1,212 @@
...
+                let command = if let Some(contract_location) = get_contract_location(uri) {
+                    LspNotification::ContractChanged(contract_location)
+                } else if let Some(manifest_location) = get_manifest_location(uri) {
+                    LspNotification::ManifestChanged(manifest_location)
+                } else {
+                    log!("Unsupported file opened");
+                    return Promise::resolve(&JsValue::null());
+                };
+
+                let editor_state = self.editor_state.clone();
```
lgalabru, line 132: I was confused for a moment. `self.editor_state` is a `Arc<RwLock<EditorState>>`. Should this be reflected in the name?

components/clarity-repl/src/repl/interpreter.rs
lgalabru: Should we take an opinionated stance in this specific tool, and just rely on `v2`? In my mind pushing for best practices is fine in the context of code edition, no?

lgalabru (commented):
Great work on this PR @hugocaillard!
I had few comments / questions, and can think of 2 action items:
- A pass on the changes for handling `unwrap`: using this method is fine during the development process, but create technical debt. For every new unwrap introduced, could you evaluate if the error needs to be cascaded (using `?`) or be panicking (ideally using `.expect(<message>)`)?
- Create (or point to) an issue with the remaining todos. It is great that we were able to phase this work stream, but the amount of remaining work is unclear to me :) 

Thank you for this effort, we're getting close :)

components/clarinet-deployments/src/requirements.rs
```diff
@@ -1,19 +1,34 @@
...
+                .read_contract_content(manifest_location.clone(), contract_location.to_string())
+                .await
+            {
+                Ok((_, source)) => Ok(source),
+                Err(err) => Err(err),
+            }
+        }
+    };
+
+    if contract_source.is_ok() {
```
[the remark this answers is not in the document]
lgalabru: Hmm nothing that I can think of. 
The caching is a speed optimization, I don't think that we should abort the deployment plan processing if caching can't be performed.

components/clarinet-deployments/src/lib.rs
```diff
@@ -619,3 +647,93 @@ pub async fn generate_default_deployment(
 
     Ok((deployment, artifacts))
 }
+
+pub fn generate_simnet_deployment_for_snippet(
```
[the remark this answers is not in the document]
lgalabru: @hugocaillard correct me if I'm wrong, but I think that the function`generate_simnet_deployment_for_snippet` is dead code, and we can get rid of it. I think I created this naive function for our initial experiments, it was a way to start getting some feedbacks from the parsing / analysis, without supporting the complexity of deployment plans.

components/clarinet-deployments/src/lib.rs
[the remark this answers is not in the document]
lgalabru: See https://github.com/hirosystems/clarinet/pull/434/files/fa6cb0023490253f5b33b992ae02863f64542d1c#r946596805

components/clarinet-deployments/src/lib.rs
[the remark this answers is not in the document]
lgalabru: I like futures, but I also like the idea of avoiding futures when we can :)
We have 2 versions of `generate_default_deployment`. the version in the `clarinet-deployment` crate is the `core` one and have this `optional` FileAccessor. The LSP crate uses it, but clarinet does not and for that reason wrap this `async fn generate_default_deployment(_, _, _, Option<_>)` into a  `fn generate_default_deployment(_, _, _)` - used everywhere in clarinet.

components/clarity-lsp/src/vscode_bridge.rs
hugocaillard: Yes good idea, would `editor_state_rw_lock` (or just `editor_state_lock`?) be a better name?

components/clarinet-files/src/lib.rs
```diff
@@ -18,11 +18,31 @@ pub use project_manifest::{
...
+pub type PerformFileWrite = Pin<Box<dyn Future<Output = Result<(), String>>>>;
+
+pub trait FileAccessor {
```
hugocaillard, line 31: @lgalabru 
Do we still need a trait for `FileAccessor`? 
The original idea was to have a Native and a Wasm FileAccessor, but since we only have a wasm one, I'm not sure we still need the trait. wdyt?

components/clarinet-files/src/lib.rs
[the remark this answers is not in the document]
hugocaillard: Exactly

components/clarinet-files/src/network_manifest.rs
```diff
@@ -233,6 +230,24 @@ impl NetworkManifest {
         NetworkManifest::from_location(&network_manifest_location, networks)
     }
 
+    pub async fn from_project_manifest_location_using_file_accessor(
+        location: &FileLocation,
+        networks: &(BitcoinNetwork, StacksNetwork),
+        file_accessor: &Box<dyn FileAccessor>,
```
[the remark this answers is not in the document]
hugocaillard: @obycode I tried to cascade this change but got stuck at this line https://github.com/hirosystems/clarinet/blob/2a27e628affa6f6996d72f1dc614dbf405e54b7a/components/clarity-lsp/src/common/backend.rs#L52 with the error: 

```
the size for values of type `(dyn FileAccessor + 'static)` cannot be known at compilation time
the trait `Sized` is not implemented for `(dyn FileAccessor + 'static)`rustc[E0277](https://doc.rust-lang.org/error-index.html#E0277)
option.rs(518, 17): required by a bound in `std::option::Option`
```

components/clarinet-deployments/src/lib.rs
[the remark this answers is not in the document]
hugocaillard: @obycode @lgalabru I'll mark this one as resolved unless you want to add something?

components/clarity-repl/src/repl/interpreter.rs
obycode: The problem though is that in the PR process to stacks-blockchain, the v2 parser got some changes that are not backwards compatible, but that was okay since it is gated for 2.1. If we start using it before 2.1, then it would allow some things that would be rejected by the node. For example, some spacing rules are different:

```clarity
{foo:1}
```

This is valid with parser v2, but not with v1. If I get the merge finished, then we can just use the merged version which will use only v2 for 2.1 epoch and only v1 for before.

components/clarinet-files/src/lib.rs
```diff
@@ -257,10 +241,63 @@ impl FileLocation {
         }
     }
 
-    pub fn get_project_manifest_location(&self) -> Result<FileLocation, String> {
-        let mut project_root_location = self.get_project_root_location()?;
-        project_root_location.append_path("Clarinet.toml")?;
-        Ok(project_root_location)
+    pub async fn get_project_manifest_location(
+        &self,
+        file_accessor: Option<&Box<dyn FileAccessor>>,
```
hugocaillard, line 246: @lgalabru @obycode 
Not sure if it's a good idea to pass `file_accessor` here. wdyt?

components/clarity-repl/src/repl/interpreter.rs
hugocaillard: I ended up restoring the initial logic, but we skip the v1 check in the wasm context

obycode (dismissed):
Looks great. Added some minor comments/questions.

components/clarinet-files/src/lib.rs
hugocaillard: Considering the simplicity it brings, I'd rather remove it now, and re-implement the trait in the future if/when needed

components/clarinet-files/src/lib.rs
hugocaillard: Ok I thing I need to revisit some stuff here.

The thing is that `get_project_root_location` calls url_exists https://github.com/hirosystems/clarinet/blob/61b0fdc4cec84fbb19451c6a9ece4e49ba2082a1/components/clarinet-files/src/lib.rs#L212, whichhas not been implemented, and won't be since it would need to be `async`, but making `get_project_root_location` async would lead to tons of changes.

So I probably want to delete `FileLocation::url_exists` and rework `get_project_root_location` but I'm not sure how yet

components/clarinet-files/src/lib.rs
hugocaillard: Attempt to remove the trait here: https://github.com/hirosystems/clarinet/commit/329ef9cc0fe878f196fa00dec2032f88dd1910b8

It's brings simplification and is easy to do but lead to unexpected issue: the WASM FS Accessor can't be used in place of the FileAccessor trait in non-wasm context

obycode (approved):
LGTM! Ship it!

PR #480, opened 2022-07-13
fix: microblocks + hyperchain adjustments 

lgalabru:
This PR:
- Adds the ability to use microblocks in deployment plans
- Displays microblocks in the clarinet terminal ui
- Fixes a bug when setting microblock's block identifiers + parent block identifiers 
- Fixes a bug in the hyperchain integration where the contract_id of the hyperchain contract was not using the remapped version. 
- Fixes a bug where clarinet integrate would take ages to terminate itself.


components/clarinet-files/src/network_manifest.rs
```diff
@@ -422,7 +433,7 @@ impl NetworkManifest {
                     .unwrap_or("devnet".to_string()),
                 bitcoin_controller_block_time: devnet_config
                     .bitcoin_controller_block_time
-                    .unwrap_or(30_000),
+                    .unwrap_or(90_000),
```
obycode, line 436: Oh bummer, did it need to be slowed down to function properly?

src/integrate/ui/mod.rs
```diff
@@ -97,30 +99,52 @@ pub fn start_ui(
...
+                        for i in indices_to_remove {
+                            app.mempool.items.remove(i);
+                        }
+                        for block_update in update.new_microblocks.into_iter() {
+                            app.display_microblock(block_update);
+                        }
                     }
-                } else {
-                    // TODO(lgalabru)
+                    _ => {} // handle display on re-org, theorically unreachable in context of devnet
```
obycode, line 147:
```suggestion
                    _ => {} // handle display on re-org, theoretically unreachable in context of devnet
```

vendor/orchestra-event-observer/src/indexer/stacks/blocks_pool.rs
```diff
@@ -670,7 +691,7 @@ impl StacksBlockPool {
                     Some(block) => block.clone(),
                     None => {
                         error!(
-                            "unable to retrive block {} from block store",
+                            "unable to retrive Stacks {} from block store",
```
obycode, line 694:
```suggestion
                            "unable to retrieve Stacks {} from block store",
```

obycode (dismissed):
Looks good!

components/clarinet-files/src/network_manifest.rs
lgalabru: No, 30s still works as expected, but now that microblocks are correctly handled and we have visual feedback for it in the terminal UI, using a longer block time might be more faithful to what will happen on Mainnet.

obycode (approved):

Issue #365, opened 2022-05-20
Compute and display diffs between deployment plans 


lgalabru:
When using commands like `clarinet integrate`, and a file `default.devnet-plan.yaml` is already present, we should probably be recomputing a plan, and comparing it with the one in place, to ensure that developers are not deploying a stale plan.

lgalabru:
Related: https://github.com/hirosystems/clarinet/issues/433

PR #488, opened 2022-07-15
feat: always compute new deployment plan

lgalabru:
This PR is taking on https://github.com/hirosystems/clarinet/issues/365  - over the last 2 weeks, we had 3 occurrences of developers experiencing issues caused by this limitation.

Screen Shot 2022-07-15 at 12 27 46

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -736,6 +736,7 @@ dependencies = [
  "serde_json",
  "serde_yaml",
  "shell-escape",
+ "similar",
  "stacks-rpc-client",
  "strum",
  "swc_common",
@@ -4445,6 +4446,12 @@ checksum = "e51e73328dc4ac0c7ccbda3a494dfa03df1de2f46018127f60c693f2648455b0"
 dependencies = [
  "libc",
 ]
+
+[[package]]
+name = "similar"
+version = "2.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e24979f63a11545f5f2c60141afe249d4f19f84581ea2138065e400941d83d3"
 
 [[package]]
 name = "siphasher"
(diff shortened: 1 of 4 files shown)


obycode:
So `--use-computed-deployment-plan` is basically a shortcut for calling `clarinet deployments generate` before whatever command they're trying to run? Could this solution be a bandaid hiding the real problem of users not understanding the deployment plan?

lgalabru:
> So --use-computed-deployment-plan is basically a shortcut for calling clarinet deployments generate before whatever command they're trying to run? 

There are a few subtleties at play.
With the current approach, we're tying to "mask" deployments plans from users when they play with simnet. So when you launch a console, or run your tests, a plan is generated under the hood, but not serialized to disk. If you call `clarinet deployment generate --simnet`, then a deployment is created on disk and saved, and `clarinet console` / `clarinet test` will tap into this plan.
When developers use `clarinet integrate`, a plan is generated **and** saved to disk. the assumption behind this is that these developers are a bit more familiar with clarity at this point, and introducing plans sounds reasonable at this point and the plan is created so that they can tweak it. The issue that this PR is solving, is that from there, if they go back to coding mode and add more contracts / requirements, or tweak their devnet configuration the previous plan is not recomputed, and developers are using `clarinet integrate` with an outdated plan.

 > Could this solution be a bandaid hiding the real problem of users not understanding the deployment plan?

In an ideal scenario, we have a "deployment plan flow" that just work, and developers don't have to look too much into it. If they have special needs, it's there and they can tweak them.

obycode:
Ok, thanks for that clarification. I didn't realize that `clarinet integrate` saves a deployment plan to disk, so I was thinking that if there was a plan saved, they did it on purpose and they should maintain it. If it could be auto-generated, then I'm okay with auto re-generating it.

Issue #496, opened 2022-07-21
Runtime errors when fuzzing a public function taking string-ascii


moodmosaic:
### Contract

```clarity
(define-data-var str (string-ascii 256) "")

(define-public (set-str (new-str (string-ascii 256)))
    (begin
        (ok (var-set str new-str))
    )
)
```

### Fuzz

```ts
import {
  Account,
  Chain,
  Clarinet,
  Tx,
  types,
} from "https://deno.land/x/clarinet@v0.31.1/index.ts";

// FIXME https://github.com/hirosystems/clarinet/issues/393
// FIXME import fc from "https://esm.sh/fast-check";
import fc from "https://cdn.skypack.dev/fast-check"; // Workaround for #393.

Clarinet.test({
  name: "[Exampe] ascii fuzz-test",
  fn(chain: Chain, accounts: Map<string, Account>) {
    fc.assert(
      fc.property(
        fc.ascii(), // https://ss64.com/ascii.html
        fc.constantFrom(...accounts.values()),
        (str: string, sender: Account) => {
          console.log(str);

          const block = chain.mineBlock([
            Tx.contractCall(
              "example",
              "set-str",
              [types.ascii(str)],
              sender.address,
            ),
          ]);
          console.log(block);
        },
      ),
      { numRuns: 10, unbiased: false }, // https://github.com/dubzzz/fast-check/discussions/2978
    );
  },
});
```

### Output

```
$ clarinet --version
clarinet 0.32.0

$ clarinet test
Running file:///.../empty-project/tests/example_test.ts
>
{ height: 2, receipts: [ { result: "(ok true)", events: [] } ] }
▲
Runtime error: example::set-str("▲") -> illegal ascii string "▲"
{ height: 3, receipts: [] }
▼
Runtime error: example::set-str("▼") -> use of unresolved contract 'ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG.example'
{ height: 4, receipts: [] }
⌂
Runtime error: example::set-str("⌂") -> use of unresolved contract 'STNHKEPYEPJ8ET55ZZ0M5A34J0R3N5FM2CMMMAZ6.example'
{ height: 5, receipts: [] }
&
Runtime error: example::set-str("&") -> use of unresolved contract 'STNHKEPYEPJ8ET55ZZ0M5A34J0R3N5FM2CMMMAZ6.example'
{ height: 6, receipts: [] }
`
[...]

PR #500, opened 2022-07-22
Fix 2 errors in the testing framework

obycode:
See #496.

### DIFF
--- a/deno/index.ts
+++ b/deno/index.ts
@@ -360,11 +360,11 @@ export namespace types {
   }
 
   export function ascii(val: string) {
-    return `"${val}"`;
+    return JSON.stringify(val);
   }
 
   export function utf8(val: string) {
-    return `u"${val}"`;
+    return `u${JSON.stringify(val)}`;
   }
 
   export function buff(val: ArrayBuffer | string) {
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -752,7 +752,14 @@ impl Session {
         );
 
         self.set_tx_sender(sender.into());
-        let result = self.interpret(snippet, None, None, true, Some(test_name.clone()), None)?;
+        let result = match self.interpret(snippet, None, None, true, Some(test_name.clone()), None)
+        {
+            Ok(result) => result,
+            Err(e) => {
+                self.set_tx_sender(initial_tx_sender);
+                return Err(e);
+            }
+        };
         if let Some(ref cost) = result.cost {
             self.costs_reports.push(CostsReport {
                 test_name,


deno/index.ts
```diff
@@ -360,11 +360,11 @@ export namespace types {
   }
 
   export function ascii(val: string) {
-    return `"${val}"`;
+    return JSON.stringify(val);
   }
 
   export function utf8(val: string) {
-    return `u"${val}"`;
+    return `u${JSON.stringify(val)}`;
```
moodmosaic, line 367: Good catch here 👍

moodmosaic:
Shall we re-run that fuzz test before merging?

moodmosaic (commented):
Shall we re-run the fuzz test before merging?

obycode:
> Shall we re-run the fuzz test before merging?

Yes, I've been running it and it looks good. You will need to restrict the string parameter a bit more to avoid `illegal ascii string` and `unknown escape character` errors, but I think this fixes the problems in the test infrastructure.

obycode:
Example output with this fix:

```
Running file:///Users/brice/work/496-fuzz/tests/example_test.ts
Runtime error: example::set-str("") -> illegal ascii string ""
{ height: 2, receipts: [] }
{ height: 3, receipts: [ { result: '(ok " ")', events: [] } ] }
{ height: 4, receipts: [ { result: '(ok "!")', events: [] } ] }
Runtime error: example::set-str("\u0019") -> unknown escape character, 'u'
{ height: 5, receipts: [] }
{ height: 6, receipts: [ { result: '(ok " ")', events: [] } ] }
{ height: 7, receipts: [ { result: '(ok "%")', events: [] } ] }
{ height: 8, receipts: [ { result: '(ok """)', events: [] } ] }
{ height: 9, receipts: [ { result: '(ok """)', events: [] } ] }
{ height: 10, receipts: [ { result: '(ok "2")', events: [] } ] }
{ height: 11, receipts: [ { result: '(ok """)', events: [] } ] }
* [Exampe] ascii fuzz-test ... ok (32ms)
```

moodmosaic:
Sounds good. Yes, making the infrastructure more robust is the ultimate goal. Thank you 👍

On that fuzz test, we can tweak the ascii generator if needed and see what happens. 🎲

moodmosaic:
Might be out of the scope of this PR, but while we're here try to re-run your test this time by changing numRuns: 10 to numRuns: 5000. See if anything else breaks.

obycode:
I just tried this (without any modifications to the ascii generator). The test finishes successfully, with a bunch of runtime errors seen from illegal characters.

moodmosaic:
👍 Thanks @obycode

lgalabru (approved):
Thanks @obycode!

obycode:
Not sure why those two "Generate test coverage" jobs didn't run. I'm going to try re-running. @CharlieC3, do you have any thoughts? Has this new check run at all before?

obycode:
Those jobs don't show here - https://github.com/hirosystems/clarinet/actions/runs/2721283881/usage

obycode:
Oh, I see, those status bits were set, but this branch doesn't have those jobs. Rebasing on top of main now...

CharlieC3:
@obycode Looks like those jobs are commented out in the workflow in both branches?
https://github.com/hirosystems/clarinet/blob/fix/496-sender/.github/workflows/ci.yaml#L91-L165
https://github.com/hirosystems/clarinet/blob/develop/.github/workflows/ci.yaml#L91-L165

obycode:
Oh, you're right... what was I just looking at?

Ahh, in the in-progress `feat/ls-vscode-commands` they're un-commented. I'll just disable those checks from the required list for now.

PR #504, opened 2022-08-01
docs: update requirements docs in readme

hugocaillard:
The `requirements` in readme show a deprecated syntax

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,10 +435,9 @@ Before referring to contracts deployed on Mainnet, they should be explicitily be
 ```toml
 [project]
 name = "my-project"
-requirements = [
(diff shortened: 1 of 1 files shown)


lgalabru:
Thanks @hugocaillard! unfortunately, we can't use requirements in our examples because they are also used by the CI for running some end to end test.
By adding some requirements, we'd be relying on internet connectivity for these tests. We tried that in the past, and it turned out to it introduced flakiness in our test suite.
If we want to test requirements, I think we'd need to leverage our refactoring on the `FileAccessor` present in your LSP request, for mocking network calls.

hugocaillard:
@lgalabru Alright makes sense. I wasn't sure if it was only "examples", or also used as test data.
I'll that change and only leave the readme update then

obycode:
See also - community PR #502

hugocaillard:
Ah! I'll close this one

Issue #423, opened 2022-06-22
Revisit deno vendoring


lgalabru:
It looks like the deno team started unbundling their stack. We should explore what is possible with these 2 repos:
- https://github.com/denoland/deno_emit
- https://github.com/denoland/deno_graph

Thanks to this effort, we might be able to get rid of our vendored fork of deno - which would be way healthier.
Would resolve https://github.com/hirosystems/clarinet/issues/393

lgalabru:
Undocumented, unfortunately, but happy to hop on a call if you'd like to have an overview.

moodmosaic:
Yep, [let's document some of this](https://calendly.com/moodmosaic/30min) so that it's available for anyone else that might want to dive in.

moodmosaic:
🏓

hstove:
I understand that un-vendoring would be a pain, but what about upgrading the base version of `Deno`? For example, I'd like to write [BDD-style tests](https://deno.land/manual/testing/behavior_driven_development#behavior-driven-development), but the version of Deno that Clarinet runs doesn't have support for `Deno.TestContext`.

moodmosaic:
If you're interested in the BDD-style hooks mentioned on that Deno link, perhaps you'll find the work from @LNow and myself in [this](https://github.com/LNow/testing-example) repo interesting.

lgalabru:
To be addressed in https://github.com/hirosystems/clarinet/pull/511.

Issue #398, opened 2022-06-03
`clarinet fuzz` command and heterogeneous test-suites


moodmosaic:
*This was started as two separate GitHub issues, but then I thought I should merge them.* — The idea is:

`clarinet test` can support property tests, and those tests may be written in either TypeScript or Clarity
`clarinet fuzz` can then turn those tests into fuzz tests

### Context

[Clarinet tests are essentially Deno tests](https://blog.nikosbaxevanis.com/2022/03/05/clarity-property-based-testing-primer/#summary). This has the interesting side-effect of being able to use what's already available in JS/TS ecosystem when testing Clarity code. (For example, [fast-check](https://github.com/dubzzz/fast-check) and [dspec](https://github.com/LNow/dspec) can be used, as we've done [here](https://github.com/LNow/testing-example/pull/1/files) with @LNow.)

A typical workflow is declaring functions in Clarity and then writing tests in TypeScript. Clarinet can then *discover* those tests (with the help of Deno) and run them. It would be practical however to also discover [tests written in Clarity](https://github.com/jcnelson/poxl/tree/main/tests) and run those as well.

### Discoverability

In addition to existing test suite(s) in TypeScript, Clarinet can execute as a test any Clarity function that meets the following criteria:

* The method is public
* The method name starts with `test` (configurable in `Clarinet.toml`)

(If a public function  named `beforeEach` is present it can be executed before each test is run. Such a function can exist in the style of testing @jcnelson does [here](https://github.com/jcnelson/poxl/blob/main/tests/test-poxl.clar), may be discussed also on a  separate thread, however.)

<table>
  <thead>
    <tr>
      <th>type</th>
      <th>prefix</th>
      <th>arguments</th>
      <th>semantics</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>concrete</td>
      <td><code class="language-plaintext highlighter-rouge">test</code></td>
      <td>no</td>
      <td>single execution with concrete values</td>
    </tr>
    <tr>
      <td>property</td>
      <td><code class="language-plaintext highlighter-rouge">test</code></td>
      <td>yes</td>
      <td>multiple executions with randomly generated concrete values (smaller values)</td>
    </tr>
    <tr>
      <td>fuzz</td>
      <td><code class="language-plaintext highlighter-rouge">test</code></td>
      <td>yes</td>
[...]

Issue #393, opened 2022-06-01
[BUG] `clarinet test` is not able to run 100% correct typescript code


LNow:
Together with @moodmosaic we're exploring the possibilities of adding property-based testing solutions to clarity world.
Recently code that is 100% correct stopped working. At first we though that `fast-check`, a property-based testing library we use is at fault but I managed to eliminate this possibility. Same with recently updated `deno_std` library.

It looks like Deno runtime embedded in Clarinet is the root cause of our problems.

Sample repository: https://github.com/LNow/fast-check-test

When we run `deno test` everything works perfectly fine.
When we run `clarinet test` it fails with following error:
```
TS7053 [ERROR]: Element implicitly has an 'any' type because expression of type 'unique symbol' can't be used to index type 'Record<symbol, unknown>'.
  Property '["https://deno.land/std@0.141.0/node/internal/crypto/_keys.ts".kKeyType]' does not exist on type 'Record<symbol, unknown>'.
    obj != null && (obj as Record<symbol, unknown>)[kKeyType] !== undefined
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    at https://deno.land/std@0.141.0/node/internal/crypto/_keys.ts:7:20

TS7053 [ERROR]: Element implicitly has an 'any' type because expression of type 'unique symbol' can't be used to index type 'Record<symbol, unknown>'.
  Property '["https://deno.land/std@0.141.0/node/internal/crypto/constants.ts".kKeyObject]' does not exist on type 'Record<symbol, unknown>'.
    obj != null && (obj as Record<symbol, unknown>)[kKeyObject] !== undefined
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    at https://deno.land/std@0.141.0/node/internal/crypto/_keys.ts:13:20

Found 2 errors.
```

lgalabru:
We should be able to update the typescript transpiler being embedded in Clarinet.

moodmosaic:
@lgalabru, @LNow, in case this helps to troubleshoot further, the following patch makes `clarinet test` to run successfully:

```diff
$ git diff
diff --git a/fast_check.ts b/fast_check.ts
index 9b1728f..968e4bd 100644
--- a/fast_check.ts
+++ b/fast_check.ts
@@ -1,2 +1,2 @@
[...]

Issue #361, opened 2022-05-20
Ability to fork contract easily


lgalabru:
At some point, clarinet was able to facilitate the fork of contracts, thanks to the command:
```
clarinet contracts fork <contract-id>
```
This command broke a few times in presence of requirements, and was removed with #333.
If there's some demand for this feature, we could re-implement it pretty quickly with the new deployments design, but I think that this feature might be superseded by https://github.com/hirosystems/clarinet/issues/341.

moodmosaic:
Yes, perhaps a template can also be derived from an existing contract. The best I can currently think of is via adding support for multiple sources, as already mentioned in #341, for example:

```
$ clarinet new --add-source <contract-id> [other-options-here]
```

LNow:
I used this feature quite often when I was trying understand some contracts which were already deployed.
Anything that allows me to download a contract and save it as a file is a "must have" for me.

moodmosaic:
@lgalabru, how can one fork/fetch [this contract from testnet](https://explorer.stacks.co/txid/0x5864dabc9122732e16fcebd5ddaa727db8614eaee59499967c18011c1ddbd5b8?chain=testnet) locally via deployment plans?

lgalabru:
@moodmosaic this can be achieved with the command:
```bash
clarinet requirements add ST3QFME3CANQFQNR86TYVKQYCFT7QX4PRXM1V9W6H.cargo
```

moodmosaic:
@lgalabru, thank you. — Sidenote: running the command above touched Clarinet.toml in more places than I would expect:

```diff
@@ -1,21 +1,25 @@
-
 [project]
 name = "cargo"
 authors = []
+description = ""
 telemetry = false
-cache_dir = "./.requirements"
+cache_dir = "C:\\Snapshot\\dev\\pub\\stacks\\tmp\\cargo\\.\\.requirements"
+boot_contracts = ["pox", "costs-v2", "bns"]
+
+[[project.requirements]]
+contract_id = "ST3QFME3CANQFQNR86TYVKQYCFT7QX4PRXM1V9W6H.cargo"
+
+[contracts]
 
-# [contracts.counter]
-# path = "contracts/counter.clar"
+[repl]
+costs_version = 2
+parser_version = 2
 
 [repl.analysis]
 passes = ["check_checker"]
-check_checker = { trusted_sender = false, trusted_caller = false, callee_filter = false }
 
-# Check-checker settings:
[...]

PR #511, opened 2022-08-07
Revisit Deno vendoring

lgalabru:
With this PR, we are revisiting the integration points between Deno and Clarinet.
As explained in https://github.com/hirosystems/clarinet/issues/423, developers have been busy unbundling Deno into smaller crates, which is exciting for Clarinet. 
In this new approach, Clarinet relies as much as possible on the canonical version of Deno (using the libs published on crates.io).
The remaining code being imported in clarinet is:
- Glue code for configuring a Deno test environment. The approach here is to import a large portion of `deno-cli`, call the right functions and let the linker eliminating the dead code.
- Deno runtime: we could rely on `deno-runtime` but this crate does not give the ability to select features. as such, things like WebGL are being imported and compiled, which adds a lot of complexity to the compilation process.

Todo
- [ ] More tests
- [ ] More pruning
- [ ] Explicit message for developers using v0.33.0 -> need to upgrade to v0.34.0
- [ ] Remove gitsubmodule


moodmosaic:
From an outside-in perspective, things look good. #393 and related/external issues seem to be solved. I should be able to run this with more test-suites soon. Thank you @lgalabru 👍

moodmosaic:
Looks like [this](https://github.com/hirosystems/clarinet/issues/361#issuecomment-1154122221) functionality no longer works in this branch:

```bash
$ clarinet requirements add ST3QFME3CANQFQNR86TYVKQYCFT7QX4PRXM1V9W6H.cargo
Updated Clarinet.toml with requirement ST3QFME3CANQFQNR86TYVKQYCFT7QX4PRXM1V9W6H.cargo

# Though nothing was actually added.
```

moodmosaic:
@lgalabru:

Deno supports using a TS config file, though like the rest of Deno, the detection and use of a configuration file is not automatic. To use a TS config file with Deno, we have to provide a path on the command line, for example:

```
deno test --config ./deno.json tests.ts
```

>**IMPORTANT** Deno prefers naming that file `deno.json` isntead of `tsconfig.json` as explained [here](https://deno.land/manual/typescript/configuration).

It'd be useful if we could load such configuration in Deno when running Clarinet `test`, for example:

```bash
# Explicit
clarinet test --deno-config ./deno.json

# Implicit (if there's a deno.json file present in the root dir, note the user that it's been loaded.)
clarinet test
```

This may be useful for #398 where we might have to enable [decorators ](https://www.typescriptlang.org/docs/handbook/decorators.html) so that we can reflect on the test function arguments and pick the right fast-check [arbitrary](https://github.com/dubzzz/fast-check/blob/main/packages/fast-check/documentation/Arbitraries.md), and also allow the user through [metadata](https://rbuckton.github.io/reflect-metadata/) to override our built-in arbitraries.

PR #521, opened 2022-08-24
Feat: lsp insights

hugocaillard:
- [ ] Base PR on develop once #434 is merged

Fix: https://github.com/hirosystems/clarinet-vscode/issues/6


components/clarity-repl/src/clarity/analysis/types.rs
```diff
@@ -151,10 +151,23 @@ impl ContractAnalysis {
...
+        }
+        if let Some(function_type) = self.read_only_function_types.get(name) {
+            return Some(("read_only".into(), function_type));
+        }
+        if let Some(function_type) = self.private_function_types.get(name) {
+            return Some(("private".into(), function_type));
+        }
+        None
+    }
+
```
lgalabru, line 170: Could we move this code in the `lsp` module instead? 
We're about to break the ability to compile `clarity-repl` to wasm, so we might have to code freeze that module.

hugocaillard: Sure thing, I'll try to make a pass on this PR today. It's been a bit abandoned

hugocaillard:
Closing for now. Keeping the branch for reference

PR #533, opened 2022-09-13
fix: cleaner path in lsp notifications

hugocaillard:
Avoid showing the full vfs path of a contract in the lsp notifications.

#### Before
Web
Screenshot 2022-09-13 at 15 06 51
Desktop
Screenshot 2022-09-13 at 15 07 37

#### After
Web & Dekstop

Screenshot 2022-09-13 at 15 43 51

`get_relative_location` couldn't be used because it calls `get_project_root_location` that only work on native fs path (not on urls) https://github.com/hirosystems/clarinet/blob/e4c3a4122e13aa7272f46323e54f417cf292f9ec/components/clarinet-files/src/lib.rs#L208

So I took advantage of the `contracts_lookup` prop to retrieve the manifest location.
In the future, we probably want to call more `get_relative_location_from_manifest` and `get_root_location_from_manifest_location` (instead of the non-`from-manifest` ones that call the native FS)

### DIFF
--- a/components/clarinet-files/src/lib.rs
+++ b/components/clarinet-files/src/lib.rs
@@ -301,13 +301,12 @@ impl FileLocation {
         Ok(network_manifest_location)
     }
 
-    pub fn get_relative_location_from_manifest(
+    pub fn get_relative_path_from_base(
         &self,
-        manifest_location: &FileLocation,
+        base_location: &FileLocation,
     ) -> Result<String, String> {
-        let base = self.get_root_location_from_manifest_location(manifest_location)?;
         let file = self.to_string();
-        Ok(file[(base.to_string().len() + 1)..].to_string())
+        Ok(file[(base_location.to_string().len() + 1)..].to_string())
     }
 
     pub fn get_relative_location(&self) -> Result<String, String> {
--- a/components/clarinet-files/src/lib.rs
+++ b/components/clarinet-files/src/lib.rs
@@ -307,7 +307,7 @@ impl FileLocation {
     ) -> Result<String, String> {
         let base = self.get_root_location_from_manifest_location(manifest_location)?;
         let file = self.to_string();
-        Ok(file[(base.to_string().len())..].to_string())
+        Ok(file[(base.to_string().len() + 1)..].to_string())
     }
 
     pub fn get_relative_location(&self) -> Result<String, String> {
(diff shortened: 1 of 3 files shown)


components/clarinet-files/src/lib.rs
```diff
@@ -242,6 +213,59 @@ impl FileLocation {
         }
     }
 
+    pub fn get_project_root_location(&self) -> Result<FileLocation, String> {
```
hugocaillard, line 216: This method has been moved closer to "get_root_location_from_manifest_location" but unchanged

hugocaillard:
@lgalabru @obycode 
I could also store the manifest_location and the base_location in the contract_loop hashmap, since we often need it, wdyt?
```rs
pub contracts_lookup: HashMap<FileLocation, (FileLocation, FileLocation)>, // might be a bit hard to use
```

hugocaillard:
@lgalabru @obycode PR updated with a custom struct in the `contract_lookup` hash map, that could eventually have more properties in the future (and be more maintainable than a tuple). Let me know what you think.
If this is validated, I could go further and fully remove `get_project_manifest_location` and some calls to `get_project_root_location` that perform FS/VFS calls

components/clarity-lsp/src/common/state.rs
```diff
@@ -87,9 +94,23 @@ impl EditorState {
     }
 
     pub fn index_protocol(&mut self, manifest_location: FileLocation, protocol: ProtocolState) {
+        let base_location = manifest_location
+            .get_root_location_from_manifest_location(&manifest_location)
```
hugocaillard, line 98: Seems like an anti-pattern (self and manifest_location are the same) 🤔 
I initially made `get_root_location_from_manifest_location` to call it on `contract_location`s (which is still possible but not used anymore). wdyt?

obycode: It looks like `get_root_location_from_manifest_location` doesn't need to be a method of `FileLocation` at all -- it doesn't use `self`. So what about just making it a function?

obycode (approved):
LGTM 👍. I added a reply to your question.

hugocaillard:
@obycode Thanks for the review. Since `get_root_location_from_manifest_location` was only called once I simplified it and put it directly in `index_protocol`

obycode (approved):
Thanks @hugocaillard!

PR #542, opened 2022-09-21
fix: set pg docker image tag to postgres:14

rafaelcr:
This will make the postgres version consistent with the API


lgalabru:
LGTM, thanks @rafaelcr!

lgalabru (approved):

PR #553, opened 2022-10-03
fix: few fixes before release the extension

hugocaillard:
- Removed the clarity insights options from the config so it's not visible to the users (but the code is still here)
- Fix an issue that make the extension crash when opening a toml file that is not a Clarinet.toml
- Extension won't activate on any toml any way
- Remove useless logs
- Better error message when file is missing


lgalabru (approved):
LGTM, thanks @hugocaillard!
I'll go ahead and merge, as I need the fix from that branch for ongoing QA.

PR #562, opened 2022-10-05
LSP: remove diagnostic notifications

hugocaillard:
For simplicity, remove notification message.
- no other language extension (that I know of) do that
- there are three other visual clues that there’s an error: 1 in the tree view, 2 in the code, 3 in the problem panel
- keep vscode notification for actual errors (missing file, requirement not found, etc)


lgalabru (approved):
LGTM, thanks @hugocaillard!

Issue #558, opened 2022-10-04
LSP: Error when trying to write requirements contracts on readonly URL


hugoclrd:
On vscode.dev the extension tries to write `requirements` contracts on readonly github url fails to do so on vscode.dev and github.dev with the following unhandled error:

```
Request vfs/writeFile failed with message:
Unable to modify readonly file 'github:/......cache/requirements/SP<....>.clar' s
@https:<...>0.2.8/extension/server/dist/serverBrowser.js
```

PR #563, opened 2022-10-05
LSP: ignore didOpen notifications on github urls

hugocaillard:
Fixes: https://github.com/hirosystems/clarinet/issues/558

On vscode.dev and github.dev, the `didOpen` (and `didClose`) notifications are sent twice.
- 1st time with `vscode-vfs://<....>/file.(clar|toml)`
- 2nd time with `github://<.......>/file.(clar|toml)`

I don't know if it's intended (link to vscode issue coming soon) but it leads to a big issue on our side, the `manifest.project.cache_location` is changed 

Maybe we could also make this location more robust (less subject to changes when a contract is opened), in the meantime, this PR fixes the root cause by only allowing didOpen notifications on URLs starting with `["file", "vscode-vfs", "vscode-test-web"]`.
It also limits the number of notifications between JS <-> Rust 👍


lgalabru (approved):
🚀

PR #569, opened 2022-10-07
fix: completion item kind mapping

hugocaillard:
- [x] Put the right completion item kind, the fix could probably be done on the rust side as well but we already have the mapping on completion items occurring here. Might need some future investigation to be able to remove the mapping and send clean data from rust

- [x] Don't send completion request when it might fail while the `editor_state_lock` is locked (server handling a notification), it shouldn't be noticeable but might avoid some "auto-completion request failed" popup while typing


lgalabru (approved):
LGTM, thanks @hugocaillard!

PR #570, opened 2022-10-07
Cache boot contracts

lgalabru:


hugocaillard (approved):
From >2500ms to <100ms 🚀

PR #578, opened 2022-10-10
refacto: performance optimisation with parser v2

hugocaillard:
- [x] Remove logs before making it ready

### DIFF
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -343,6 +343,7 @@ pub async fn build_state(
         }
     };
 
+    web_sys::console::time_with_label("generate default deploy");
     let (deployment, mut artifacts) = generate_default_deployment(
         &manifest,
         &StacksNetwork::Simnet,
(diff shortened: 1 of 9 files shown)


components/clarity-vscode/client/src/clientBrowser.ts
```diff
@@ -19,7 +19,7 @@ export async function activate(context: ExtensionContext) {
     serverWorkerReady = resolve;
     workerTimeout = setTimeout(() => {
       reject(new Error("worker timeout"));
-    }, 5000);
+    }, 10000);
```
hugocaillard, line 22: On (very) slow networks, the server can take a few seconds to load, to be safe let's increase the delay before throwing an error

lgalabru (approved):
Testing CI - Do not merge

components/clarity-lsp/src/common/state.rs
```diff
@@ -342,9 +344,14 @@ pub async fn build_state(
...
 
-    let (deployment, mut artifacts) =
-        generate_default_deployment(&manifest, &StacksNetwork::Simnet, false, file_accessor)
-            .await?;
+    let (deployment, mut artifacts) = generate_default_deployment(
+        &manifest,
+        &StacksNetwork::Simnet,
+        false,
+        file_accessor,
+        Some((StacksEpochId::Epoch21, ClarityVersion::Clarity2)),
```
obycode, line 352: I think this should be `Clarity1` until 2.1 is live, so that contracts are checked in that context. It will still use the new parser.

obycode:
Also, you can amend your commit message typo so that the tool picks it up correctly.

components/clarity-lsp/src/common/state.rs
hugocaillard: So I ended up adding an argument only for the Epoch

obycode: Sounds good. It should probably be `Epoch2_05`, but I'm not sure if that will matter or not.

This brings a question to mind: how will we handle this later, after 2.1 is live. How do we decide which epoch and clarity version to use for the dependencies?

obycode: By `It should probably be `Epoch2_05`, I meant on line 381.

obycode (approved):
Looks great!

lgalabru (approved):
👍 thanks @hugocaillard!

PR #582, opened 2022-10-11
fix: propagate the eval hooks passed to `eval`

obycode:
This was accidentally removed during some merging, and it prevents the debuggers from working.


lgalabru (approved):
Ah, thanks @obycode!

PR #590, opened 2022-10-14
feat: add a simple easter egg to the REPL

obycode:
When a developer types `/-/` in the REPL (`clarinet console`), it will pull a message from a GitHub gist and display it. This is just a super quick PoC to decide if we want to include it. The downside is that it requires pulling `reqwest` into `clarity-repl`.

An alternative is to only change the message once per release, then it doesn't need to pull anything from the Internet, but it could limit the fun that the marketing team can have with it.


lgalabru (commented):
Thanks @obycode, that looks fun! 
Could we host the gist under `hirosystems` and make sure that `reqwest` is only being pull when the `cli` feature flag is being pulled? Otherwise this would be breaking `wasm` builds.

obycode:
It seems that organizations can't have gists unfortunately (see https://github.com/community/community/discussions/7923). I chose a gist as a super easy place to play with, but it could be anywhere else too. We just need someplace that will be easy for anyone from the marketing team to put some text. Probably just a regular repo under the org would be fine.

Sounds good about the `cli` feature flag. I can also look into pulling it some other way, that would be wasm compatible. Doesn't the wasm code already use `reqwest` to download the requirements though?

obycode:
Copy from Elena for release notes when this makes it in:

> :shushing_face: Easter Egg
> Pay attention devs, because you might find
> a little surprise we recently designed
> With the Hiro brand, you can command
> a group of gourds in Clarinet land

components/clarity-repl/src/repl/session.rs
```diff
@@ -1311,3 +1318,27 @@ mod tests {
...
+            (rt.handle().clone(), Some(rt))
+        }
+    };
+    let response = handle.block_on(async { future.await });
+    response
+}
+
+async fn fetch_message() -> Result<String, reqwest::Error> {
+    const gist: &str =
+        "https://gist.githubusercontent.com/obycode/d10b5a40cd926275ff6bde07d2270f2a/raw/";
```
lgalabru, line 1340: @aravindgee could you help us and give us access to a storage bucket / cdn for these assets?

aravindgee: Uploaded to one of our public buckets, here's the URL: https://storage.googleapis.com/hiro-public/assets/clarinet-egg.txt

[[FYI a better CDN for Hiro Assets is in the works](https://github.com/hirosystems/devops/issues/1048)]

How frequently are we looking to update this file?

lgalabru: Branch updated with new URL - will merge!

Issue #591, opened 2022-10-14
code coverage reports broken


whoabuddy:
I've run into this issue across two repositories now, see:

- citycoins/protocol#7
- citycoins/protocol#2 and [related action](https://github.com/citycoins/protocol/actions/runs/3237277669/jobs/5304139013)
- citycoins/contracts#210 and [related action](https://github.com/citycoins/contracts/actions/runs/3243167638/jobs/5317482634)

Snippet from an action log showing the file was found and uploaded:

```
[2022-10-12T19:17:24.027Z] ['info'] Processing ./coverage.lcov...
[2022-10-12T19:17:24.030Z] ['info'] Detected GitHub Actions as the CI provider.
[2022-10-12T19:17:24.033Z] ['info'] Pinging Codecov: https://codecov.io/upload/v4?package=github-action-2.1.0-uploader-0.3.2&token=*******&branch=feat%2Finitial-config&build=3237277669&build_url=https%3A%2F%2Fgithub.com%2Fcitycoins%2Fprotocol%2Factions%2Fruns%2F3237277669&commit=6878329a94ee47fc2e9cb558d90537acb30bc0aa&job=Smart+Contract+Tests&pr=2&service=github-actions&slug=citycoins%2Fprotocol&name=&tag=&flags=&parent=
[2022-10-12T19:17:24.805Z] ['info'] https://app.codecov.io/github/citycoins/protocol/commit/6878329a94ee47fc2e9cb558d90537acb30bc0aa
https://storage.googleapis.com/codecov/v4/raw/2022-10-12/395E17979B9AAFF5990093CE2CC37CA7/6878329a94ee47fc2e9cb558d90537acb30bc0aa/85236748-2a26-4ebf-8764-f3e341825dc6.txt?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Credential=GOOG1EJOGFN2JQ4OCTGA2MU5AEIT7OT5Z7HTFOAN2SPG4NWSN2UJYOY5U6LZQ%2F20221012%2FUS%2Fs3%2Faws4_request&X-Amz-Date=20221012T191724Z&X-Amz-Expires=10&X-Amz-SignedHeaders=host&X-Amz-Signature=2a3d97d6bff250acb01969c32ceb3fc912b7c9ea2d75bfe8fac6d23bd34fe3b6
[2022-10-12T19:17:24.805Z] ['info'] Uploading...
[...]

PR #592, opened 2022-10-14
fix: fix coverage tracking

obycode:
It seems these lines were commented out during the merge and never put back. This simple fix should be all we need.

Fixes: #591


lgalabru (approved):
Thanks @obycode!

Issue #54, opened 2021-06-24
[Feature] Add deps.ts file for less error prone library upgrades.


LNow:
Right now all test files and any other files that uses some part of Clarinet imports Clarinet library at the top of each file.

For example:
```typescript
import {
  Account,
  Chain,
  Clarinet,
  Tx,
  types,
} from "https://deno.land/x/clarinet@v0.13.0/index.ts";
```

It's ok when we have only few files and we know that we won't have to or don't want to upgrade Clarinet library very often. But when we'll have more files suddenly upgrading to new version is a bit challenging, as we have to find every single file with import and update it.

Deno [documentation ](https://deno.land/manual/examples/manage_dependencies
) suggests that it can be beneficial to use `deps.ts` file.

I think that Clarinet should suggest this approach when we create a new project. That way it would teach users a good practice that will save them some time in the future.

We tried it out in CityCoin and I like this solution:
https://github.com/citycoins/citycoin/blob/main/deps.ts
https://github.com/citycoins/citycoin/blob/bf7e22ff8bf29f74a0a5a60c5b0737c2d1cd18f3/src/client.ts#L1-L5

LNow:
@philipdesmedt, @nieldeckx you may want to try this approach in Arkadiko. I saw that you have quite a lot of test files.

philipdesmedt:
@LNow This is a good refactoring. I'll talk to Niel and try this over the next week or so

lgalabru:
Interesting, thank you @LNow for this approach!

MarvinJanssen:
Great idea, will adopt this as well.

PR #618, opened 2022-10-19
Refactor: improve deno lib types and add some tests

### LINKED ISSUE #573: Deno: refactor test lib
The clarinet-deno lib could be improved it two ways
- fix many Deno lint warnings (const vs let, no unused var, etc)
- improve typing (String instead of string, avoid `Object`, no any)

hugocaillard:
- Fix deno lint warnings
    - const over let
    - no any
- Remove unnecessary type annotations
- Remove the use of `String` or `Object` as types
- Add types to blockchain events
- Add some unit tests

Other improvements could be done (more tests, improve error messages, etc) but this is a good start

Fixes #573


components/clarinet-deno/ext/stacksjs-helper-generator.ts
```diff
@@ -127,62 +124,42 @@ Clarinet.run({
...
-      ]);
+      code.push([`}`, ``]);
     }
-
-    const write = await Deno.writeTextFile(
-      "./artifacts/contracts.ts",
-      code.flat().join("\n"),
-    );
   },
 });
```
lgalabru, line 164: Nice, thanks. could you please run this extension on a contract, and make sure that the output is correctly formatted?

components/clarinet-deno/index.test.ts
```diff
@@ -0,0 +1,113 @@
...
+      },
+    },
+  ];
+
+  assertObjectMatch(events.expectPrintEvent(id, '"ok"'), {
+    contract_identifier: id,
+    topic: "print",
+    value: '"ok"',
+  });
+});
```
lgalabru, line 113: Thanks you for adding theses tests!

components/clarity-vscode/package.json
```diff
@@ -24,7 +24,7 @@
     "test": "node client/dist/tests/runTests.js",
     "lint": "eslint ./client/src ./server/src --ext .ts,.tsx",
     "dev:watch": "webpack -c ./webpack.config.dev.js -w",
-    "dev:browser": "vscode-test-web --extensionDevelopmentPath=. ./test-data --open-devtools --browser=chromium",
+    "dev:browser": "vscode-test-web --extensionDevelopmentPath=. ../../../../citycoins/protocol --open-devtools --browser=chromium",
```
lgalabru, line 27: This looks like a machine specific directory layout?

components/clarinet-deno/index.ts
```diff
@@ -388,115 +405,115 @@ export namespace types {
...
+      amount: number | bigint,
+      recipient: string,
+      assetId: string
+    ): ExpectFungibleTokenMintEvent;
     expectFungibleTokenBurnEvent(
-      amount: Number | bigint,
-      sender: String,
-      assetId: String,
-    ): Object;
+      amount: number | bigint,
```
lgalabru, line 456: Naive question, is moving from `String` to `string` and `Number` to `number`  going to be backward compatible with current developer codebases?

lgalabru (commented):
LGTM thanks @hugocaillard! Just a few questions.
Since we're updating this library, what's your take on this https://github.com/hirosystems/clarinet/issues/54? Should we adopt this approach, or close the issue?
Also, @moodmosaic is there anything new you'd like to see here? Is this colliding with your current workstream?

moodmosaic:
@lgalabru, I looked at the changes in `components/clarinet-deno/index.ts` and they all seem fine to me. 👍

components/clarinet-deno/index.ts
hugocaillard: Yes, it'll probably be mostly transparent.

It could lead to changes in the way that it fixes type checking issues such as:
```ts
const n1: Number = 1;
n1 + 1; // Error: Operator '+' cannot be applied to types 'Number' and 'Number'.

const n2: number = 1;
n2 + 1; // Ok
```

components/clarinet-deno/index.test.ts
lgalabru: Could we also use the `types.x()` helpers for creating some round trips testing?

hugocaillard:
@lgalabru I looked at #54, I think it's about handling it for `clarinet contract new`. I'll do it in future improvements, I really wanted to fix the types here

components/clarinet-deno/index.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+Deno.test("types.buff", () => {
+  assertStrictEquals(types.buff("stx"), "0x737478");
+  assertStrictEquals(types.buff(Int8Array.from([115, 116, 120])), "0x737478");
+  types.buff("stx").expectBuff(Int8Array.from([115, 116, 120]));
+});
+
+Deno.test("types.list", () => {
+  assertStrictEquals(types.list([1, 2, 3]), "(list 1 2 3)");
```
hugocaillard, line 181: @lgalabru 
Small inconsistency here;

`.expectList()` expects something in the form of `"[1, 2, 3]"`.
So `types.list(...).expectList(...)` will always throw (compared to other method were you can chain `types.int(...).expectInt(...)`

I don't things it leading to issue in the real world because `chain.mineBlock` results are in the form `[ ...]

components/clarinet-deno/index.test.ts
```diff
@@ -0,0 +1,187 @@
...
+Deno.test("types.uint", () => {
+  assertStrictEquals(types.uint(2), "u2");
+  assertStrictEquals(types.uint(2n), "u2");
+  types.uint(2).expectUint(2);
+});
+
+Deno.test("types.principal", () => {
+  const addr = "ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5";
+  assertStrictEquals(types.principal(addr), `'${addr}`);
+  types.principal(addr).expectPrincipal(`'${addr}`);
```
hugocaillard, line 167: @lgalabru 
Inconsistency here:
`expectPrincipal` expects a leading `'`, but should it?

(Like, `expectUtf8` doesn't expect a leading `u`)

components/clarinet-deno/index.ts
```diff
@@ -641,17 +658,17 @@ String.prototype.expectTuple = function () {
...
-  for (let element of elements) {
-    for (var i = 0; i < element.length; i++) {
+  const tuple: Record<string, string> = {};
+  for (const element of elements) {
+    for (let i = 0; i < element.length; i++) {
       if (element.charAt(i) === ":") {
-        let key: string = element.substring(0, i);
-        let value: string = element.substring(i + 2, element.length);
+        const key = element.substring(0, i).trim();
+        const value = element.substring(i + 2).trim();
```
hugocaillard, line 671: The `trim`s here could break existing tests (in some rare cases), but probably for the better ;

`expectTuple("{ id: 1 }")` used to return `{ " id": "1 "}`

lgalabru: ah indeed, thanks!

components/clarinet-deno/index.test.ts
lgalabru: Good to know!
Is this a non issue, or do you want to leave it as is? If former, could we add comment about it in the code? If latter, can we create an issue? thanks!

components/clarinet-deno/index.test.ts
lgalabru: Thanks for this tests!
Is this a non issue, or do you want to leave it as is? If former, could we add comment about it in the code? If latter, can we create an issue? thanks!

hugocaillard: To be explored in #627

components/clarinet-deno/index.test.ts
hugocaillard: To be explored in #627

lgalabru (approved):
LGTM, thanks for this upgrade @hugocaillard!

PR #619, opened 2022-10-19
fix: wasm build

hugocaillard:
`reqwest` was not found when building for wasm. Not sure why this regression appeared.

Auto formatted the file

### DIFF
--- a/components/clarinet-deployments/Cargo.toml
+++ b/components/clarinet-deployments/Cargo.toml
@@ -15,7 +15,7 @@ chainhook_types = { package = "chainhook-types", path = "../chainhook-types-rs" 
 stacks_rpc_client = { package = "stacks-rpc-client", path = "../stacks-rpc-client", optional = true }
 
 # CLI
-reqwest = { version = "0.11", default-features = false, optional = true, features = [
+reqwest = { version = "0.11", default-features = false, features = [
     "json",
     "rustls-tls",
 ] }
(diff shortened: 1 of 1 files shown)


obycode (dismissed):
lgtm

components/clarinet-deployments/Cargo.toml
```diff
@@ -12,26 +12,36 @@ serde_yaml = "0.8.23"
...
 clarinet_files = { package = "clarinet-files", path = "../clarinet-files", default-features = false }
 chainhook_types = { package = "chainhook-types", path = "../chainhook-types-rs" }
-stacks_rpc_client = { package = "stacks-rpc-client", path = "../stacks-rpc-client", optional = true  }
+stacks_rpc_client = { package = "stacks-rpc-client", path = "../stacks-rpc-client", optional = true }
 
 # CLI
 reqwest = { version = "0.11", default-features = false, optional = true, features = [
     "json",
     "rustls-tls",
 ] }
```
lgalabru, line 21: if `reqwest` is always used, then we can remove `optional = true` here, and remove `reqwest` from the `wasm` and `onchain` feature gates.

hugocaillard: Alright sounds good I'll do that. I wasn't sure why it was optional in the first place

lgalabru (approved):
Thanks @hugocaillard!

PR #624, opened 2022-10-20
feat: import clarity-jupyter-kernel

lgalabru:
Import https://github.com/hirosystems/clarity-jupyter-kernel as component so it keeps getting passive updates and don't become an abandonware.


hugocaillard (approved):
lgtm

PR #625, opened 2022-10-21
VSCode: version bump and upgrade dependencies

hugocaillard:
Not much to review here. Can be safely merged.
Microsoft recommends to keep vsce always up to date so I applied all minor upgrades


lgalabru (approved):
Thanks @hugocaillard!

Issue #645, opened 2022-11-07
Check-checker should ignore boolean arguments


obycode:
There is no need to check the values of a boolean before its use.

cylewitruk:
Found in:

- Public function definitions
- Private function definitions which are annotated with #[allow(unchecked_params)]

PR #650, opened 2022-11-08
Feat/645 check checker ignore bool

cylewitruk:
Fixes:  #645

### DIFF
--- a/components/chainhook-node/src/block/ingestion.rs
+++ b/components/chainhook-node/src/block/ingestion.rs
@@ -94,7 +94,7 @@ pub fn start(
                 return Err(format!("Redis: {}", message.to_string()));
             }
         };
-        let indexer = Indexer::new(stacks_thread_config.indexer.clone());
+        let _indexer = Indexer::new(stacks_thread_config.indexer.clone());
 
         // Retrieve the former highest block height stored
         let former_tip_height: u64 = con.get(&format!("stx:tip")).unwrap_or(0);
--- a/components/chainhook-node/src/config/mod.rs
+++ b/components/chainhook-node/src/config/mod.rs
@@ -81,7 +81,6 @@ impl Config {
     pub fn expected_redis_config(&self) -> &RedisConfig {
         match self.storage.driver {
             StorageDriver::Redis(ref conf) => conf,
-            _ => panic!("expected redis configuration"),
         }
     }
 
--- a/components/clarity-repl/Cargo.toml
+++ b/components/clarity-repl/Cargo.toml
@@ -54,7 +54,7 @@ memchr = { version = "2.4.1", optional = true }
 pico-args = { version = "0.4.0", optional = true }
 rustyline = { version = "9.1.1", optional = true }
 prettytable-rs = { version = "0.8.0", optional = true }
-hiro_system_kit = { package = "hiro-system-kit", path = "../hiro-system-kit" }
+hiro_system_kit = { package = "hiro-system-kit", path = "../hiro-system-kit", default-features = false }
 
 # WASM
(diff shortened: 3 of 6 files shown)


components/clarity-repl/src/analysis/check_checker/mod.rs
```diff
@@ -371,7 +371,9 @@ impl<'a> ASTVisitor<'a> for CheckChecker<'a, '_> {
         // Upon entering a public function, all parameters are tainted
         if let Some(params) = parameters {
             for param in params {
-                self.add_taint_source(Node::Symbol(param.name), param.decl_span);
+                if param.type_expr.match_atom() != Some(&ClarityName::from("bool")) {
+                    self.add_taint_source(Node::Symbol(param.name), param.decl_span)
+                }
```
obycode, line 376: Let's standardize this for more flexibility in the future. Maybe something like this:

```rust
                match TypeSignature::parse_type_repr(param.type_expr, &mut ()) {
                    Ok(TypeSignature::BoolType) => (),
                    Ok(_) => self.add_taint_source(Node::Symbol(param.name), param.decl_span),
                    _ => {
                        // This error will have already been reported by the type-checker. 
                        // Just stop traversing.
                        return false;
                    }
                }
```

components/clarity-repl/src/analysis/check_checker/mod.rs
```diff
@@ -3023,6 +3025,31 @@ mod tests {
...
 
+    #[test]
+    fn allow_unchecked_bool_in_public_function() {
+        let mut settings = SessionSettings::default();
+        settings.repl_settings.analysis.passes = vec![Pass::CheckChecker];
+        let mut session = Session::new(settings);
+        let snippet = "
+(define-public (allow_tainted (amount bool))
+    (ok amount)
+)
```
obycode, line 3036: This contract would not trigger a check-checker warning now. Returning tainted data from a public function is allowed. How about using the tainted data to set a data var instead:

```clarity
(define-data-var myvar bool false)
(define-public (tainted-var-set (b bool))
    (ok (var-set myvar b))
)
```

obycode (commented):
Thanks for this PR! I have a couple suggestions to improve it.

hugocaillard:
Took the liberty to change the description so that it's linked to the issues (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue)

components/clarity-repl/src/analysis/check_checker/mod.rs
cylewitruk: Ah didn't know that, I updated to use your example.. and learned something 😅

obycode (commented):
LGTM! Thanks for the updates. The new commits came from a different user and only 1 of the users has signed the CLA. Can you fix that by either signing for both or re-committing with the other user and force pushing?

obycode:
Also, please run `cargo fmt` (see [failing check](https://github.com/hirosystems/clarinet/actions/runs/3427688613/jobs/5713338122)).

cylewitruk:
> LGTM! Thanks for the updates. The new commits came from a different user and only 1 of the users has signed the CLA. Can you fix that by either signing for both or re-committing with the other user and force pushing?

Well that didn't turn out so pretty 😅

obycode:
> Well that didn't turn out so pretty 😅

Yeah, I think something went wrong there 😄

cylewitruk:
> > Well that didn't turn out so pretty 😅
> 
> Yeah, I think something went wrong there 😄

Will re-do with a new PR.. easiest since it's so small changes..

Issue #627, opened 2022-10-26
Deno: Test lib improvements ideas


hugocaillard:
Improvement ideas for the clarinet test lib following #618 

- [ ] `.expect<...>Event` methods will always throws the same error messages (`Unable to retrieve expected <...> Event`)  which make them hard to debug (see #540)
- [ ] `.expect<...>Event` methods expects all the event to be emitted from the same contract (see #540)
- [ ] `.expectPrincipal()` expects a leading `'` instead of the raw value, which is inconsistent with other `expect<Type>()` methods ([comment)](https://github.com/hirosystems/clarinet/pull/618#discussion_r1000422603)
- [ ] `.expectList()` expects something in the form of `"[1, 2, 3]"`, so types.list(...).expectList(...) will always throw. Which is inconsistent with other `expect<Type>()` methods ([comment)](https://github.com/hirosystems/clarinet/pull/618#discussion_r1000421230)

lgalabru:
Related: https://github.com/hirosystems/clarinet/issues/258

hugocaillard:
Closing in favor of clarinet-sdk https://github.com/hirosystems/clarinet/issues/1065

PR #651, opened 2022-11-08
Fix: deno expect events

hugocaillard:
Fixes a regression introduced in #618.
Because the "events" array can contain anything (not just the expected event type) events should not be destructured outside of the `try catch`.

Note: further improvements should be expected in `expect<Event>` methods as part of #627 to make error messages easier to understand.


hugocaillard:
Thanks @moodmosaic for spotting it

lgalabru (approved):
Thanks for the fix and the new unit test!

Issue #630, opened 2022-10-28
LSP: introduce "current_editor_contract_state"


hugocaillard:
The Clarity LSP fully relies on the `EditorState`, which represents the state of all the contracts (and manifest) on the **File System**. For some features, we want to consider the content of the file currently opened in the editor, **including the unsaved changes**.

## Current behavior
Every time the LSP client sends a request or a notification, the server will try to acquire read or write on the sate `RwLock<EditorState>`.  Notifications and Requests are even competing for it.
In the first naive implementation of the DocumentSymbols request (#626) the idea was to rely on the `EditorState`, leading to significant number of request try to read the state, and potentially outdated data if the file contains unsaved changes.

## Expected behavior
- Some request such as Document Symbols should rely on up to date data and not on possibly outdated data from the file system.
- On Server initialization, the Client shouldn't have to wait for the whole state to be computed to send requests

## Solution
Along with the `editor_state_lock`, implement a `current_editor_state_lock` that only keeps track of the file being currently modified, including the **unsaved change**.
This current state would be updated on `DidOpenTextDocument` and `DidChangeTextDocument`.
It will be requested on Document Symbols requests, Inlay Hints requests.
Advantages:
- Symbols and Inlay Hints will live (up to date we the current changes being made on the document)
- Less competition to acquire read on the `editor_state_lock`
- Symbols, Hover, Inlay hints (and probably more requests would give up to date data based in the current state of an edited/unsaved contract

## Risks
- How does `DidOpenTextDocument` when multiple contracts are open side by side?
- The concurrency when update `editor_state_lock` and `current_editor_state_lock` might lead to some minor issues
- If the contract has dependencies, we might still need to read the `editor_state` to get Inlay Hints data

lgalabru:
Now that we have a working full loop, I think that we starting to think more about "partial analysis" (tracked here https://github.com/hirosystems/clarinet/issues/604) could be a great avenue of optimization.

hugocaillard:
I think #604 is an optimization of the current behavior, that's is needed, and kind of pairs with this one but these are slightly different.

This one is about differentiating dirty files from saved ones.
Let's say _contract1.clar_ exposes a `get_id` function, called from other contracts. If the function is renamed to `get_ids` (but unsaved), we don't want to display diagnostic errors in other contracts (until _contract1.clar_ is saved).
But the auto-completion request, within _contract1.clar_ should provide `get_ids`.

That's the default behavior of rust-analyzer.
[...]

Issue #626, opened 2022-10-21
LSP: implement symbols


hugocaillard:
Implement request symbols capabilities on the server.
It allow to build other features such as go to definition, replace all, better code navigation, etc.

Issue #610, opened 2022-01-26
LSP: add documentation for builtin functions and keywords on mouse-over


obycode:

hugocaillard:
Preview

<img width="952" alt="Screenshot 2022-11-08 at 14 22 24" src="https://user-images.githubusercontent.com/911307/200576190-0dc0a44e-2b0e-45ef-ade0-b4b2c77a7959.png">

PR #652, opened 2022-11-08
Feat: lsp current editor state

hugocaillard:
Resolves: #630

This PR does not introduce reals changes for Clarity developers, instead it prepares the LSP for #610 and #626 by keeping track of the state of the files being modified (but not saved).

It'll bring nice addition better performances and more reliability by having the editor_state being locked for smaller period of times and less often (thanks @lgalabru for the help on that) with the introduction of
```rust
pub struct EditorStateInput {
    editor_state: Option<EditorState>,
    editor_state_lock: Option<Arc<RwLock<EditorState>>>,
}
```

~Since the AST is now computed every time a contract is changed, I was able to display diagnostic errors for invalid AST in real time~ I removed it because it overrides other diagnostics (warning and type errors). Will spend more time on that

### DIFF
--- a/components/clarinet-cli/Cargo.toml
+++ b/components/clarinet-cli/Cargo.toml
@@ -136,7 +136,7 @@ strum = { version = "0.23.0", features = ["derive"] }
 bitcoin = "0.28.1"
 segment = { version = "0.1.2", optional = true }
 mac_address = { version = "1.1.2", optional = true }
-tower-lsp = { version = "0.14.0", optional = true }
+tower-lsp = { version = "0.17.0", optional = true }
 hex = "0.4.3"
 serde_yaml = "0.8.23"
 chainhook_event_observer = { package = "chainhook-event-observer", default-features = false, path = "../chainhook-event-observer" }
--- a/components/clarinet-files/src/project_manifest.rs
+++ b/components/clarinet-files/src/project_manifest.rs
@@ -135,13 +135,14 @@ impl ProjectManifest {
         }
 
         let project_name = project_manifest_file.project.name;
-        let mut project_root_location = manifest_location.get_parent_location()?;
+        let project_root_location = manifest_location.get_parent_location()?;
         let cache_location = match project_manifest_file.project.cache_dir {
             Some(ref path) => FileLocation::try_parse(path, Some(&project_root_location))
                 .ok_or(format!("unable to parse path {}", path))?,
             None => {
-                project_root_location.append_path(".cache")?;
-                project_root_location.clone()
+                let mut cache_location = project_root_location.clone();
+                cache_location.append_path(".cache")?;
+                cache_location
             }
         };
 
@@ -257,6 +258,7 @@ impl ProjectManifest {
                             };
                             config_contracts
                                 .insert(contract_name.clone(), clarity_contract.clone());
+
                             let mut contract_location = project_root_location.clone();
                             contract_location.append_path(contract_path)?;
                             contracts_settings.insert(contract_location, clarity_contract);
--- a/components/clarinet-files/src/project_manifest.rs
+++ b/components/clarinet-files/src/project_manifest.rs
@@ -14,6 +14,14 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use toml::value::Value;
 
+#[derive(Deserialize, Debug, Clone)]
+pub struct ClarityContractMetadata {
+    pub name: String,
+    pub deployer: ContractDeployer,
+    pub clarity_version: ClarityVersion,
+    pub epoch: StacksEpochId,
+}
+
 #[derive(Serialize, Deserialize, Debug)]
 pub struct ProjectManifestFile {
     project: ProjectConfigFile,
@@ -46,7 +54,7 @@ pub struct ProjectManifest {
     #[serde(skip_serializing)]
     pub location: FileLocation,
     #[serde(skip_serializing)]
-    pub contracts_settings: HashMap<FileLocation, ClarityContract>,
+    pub contracts_settings: HashMap<FileLocation, ClarityContractMetadata>,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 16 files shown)


components/clarity-lsp/src/vscode_bridge.rs
```diff
@@ -114,54 +83,83 @@ impl LspVscodeBridge {
...
             _ => {
                 #[cfg(debug_assertions)]
                 log!("unexpected notification ({})", method);
+                return Promise::resolve(&JsValue::FALSE);
             }
-        }
+        };
 
-        return Promise::resolve(&JsValue::NULL);
+        let mut editor_state = self.editor_state.clone();
```
hugocaillard, line 133: I tried to avoid the `mut` here because it seems to be unnecessary to do it on the  RwLock.
But `process_notification` needs to accept `&mut EditorState` since the `native_bridge` directly passes a mutable reference of the editor_state.

Maybe there's a way around in the way EditorStateInput is implemented but could find it

components/clarity-lsp/src/vscode_bridge.rs
```diff
@@ -22,10 +25,10 @@ use wasm_bindgen_futures::future_to_promise;
 
 #[wasm_bindgen]
 pub struct LspVscodeBridge {
-    editor_state_lock: Arc<RwLock<EditorState>>,
     client_diagnostic_tx: JsFunction,
     _client_notification_tx: JsFunction,
     backend_to_client_tx: JsFunction,
+    editor_state: EditorStateInput,
```
lgalabru, line 31: What's your take on keeping the previous design here, and construct a `EditorStateInput::RwLock(self. editor_state_lock.clone())` when you send it to `process_notification` ?

components/clarity-lsp/src/vscode_bridge.rs
lgalabru: I got confused with the naming here, see suggestion above.

lgalabru (commented):
I love this new design, it's looking great!
A few feedbacks to address before merging!

components/clarity-lsp/src/common/backend.rs
```diff
@@ -3,161 +3,256 @@ use crate::state::{build_state, EditorState, ProtocolState};
...
+            // store the file in the active_contracts map
+            if !editor_state.try_read(|es| es.active_contracts.contains_key(&contract_location))? {
+                let contract_source = match file_accessor {
+                    None => contract_location.read_content_as_utf8(),
+                    Some(file_accessor) => {
+                        file_accessor.read_file(contract_location.to_string()).await
+                    }
+                }?;
+
+                let clarity_version = DEFAULT_CLARITY_VERSION;
```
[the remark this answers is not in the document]
hugocaillard: I already have clarity_version in `ActiveContractData`
`ContractState` as it under `analysis: Option<ContractAnalysis>`

Should I had epoch as well to both? Not sure where to get the value from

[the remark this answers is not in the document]
hugocaillard: @lgalabru The key of the contracts in the `manifest.contracts` map are the contracts names (which is user defined in the manifest).
Not sure how to get it from here 🤔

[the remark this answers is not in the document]
lgalabru: `contract_id.name` should do the trick?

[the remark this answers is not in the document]
hugocaillard: I don't have access to contract_id herr. I'll figure something out

components/clarinet-cli/Cargo.toml
```diff
@@ -136,7 +136,7 @@ strum = { version = "0.23.0", features = ["derive"] }
 bitcoin = "0.28.1"
 segment = { version = "0.1.2", optional = true }
 mac_address = { version = "1.1.2", optional = true }
-tower-lsp = { version = "0.14.0", optional = true }
+tower-lsp = { version = "0.17.0", optional = true }
```
hugocaillard, line 139: Update tower-lsp so that it uses the latest version of lsp_types

components/clarinet-files/src/project_manifest.rs
```diff
@@ -243,16 +249,19 @@ impl ProjectManifest {
...
+                                deployer,
+                                clarity_version,
+                                epoch,
+                            };
+                            config_contracts
+                                .insert(contract_name.clone(), clarity_contract.clone());
+
+                            let mut contract_location = project_root_location.clone();
+                            contract_location.append_path(contract_path)?;
+                            contracts_settings.insert(contract_location, clarity_contract);
```
hugocaillard, line 264: `contract_settings` basically hold the same data as `config_contracts` but with the location as the key.

components/clarity-lsp/src/common/state.rs
```diff
@@ -240,6 +317,30 @@ impl EditorState {
...
         (contracts, tldr)
     }
+
+    pub fn insert_active_contract(
+        &mut self,
+        contract_location: FileLocation,
+        clarity_version: ClarityVersion,
+        source: &str,
+    ) {
+        let epoch = StacksEpochId::Epoch21;
```
hugocaillard, line 327: The LSP already enforces Epoch21 to compute diagnostic, it should be safe to enforce it here as well

components/clarinet-files/src/project_manifest.rs
```diff
@@ -243,16 +249,19 @@ impl ProjectManifest {
...
-                            );
+                            let clarity_contract = ClarityContract {
+                                name: contract_name.clone(),
+                                code_source,
+                                deployer,
+                                clarity_version,
+                                epoch,
+                            };
+                            config_contracts
+                                .insert(contract_name.clone(), clarity_contract.clone());
```
lgalabru, line 260: I think there is some room for memory optimization here, with these extra-clones.
Could we add a todo / or create an issue, and get back to it when you feel more confortable with lifetimes or if we feel like performance is becoming a bottleneck?
Thanks!

lgalabru (approved):
LGTM, thanks @hugocaillard!

components/clarinet-files/src/project_manifest.rs
hugocaillard: Partially addressed, more improvements to come as part of #668

components/clarity-lsp/src/common/backend.rs
```diff
@@ -1,163 +1,316 @@
...
 use serde::{Deserialize, Serialize};
+use std::sync::{Arc, RwLock};
+
+#[derive(Debug, Clone)]
+pub enum EditorStateInput {
+    Owned(EditorState),
+    RwLock(Arc<RwLock<EditorState>>),
+}
+
+fn try_read<F, R>(editor_state_input: &EditorStateInput, closure: F) -> Result<R, String>
```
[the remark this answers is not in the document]
hugocaillard: Indeed, thanks 🔥

Issue #557, opened 2022-10-04
Github action updates for Clarinet


saralab:
Guide
https://www.cloudwithchris.com/blog/automate-adding-gh-issues-projects-beta/

lgalabru:
Implement https://hiropbc.slack.com/archives/CEK6NUGHJ/p1664788525588889

lgalabru:
Todo
- [ ] Setup board automation (@aravindgee)
- [ ] Re-enable code coverage (@lgalabru)
- [ ] Optimize build time with this simple trick https://hiropbc.slack.com/archives/CEK6NUGHJ/p1664788525588889  (@aravindgee)
- [ ] Create release workflow for `components/clarity-vscode` - inspired from https://github.com/hirosystems/clarity-lsp  (@aravindgee)
- [ ] Automate releases of `components/chainhook-event-observer` on crates.io  (@aravindgee)
- [ ] Re-enable MUSL builds after handling https://github.com/hirosystems/clarinet/issues/560 (@lgalabru)
- [ ] Automate releases of `components/stacks-devnet` after handling https://github.com/hirosystems/clarinet/issues/560  (@lgalabru)

aravindgee:
Reviewed the status quo with Ludo, working on "Must" items first.

aravindgee:
**Optimize build time** - @lgalabru Clarinet build time is already optimized through the use of a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) ([example](https://github.com/hirosystems/clarinet/actions/runs/3394594400)).

Current total build time is close to 1hr, is this acceptable? If not, what would be an area for improvement?

hugocaillard:
Could it be feasible / desirable to only triggers the actions impacted by a change.
For example, an update to the clarity-vscode component code has no impact on clarinet or any of the rust code and we don't need to build it

aravindgee:
**todo**: Optimize build time for clarinet packages - investigate additional paths for improving build time - eg. parallelize by platform (h/t @lgalabru ) and docker image generation.

<img src="https://user-images.githubusercontent.com/3316191/206347828-2cc28b72-58d0-4651-80b2-15e4895b0d63.png" width="500" />

aravindgee:
Reviewed build time bottleneck with Ludo - since its not a major concern, we agreed to deprioritize this for the time being.

Additional concerns brought up by Hugo - 

[...]

PR #654, opened 2022-11-09
Release workflow for components/clarity-vscode

aravindgee:
Automating release of Clarity VSCode - based on older repo [hirosystems/clarity-lsp](https://github.com/hirosystems/clarity-lsp)

[[Related Issue](https://github.com/hirosystems/clarinet/issues/557)]


hugocaillard:
Thanks @aravindgee 🙌 
2 questions:

1. Now we also release on Open VSX, could a step be added to do so?
The command is pretty similar to vsce:
```
$ npx ovsx publish <file> -p <token>
```

aravindgee:
> Thanks @aravindgee 🙌 2 questions:
> 
> 1. What will trigger the release of a new version?
> 2. Now we also release on Open VSX, could a step be added to do so?
>    The command is pretty similar to vsce:
> 
> ```
> $ npx ovsx publish <file> -p <token>
> ```

1. Hi Hugo - The release is triggered when a push tagged with v** (version #) is detected in main or develop branches. This was adapted from the existing publishing logic in the `clarity-lsp` repo.

2. Also added the ability to publish to Open VSX. The action will require two secrets to be added to the repo - `MARKETPLACE_TOKEN` (for the VSCode Marketplace), `OVSX_TOKEN` (for Open VSX). I do not have the privileges for this, could you make the update?

lgalabru (approved):
Thanks @aravindgee! I think the release of new `v` tag make sense, because it's giving us a way to simplify a manual release flow for our release candidates `rc-x` (see release playbook). What do you think @hugocaillard?
Could you approve this PR and merge once the required secrets are present in the repo? Thanks!

lgalabru:
Ah, I actually missed the build error, @aravindgee could we get back to this once the chainhook-cloud deployment is behind us? Thanks!

lgalabru (changes requested):
Ah, I actually missed the build error, @aravindgee could we get back to this once the chainhook-cloud deployment is behind us? Thanks!

lgalabru:
Note: this action will also have to build the clarity-lsp wasm. 
```bash
cd components/clarity-lsp
wasm-pack build --target web  -- --no-default-features --features wasm
```

hugocaillard:
@lgalabru The wasm-pack build is already implemented in the node script. It's even [configured in webpack](https://github.com/hirosystems/clarinet/blob/b0f4b116a4de91402ee981a0aa62a3a6bb57ff6e/components/clarity-vscode/webpack.config.js#L91-L96). So only 1 command to run in dev mode or for the production build

hugocaillard:
@lgalabru 

> think the release of new v tag make sense, because it's giving us a way to simplify a manual release flow for our release candidates 

Does this mean we would sync the extensions releases with the clarinet releases?

hugocaillard:
What are the best practices regarding the tokens (vscode marketplace + ovsx)? Can I set tokens linked to my own accounts?

hugocaillard:
What are the best practices regarding the tokens (vscode marketplace + ovsx)? Can I set tokens linked to my own accounts?

lgalabru:
> Does this mean we would sync the extensions releases with the clarinet releases?

Releasing clarinet would induce a release of the extension, but we would have a way to send releases of the extension whenever. 
Do you foresee an issue with that flow?

aravindgee:
@lgalabru Would now be a good time to revisit this PR? I believe this automation has Q4 as the targeted date of completion; just making sure that we have enough time to finalize it. No rush otherwise!

lgalabru:
hey @aravindgee, yes, let's do this, thanks!

.github/workflows/release-clarity-vscode.yaml
```diff
@@ -4,10 +4,8 @@ on:
   push:
     tags:
       - 'v*'
-    paths-ignore:
-      - '**/CHANGELOG.md'
-      - '**/package.json'
-      - '**/package-lock.json'
+    paths:
+      - components/clarity-vscode
```
lgalabru: Really cool trick, I'll keep it in mind.
We would need to include a few other paths:
- components/clarity-repl
- components/clarity-lsp
- components/clarinet-files
- components/clarinet-deployments

aravindgee: Handled in PR #684

aravindgee:
Regarding build time - the runner/container used by Github actions to run the build seems to be restricting the compiler's performance. Build completes in a few minutes when tested on externally hosted containers/VMs. 

This is a bigger bottleneck than clarinet's internal dependencies, checking to see if there's any customization available to improve runner's performance.

aravindgee:
Added: [Build clarity-vscode on changes to clarinet.](https://github.com/hirosystems/clarinet/pull/654/commits/ba8d1899d33e911d48b12e6f679fb10347385567)

@lgalabru are we good to merge this PR?

.github/workflows/release-clarity-vscode.yaml
```diff
@@ -0,0 +1,108 @@
+name: Build Release for Clarity VSCode
+on:
+  pull_request:
+  push:
+    tags:
+     - 'v*'
```
lgalabru, line 6: Could we use a prefix the git tag with a namespace, so something like `vscode-v*`?

.github/workflows/release-clarity-vscode.yaml
```diff
@@ -0,0 +1,108 @@
...
+     - '**/CHANGELOG.md'
+     - '**/package.json'
+     - '**/package-lock.json'
+    branches:
+     - main
+     - develop
+    paths:
+     - components/clarity-vscode
+     - components/clarinet-cli
+     - components/chainhook-event-observer
```
lgalabru, line 17: Paths are incorrect

.github/workflows/release-clarity-vscode.yaml
hugocaillard: That would be super useful indeed.
Not sure if it's super common to handle multiple release / version flow with a tag like that but this would keep a nice developer experience (thanks to the single repo) and add automatic CD

.github/workflows/release-clarity-vscode.yaml
hugocaillard: Here's the complete list:
```
components/clarity-lsp
components/chainhook-types
components/clarinet-deployments
components/clarinet-files
components/clarinet-utils
components/clarity-repl
components/hiro-system-kit
components/stacks-rpc-client
```

hugocaillard: And `components/clarity-vscode` (which is the component we want to build)

.github/workflows/release-clarity-vscode.yaml
hugocaillard: cc @aravindgee, thoughts?

smcclellan:
@aravindgee @lgalabru What's next with this PR?

hugocaillard:
Can be close in favor of #1078

PR #657, opened 2022-11-09
feat: chainhook-node replay mode

lgalabru:
In this PR, we are modifying the developer experience a bit by adding a new mode:
```bash
$ chainhook-node replay --testnet
```
letting developers replay devnet / testnet / mainnet dumps.
I will continue to refactor and tweak the CLI interface in a subsequent PR, but the current approach is already very useful.


hugocaillard (approved):
lgtm!

PR #662, opened 2022-11-10
Cli.rs file: updated links to point to the right docs location

LakshmiLavanyaKasturi:
Updated cli.rs file to point to the correct docs location for Clarinet testing.


lgalabru (approved):
Looking good, thanks @LakshmiLavanyaKasturi!

PR #664, opened 2022-11-11
feat: introducing clarity-fmt

lgalabru:
Adding a boilerplate for a `clarity-fmt` component, in case someone wants to explore the idea of writing a clarity linter.
For working on this, we would have new component `clarity-fmt` that would be compilable as binary and a library.
A basic structure for creating a CLI was added, so currently
```bash
clarity-fmt fix test.clar
```
would load `test.clar` in memory, analyse it and submit the analysis to a ClarityLinter visitor. A basic backbone is available it needs to be implemented.

This component would also be available as a library, that the vscode extension can rely on for "right click > format document".
That should be a fun project!


hugocaillard:
I'm confused if this is a linter or a formatter.

The linter should help to detect bad practices or invalid code. We already do it in some ways but it could be enhanced: detect unused variables/methods, enforce naming case, auto-fix some bug (like uint vs int), detect useless `begin` etc.

The formatter should handle max line length, trailing white-spaces, line breaks, tabs vs spaces, tabs length, etc. It could be done by having the possibility to convert an AST into clean Clarity code, but the output code should have the exact same AST as the input one, so that it's considered safe to auto format code

lgalabru:
Closing in favor of #799

PR #672, opened 2022-11-16
fix: clarity version and epoch parsing in manifest

hugocaillard:
Having the wrong data type for clarity version or epoch would fallback to default instead of throwing an error.

example
```toml
[contracts.counter]
clarity_version = "2" # should be integer
epoch = 3 # should be float
```
Would silently default to clarity 1 (default) and epoch 2.0 (default)

`epoch` is now expecting a float and wrong format will return errors

### DIFF
--- a/components/clarinet-files/src/project_manifest.rs
+++ b/components/clarinet-files/src/project_manifest.rs
@@ -14,6 +14,9 @@ use std::path::PathBuf;
 use std::str::FromStr;
 use toml::value::Value;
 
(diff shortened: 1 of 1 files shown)


obycode (approved):
Thanks! 🚀

obycode (dismissed):
LGTM!

obycode (approved):

lgalabru (approved):
Thanks @hugocaillard!

PR #675, opened 2022-11-17
fix: re-implement `::reload` command in REPL

obycode:
Fixes: #637


obycode:
Bonus with this new implementation over the old way (in addition to it being really simple) is that it does a complete reload, so even changes in Clarinet.toml are captured.

hugocaillard (approved):
lgtm!

Issue #403, opened 2022-06-08
LSP: Implement go to definition


jonkho:
Feature description: A ‘jump to function definition’ feature so that as code is written we can jump to the signature  and read the code. And a “jump back" feature to resume where you just left off.

PR #676, opened 2022-11-18
Feat: lsp go to definition

hugocaillard:
Resolves: #403

- in same contract
  - go to `define-*`
    - [x] go to function definition (in same contract)
    - [x] go to `define-constant`
    - [x] go to `define-data-var`
    - [x] go to `define-map`
  - [x] go to function args
  - [x] go to let bindings
  - [x] go to match binding
- in other contract
  - [x] go to contract-call function definition `define-public` `define-read-only`

#### constant
Screenshot 2022-11-18 at 19 10 08

#### argument
Screenshot 2022-11-18 at 19 10 48

##### peek definition
Screenshot 2022-11-18 at 19 11 18


components/clarity-lsp/src/common/requests/definitions.rs
```diff
@@ -0,0 +1,289 @@
...
+) -> Option<Location> {
+    let search = get_expression_name_at_position(position, expressions, None)?;
+
+    if let Some(location) = find_global_definition(&search, location, expressions) {
+        return Some(location);
+    }
+
+    if let Some(location) = find_local_definition(&search, location, expressions, position) {
+        return Some(location);
+    }
```
obycode, line 227: Instead of traversing the AST each time, could you traverse once and save a set of scopes and their definitions, then quickly reference that for each query?

components/clarity-lsp/src/common/requests/definitions.rs
```diff
@@ -0,0 +1,701 @@
...
+pub enum DefinitionLocation {
+    Internal(Range),
+    External(QualifiedContractIdentifier, ClarityName),
+}
+
+#[derive(Clone, Debug)]
+pub struct Definitions {
+    pub tokens: HashMap<(u32, u32), DefinitionLocation>,
+    global: HashMap<ClarityName, Range>,
+    local: HashMap<u64, HashMap<ClarityName, Range>>,
```
obycode, line 24: Can you add some documentation comments to these so that it's easier to understand what they are without reading the implementation?

obycode (dismissed):
Looks great! I just had a few requests for some documentation to make it easier to understand.

obycode (approved):
LGTM! Thanks @hugocaillard 🚀

PR #677, opened 2022-11-22
feat: Add `epoch` field to batch in deployment plans

obycode:
If an `epoch` is specified for a batch of transactions in the deployment
plan, then this batch will not be deployed until that epoch is live.
Note that the batches are still processed in order, so if a batch that
specifies `2.1` is followed by a batch that specifies `2.05`, that
second batch will run in 2.1.

Example:

```yaml
---
id: 0
name: Devnet deployment
network: devnet
stacks-node: "http://localhost:20443"
bitcoin-node: "http://devnet:devnet@localhost:18443"
plan:
  batches:
    - id: 0
      epoch: 2.05
      transactions:
        - contract-publish:
            contract-name: early
            expected-sender: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM
            cost: 830
            path: contracts/early.clar
            anchor-block-only: true
    - id: 1
      epoch: 2.1
      transactions:
        - contract-publish:
            contract-name: test
            expected-sender: ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM
            cost: 1180
            path: contracts/test.clar
            anchor-block-only: true
            clarity-version: 2
```

### DIFF
--- a/components/clarinet-deployments/src/onchain/mod.rs
+++ b/components/clarinet-deployments/src/onchain/mod.rs
@@ -33,7 +33,7 @@ use libsecp256k1::{PublicKey, SecretKey};
 
 mod bitcoin_deployment;
 
-use crate::types::{DeploymentSpecification, TransactionSpecification};
+use crate::types::{DeploymentSpecification, EpochSpec, TransactionSpecification};
 
 fn get_btc_keypair(
     account: &AccountConfig,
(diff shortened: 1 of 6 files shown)


components/clarinet-deployments/src/onchain/mod.rs
```diff
@@ -687,7 +690,31 @@ pub fn apply_on_chain_deployment(
     // Phase 2: we submit all the transactions previously encoded,
     // and wait for their inclusion in a block before moving to the next batch.
     let mut current_block_height = 0;
-    for batch in batches.into_iter() {
+    for (epoch, batch) in batches.into_iter() {
+        // Ensure we've reached the appropriate epoch for this batch
+        let after_block = match epoch {
+            EpochSpec::Epoch2_05 => network_manifest.devnet.as_ref().unwrap().epoch_2_05,
+            EpochSpec::Epoch2_1 => network_manifest.devnet.as_ref().unwrap().epoch_2_1,
+        };
```
lgalabru, line 698: the `apply_onchain_deployment` is also called by the command 
```bash
clarinet deployments apply --mainnet
```
and 
```bash
clarinet deployments apply --testnet
```

which would be an issue with this approach.

obycode: Would it make sense to developers if clarinet just ignores the specified epoch when applying a deployment plan? I think just reporting a warning if epoch is specified would be reasonable. Something like:

```
Warning: `epoch` specified for a deployment batch is ignored when applying a deployment plan. This field should only be specified for deployments plans used to launch a devnet with `clarinet integrate`.
```

lgalabru: works for me!

lgalabru (approved):
LGTM, thanks @obycode!

PR #678, opened 2022-11-22
fix: repl was build issue

hugocaillard:
Regression caused by handle_command signature change in #677


components/clarity-repl/src/repl/session.rs
```diff
@@ -235,6 +235,7 @@ impl Session {
 
     pub fn handle_command(&mut self, command: &str) -> (bool, Vec<String>) {
         let mut output = Vec::<String>::new();
+        #[allow(unused_mut)]
```
hugocaillard, line 238: `reload` is only mutated with `config="cli"`

lgalabru (approved):
Thanks @hugocaillard!

Issue #556, opened 2022-10-04
LSP: Contracts can be analysed multiple times


hugocaillard:
It's possible to declare multiple contracts identifiers with the same source file in the Clarinet.toml

```toml
[contracts.counter-1]
path = "contracts/counter.clar"

[contracts.counter-2]
path = "contracts/counter.clar"
```

In this case, the "manifest.contracts" will have twice the same source code, and we'll compute the analysis twice.
https://github.com/hirosystems/clarinet/blob/ed323ec4edbb70647fec5a0040b83d0c95ca81d3/components/clarinet-deployments/src/lib.rs#L497

It leads to some performances issues, for instance, here the `.get()` should be replaces with `.remove()` to avoid `.clone()`

### Potential solutions
- emit an error if a path is used multiple times in a manifest
- rework the iteration to avoid analyzing twice the same source

whoabuddy:
Hey @hugocaillard happy to add some color there.

The use case in that scenario was to create duplicate contracts in order to simulate an upgrade in the unit tests, where the initial contracts would be deployed and activated, then the core contract upgraded from core-v1, core-v1-1, core-v1-2 then running test scenarios run afterward.

Here is an example of one of the tests that goes through the core contract upgrade process using those:
https://github.com/citycoins/contracts/blob/develop/tests/cities/mia/upgrade/miamicoin-core-v1-v2.test.ts#L36

This is going to be simplified in future iterations and the protocol is being restructured in [citycoins/protocol](https://github.com/citycoins/protocol), so we can avoid that pattern if it's a problem.

radicleart:
Another example - again for City Coins - say there are 1000 cities all with a separate vault for each treasury but the same underlying contract (the same contract deployed 1000 times with different names). Support for this in Clarinet is exactly as in the example above..

[contracts.treasury-city1]
path = "contracts/treasury.clar"

[contracts.treasury-city2]
path = "contracts/treasury.clar"

is a legitimate requirement for declaring multiple contract identifiers with the same source file in the Clarinet.toml ?

whoabuddy:
Looks like something happened between `1.0.6` and `1.1.0` that causes contracts with this type of definition to fail `clarinet check`. [See this comment for more information](https://github.com/citycoins/protocol/pull/19#issuecomment-1325431794), and I confirmed it locally after upgrading as well.

I didn't see anything in the changelog - and per the comments above this is something we could actually use as some contract structures will be a template with multiple deployments.

[...]

PR #680, opened 2022-11-24
fix: use contracts name as index in deployements

hugocaillard:
Address regressions discovered here https://github.com/hirosystems/clarinet/issues/556#issuecomment-1325439110
Fix for Citycoins issue : https://github.com/citycoins/protocol/pull/19#issuecomment-1325431794

This PR reverts a change introduced in https://github.com/citycoins/protocol/pull/19#issuecomment-1325431794
The `manifest.contracts_settings` was introduced because the LSP needed the settings metadata indexed by file location.
But it was necessary to use it in deployments, it even broken it because multiple contracts name can refer to the same file.

Note: I also included a small fix on the LSP e3051c42c375151d907df908f481c307831783a0 (it triggered an error message when clarinet.toml files were opened)

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -489,10 +489,22 @@ pub async fn generate_default_deployment(
     let mut contracts = HashMap::new();
     let mut contracts_sources = HashMap::new();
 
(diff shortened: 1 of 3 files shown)


lgalabru (approved):
LGTM, thanks @hugocaillard!

PR #708, opened 2022-12-08
fix: keyword matching for go to definition

hugocaillard:


obycode (approved):
Thanks @hugocaillard! LGTM

Issue #709, opened 2022-12-08
Bug testing Clarity 2 contracts in v1.2.0


hstove:
Previously, I was able to write and test Clarity 2 contracts using Clarinet without issues. However, with 1.2.0, there's been a regression. When running a test that calls a Clarity 2 function, you get the error:

```
Unchecked(CostComputationFailed("Error evaluating result of cost function ST000000000000000000002AMW42H.costs.cost_slice: Unchecked(UndefinedFunction(\"cost_slice\"))"))
```

Not just for `slice?`, but for all Clarity 2 functions.

I have a demo project that reproduces this: https://github.com/hstove/clarinet-repro-cost-bug

Run `clarinet test tests/tester_test.ts` to repro

lgalabru:
Addressed with `v1.3.0`. Thanks for your help on this @hstove!

PR #711, opened 2022-12-08
fix contract calls into clarity2 contracts

hstove:
This is a fix for #709 - check out that issue for more details and a repro of the bug.

This PR does two things:
- Adds `costs-3` boot contract
  - I pulled the code in from `stacks-blockchain` and added it to various "default boot code" places, including in the project manifest
- When making a `contract-call?` in `clarity-repl`, I set the epoch to 2.1.

The latter fix was the only thing that actually fixed the issue. The exact line of code is here: https://github.com/hstove/clarinet/blob/0ca1d66cb47e39763061a9017f7e45771938c896/components/clarity-repl/src/repl/session.rs#L593 . Without this line, my issue still fails.

One thing I noticed, via some debugging, was that my contract was being added to the Clarity DB with epoch "2.0" in tests, even though I have it specified as 2.1 in my Clarinet.toml. If that bug was fixed (which I haven't figured out), would that fix the bug as well?

### DIFF
--- a/components/clarinet-cli/src/runner/api_v1.rs
+++ b/components/clarinet-cli/src/runner/api_v1.rs
@@ -28,6 +28,7 @@ use chainhook_types::StacksTransactionMetadata;
 use chainhook_types::TransactionIdentifier;
 use chrono::{DateTime, NaiveDateTime, Utc};
 use clarinet_deployments::update_session_with_contracts_executions;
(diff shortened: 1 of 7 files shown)


lgalabru (approved):
LGTM, thank you for this contribution @hstove!
As discussed off band, some additional adjustments are required, I'll merge this first patch and open a subsequent PR.

PR #713, opened 2022-12-08
feat: signature help

hugocaillard:
Screenshot 2022-12-08 at 22 03 51


components/clarity-lsp/src/vscode_bridge.rs
```diff
@@ -164,14 +165,15 @@ impl LspVscodeBridge {
...
+        let serializer = Serializer::json_compatible();
         match method.as_str() {
             Initialize::METHOD => {
                 let lsp_response = process_request(
                     LspRequest::Initialize(decode_from_js(js_params)?),
                     &EditorStateInput::RwLock(self.editor_state_lock.clone()),
                 );
                 if let LspRequestResponse::Initialize(response) = lsp_response {
-                    return encode_to_js(&response).map_err(|_| JsValue::NULL);
+                    return response.serialize(&serializer).map_err(|_| JsValue::NULL);
```
hugocaillard, line 176: The regular `to_value()` (`encode_to_js`)  methods converts Rust structs to native JS maps.
But in this case, the lsp client expects JS literal objects, which is what  `let serializer = Serializer::json_compatible();` does

components/clarinet-cli/src/lsp/native_bridge.rs
```diff
@@ -181,6 +182,21 @@ impl LanguageServer for LspNativeBridge {
...
     }
 
+    async fn signature_help(&self, params: SignatureHelpParams) -> Result<Option<SignatureHelp>> {
+        let _ = match self.request_tx.lock() {
+            Ok(tx) => tx.send(LspRequest::SignatureHelp(params)),
+            Err(_) => return Ok(None),
+        };
+
+        let response_rx = self.response_rx.lock().expect("failed to lock response_rx");
+        let ref response = response_rx.recv().expect("failed to get value from recv");
```
lgalabru, line 192: Generally speaking I would not panic if we can't get the lock, just log and return None.

hugocaillard: I'll do the changes in the PR I'm currently working on (and for all the requests) @lgalabru

Issue #725, opened 2022-12-14
Jump to definition for function used in `fold` and `map`


obycode:
The new jump to definition functionality doesn't work in these two cases:

```clarity
(define-private (concat-string (a (string-ascii 20)) (b (string-ascii 20))) (unwrap-panic (as-max-len? (concat a b) u20)))
(fold concat-string "cdef" "ab")   ;; Returns "fedcab"

(define-private (a-or-b (char (string-utf8 1))) (if (is-eq char u"a") u"a" u"b"))
(map a-or-b u"aca") ;; Returns (u"a" u"b" u"a")
```

PR #728, opened 2022-12-15
Fix: go to definition iterators

hugocaillard:
Fix: #725 
The first commit (39867c26e130ddc70aaf8e074e0f865f86cfd965) add go to definition for map, filter and fold methods.

The second one (26c72d44ca62104ff79a9e966543597a69944804) makes safer access to params in ast visitor.

### DIFF
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -312,19 +312,7 @@ impl EditorState {
                 Some(expressions) => expressions,
                 None => return vec![],
             },
-            None => {
-                let analysis = self
-                    .contracts_lookup
-                    .get(contract_location)
-                    .and_then(|c| self.protocols.get(&c.manifest_location))
-                    .and_then(|p| p.contracts.get(contract_location))
-                    .and_then(|c| c.analysis.as_ref());
-
-                match analysis {
-                    Some(analysis) => &analysis.expressions,
-                    None => return vec![],
-                }
-            }
+            None => return vec![],
         };
 
         let ast_symbols = ASTSymbols::new();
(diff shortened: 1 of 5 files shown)


components/clarity-lsp/src/common/state.rs
```diff
@@ -312,19 +312,7 @@ impl EditorState {
...
-            None => {
-                let analysis = self
-                    .contracts_lookup
-                    .get(contract_location)
-                    .and_then(|c| self.protocols.get(&c.manifest_location))
-                    .and_then(|p| p.contracts.get(contract_location))
-                    .and_then(|c| c.analysis.as_ref());
-
-                match analysis {
-                    Some(analysis) => &analysis.expressions,
```
hugocaillard, line 324: This fallback could lead to having obsolete data, better not to have it

components/clarity-repl/src/analysis/ast_visitor.rs
```diff
@@ -144,14 +144,17 @@ pub trait ASTVisitor<'a> {
...
-                        DefineFunctions::Trait => self.traverse_define_trait(
-                            expr,
-                            args.get(0)
-                                .unwrap_or(&DEFAULT_EXPR)
-                                .match_atom()
-                                .unwrap_or(&DEFAULT_NAME),
-                            &args[1..],
-                        ),
+                        DefineFunctions::Trait => {
+                            let params = if args.len() >= 1 { &args[1..] } else { &[] };
```
hugocaillard, line 148: I'm surprised that rust lets us write unsafe code like that (`args[1..]` even if args length can be 0). 
I used this `if else` pattern in multiple places, let me know if there's a better way to do so

lgalabru: That works!

lgalabru (approved):
LGTM, thanks @hugocaillard!

PR #731, opened 2022-12-19
fix: safe access to define function params

hugocaillard:
Follow up for #728.
Fix more unsafe access to Slice in AST Visitor.
Plus: allow to disable/enable some LSP features

Screenshot 2022-12-19 at 16 17 44

### DIFF
--- a/components/clarity-repl/src/analysis/ast_visitor.rs
+++ b/components/clarity-repl/src/analysis/ast_visitor.rs
@@ -56,9 +56,12 @@ pub trait ASTVisitor<'a> {
                         DefineFunctions::PrivateFunction => {
                             match args.get(0).unwrap_or(&DEFAULT_EXPR).match_list() {
                                 Some(signature) => {
-                                    let name = signature[0].match_atom().unwrap_or(&DEFAULT_NAME);
+                                    let name = signature
+                                        .get(0)
+                                        .and_then(|n| n.match_atom())
+                                        .unwrap_or(&DEFAULT_NAME);
                                     let params = match signature.len() {
-                                        1 => None,
+                                        0 | 1 => None,
                                         _ => match_pairs_list(&signature[1..]),
                                     };
                                     self.traverse_define_private(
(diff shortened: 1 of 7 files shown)


components/clarity-vscode/package.json
```diff
@@ -77,6 +77,26 @@
...
+            "description": "Allow auto completion for native and custom functions"
+          },
+          "clarity-lsp.hover": {
+            "type": "boolean",
+            "default": true,
+            "description": "Show documentation for native keywords and method on Hover"
+          },
+          "clarity-lsp.documentSymbols": {
+            "type": "boolean",
+            "default": false,
```
hugocaillard, line 93: False by default. This feature was too unstable lately compared to the benefits it brings. (Even though issues should be fixed now, it seems safer for now)

components/clarity-repl/src/analysis/ast_visitor.rs
```diff
@@ -53,59 +53,33 @@ pub trait ASTVisitor<'a> {
...
                             match args.get(0).unwrap_or(&DEFAULT_EXPR).match_list() {
                                 Some(signature) => {
-                                    let name = signature[0].match_atom().unwrap_or(&DEFAULT_NAME);
+                                    let name = signature
+                                        .get(0)
+                                        .and_then(|n| n.match_atom())
+                                        .unwrap_or(&DEFAULT_NAME);
                                     let params = match signature.len() {
-                                        1 => None,
+                                        0 | 1 => None,
```
hugocaillard, line 66: This is the actual fix

lgalabru (approved):
LGTM, thanks @hugocaillard!

components/clarity-repl/src/analysis/ast_visitor.rs
```diff
@@ -53,59 +53,33 @@ pub trait ASTVisitor<'a> {
...
-                                        name,
-                                        params,
-                                        args.get(1).unwrap_or(&DEFAULT_EXPR),
-                                    );
-                                }
-                                _ => {
-                                    false;
-                                }
-                            }
-                            true
```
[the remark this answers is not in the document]
obycode: That looks like a mistake. I can't think of why that would've been intentional. Thanks!

components/clarity-vscode/package.json
lgalabru: is it possible / would it make sense to enable this by default only on pre-releases.
it'd be great to have some eyes on experimental features if developers are willing to use pre-release software.

obycode (approved):
LGTM. Thanks for this fix!

components/clarity-repl/src/analysis/ast_visitor.rs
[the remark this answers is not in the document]
hugocaillard: Thanks for confirming @obycode

PR #737, opened 2022-12-20
Add keywords command

### LINKED ISSUE #477: .functions returns also keywords
In addition to `.functions`, a meta command `.keywords` would help beginners.

`.keywords` would display
```
true
false
block-height
burn-block-height
none
tx-sender
```

Alternatively, the keywords are displayed different in the result list of `.functions`

csgui:
Fix #477

Added support, in the clarity-repl, for a new command called `keywords`. That command returns the keywords from the Clarity language.

Below some screenshots with the new command and it's usage.

image

Thanks.

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -126,6 +126,7 @@ pub struct Session {
     pub show_costs: bool,
     pub executed: Vec<String>,
     pub current_epoch: StacksEpochId,
+    keywords_reference: HashMap<String, String>
 }
 
 impl Session {
@@ -154,6 +155,7 @@ impl Session {
             settings,
             executed: Vec::new(),
             current_epoch: StacksEpochId::Epoch2_05,
+            keywords_reference: clarity_keywords()
         }
     }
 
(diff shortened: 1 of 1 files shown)


lgalabru:
Thanks @csgui, we'll review this PR by end of day!

components/clarity-repl/src/repl/session.rs
```diff
@@ -697,8 +700,16 @@ impl Session {
...
+    pub fn lookup_functions_or_keywords_docs(&self, exp: &str) -> Option<&String> {
+        if let Some(function_doc) = self.api_reference.get(exp) {
+            return Some(function_doc);
+        } else {
+            if let Some(keyword_doc) = self.keywords_reference.get(exp) {
+                return Some(keyword_doc);
+            } else {
+                return None;
+            }
+        }
```
hugocaillard, line 712:
```suggestion
        if let Some(function_doc) = self.api_reference.get(exp) {
            return Some(function_doc);
        }
        if let Some(keyword_doc) = self.keywords_reference.get(exp) {
            return Some(keyword_doc);
        }
        None
```

Because of the early returns, the `else` are not needed here. Helps to avoid some nesting

hugocaillard (changes requested):

components/clarity-repl/src/repl/session.rs
csgui: @hugocaillard requested changes done. Thanks!

hugocaillard (approved):
LGTM!
Thanks for the contribution @csgui

PR #739, opened 2022-12-21
fix: change asset title to uSTX

csgui:
Fix #531

Assets values are shown in uSTX unit.

PR #740, opened 2022-12-21
fix: handle STX assets title

csgui:
Pull Request to fix an issue with the PR #739 .

It wasn't being considered the case where the project has more than one asset to be displayed.

**Before the fix on this PR:**
image

**After the fix on this PR:**
image

image

If any other place needs attention, please let me know. Thanks!

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -965,7 +965,7 @@ impl Session {
             let tokens = self.interpreter.get_tokens();
             let mut headers = vec!["Address".to_string()];
             for token in tokens.iter() {
-                if String::from("STX").eq(&String::from(token)) {
+                if token == "STX" {
                     headers.push(String::from("uSTX"));
                 } else {
                     headers.push(String::from(token));
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -963,7 +963,15 @@ impl Session {
         let accounts = self.interpreter.get_accounts();
         if accounts.len() > 0 {
             let tokens = self.interpreter.get_tokens();
-            let headers = vec!["Address".to_string(), "uSTX".to_string()];
+            let mut headers = vec!["Address".to_string()];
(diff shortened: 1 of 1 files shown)


hugocaillard (changes requested):
My bad for missing this as well. Thanks for spotting it!

Could you make it so "uStx" is still the first token to be displayed (to be in the first column) as it used to be?
Thanks!

csgui:
Hey @hugocaillard ,

> Could you make it so "uStx" is still the first token to be displayed (to be in the first column) as it used to be?

Yes, I can reorder the columns but it wasn't like that before. STX column was always the last one as far as I remember.

Should I work on the column reorder anyway?

Thanks!

components/clarity-repl/src/repl/session.rs
```diff
@@ -963,7 +963,15 @@ impl Session {
         let accounts = self.interpreter.get_accounts();
         if accounts.len() > 0 {
             let tokens = self.interpreter.get_tokens();
-            let headers = vec!["Address".to_string(), "uSTX".to_string()];
+            let mut headers = vec!["Address".to_string()];
+            for token in tokens.iter() {
+                if String::from("STX").eq(&String::from(token)) {
```
hugocaillard, line 968: You should be able to do `if token == "STX"` here, or is their any reason why not?

hugocaillard:
Oh right no need to re-order then. Just left a comment but otherwise it seems fine like that 👍

components/clarity-repl/src/repl/session.rs
csgui: The operator `==` relies on the `eq` function to do the comparison. Thus, this add one more underlying function call to the stack. But I think there is no need to worry with that... it's only a "muscular" memory.

And I should agree that the infix notation (1 == 1) is more readable. :-)... changed.

hugocaillard: Oh okay, didn't know about that, thanks for the explanation

hugocaillard (approved):
LGTM!
Thanks @csgui

PR #741, opened 2022-12-21
fix: docker socket handling

lgalabru:
Improving the best effort to use docker:
- use the socket path specified in the config or use a default path
- use docker defaults settings
- try to use the socket eventually living in `/Users/<name>/.docker/run/docker.sock`


hugocaillard (approved):
LGTM

PR #744, opened 2022-12-21
Feat: better completion

### LINKED ISSUE #670: LSP: Improve completion
- [x] Rework the completions to directly use lsp_types instead of going through intermediary custom types
- [x] Move the completions related code into its own file under lsp/src/common/request/
- [x] Improve parenthesis handling
- [x] Propose keywords based on the clarity version of the contract
- [x] Suggest clarity 2 aliases instead of v1 version (`element-at?`, `index-of?`) 
- [x] Improve snippets handling, developer currently have to delete the snippets keyword before actually typing code
- [x] Autocomplete constants
- [x] Auto-complete variables, maps, fts, nfts (with completion choices)
- [x] Auto complete let binding and function parameters

hugocaillard:
Fix #670


components/clarity-lsp/src/common/backend.rs
```diff
@@ -279,14 +278,19 @@ pub enum LspRequestResponse {
...
+    editor_state: &mut EditorStateInput,
+) -> LspRequestResponse {
     match command {
         LspRequest::Initialize(params) => {
             let initialization_options: InitializationOptions = params
                 .initialization_options
                 .and_then(|o| serde_json::from_str(o.as_str()?).ok())
                 .expect("failed to parse initialization options");
 
+            let _ = editor_state.try_write(|es| es.settings = initialization_options.clone());
```
hugocaillard, line 292: I should probably handle that error

components/clarity-lsp/src/common/backend.rs
```diff
@@ -279,14 +278,19 @@ pub enum LspRequestResponse {
     Hover(Option<Hover>),
 }
 
-pub fn process_request(command: LspRequest, editor_state: &EditorStateInput) -> LspRequestResponse {
+pub fn process_request(
+    command: LspRequest,
+    editor_state: &mut EditorStateInput,
```
hugocaillard, line 283: Question:
Until know, I tried to avoid mutating the state in LSP requests.
State mutation only happened in LSP notification.
So that requests don't lock the state.
Now I only need to edit the state in the Initialization request, but I considered implementing this logic in the native and vscode bridges instead of passing it as a &mut here (which opens the door to more mutation in the future). Any opinion about that?

components/clarity-vscode/package.json
```diff
@@ -83,6 +83,18 @@
             "order": 0,
             "description": "Allow auto completion for native and user-defined functions"
           },
+          "clarity-lsp.completionSmartParenthesisWrap": {
+            "type": "boolean",
+            "default": true,
+            "order": 0,
+            "description": "If set to true, the completion won't wrap a function in parenthesis if an opening parenthesis is already there"
```
sabbyanandan, line 90:
```suggestion
            "description": "If set to true, the auto-completion won't wrap a function with a parenthesis if an opening parenthesis is already there."
```

components/clarity-vscode/package.json
```diff
@@ -83,6 +83,18 @@
             "order": 0,
             "description": "Allow auto completion for native and user-defined functions"
```
sabbyanandan, line 84:
```suggestion
            "description": "Allow auto-completion for native and user-defined functions."
```

components/clarity-vscode/package.json
```diff
@@ -83,6 +83,18 @@
...
+            "type": "boolean",
+            "default": true,
+            "order": 0,
+            "description": "If set to true, the completion won't wrap a function in parenthesis if an opening parenthesis is already there"
+          },
+          "clarity-lsp.completionIncludeNativePlaceholders": {
+            "type": "boolean",
+            "default": true,
+            "order": 0,
+            "description": "If set to true, the completion of native functions will include arguments placeholder. If set to false, it will only do so for functions interacting with variables, maps, fts and nfts."
```
sabbyanandan, line 96:
```suggestion
            "description": "If set to true, the auto-completion of native functions will include arguments placeholder. If set to false, it will only do so for functions interacting with variables, maps, FTs, and NFTs."
```

sabbyanandan (changes requested):
A few subtle nit-picks. Also, for consistency, if we can add a `.` at the end for each description, that'd be great.

hugocaillard:
Done @sabbyanandan 
Thanks!
I re-requested your review so that it's not marked as "changes requested" anymore. I don't know if you can simply discard you review

sabbyanandan (approved):
Looks great! Thank you.

components/clarity-lsp/src/common/backend.rs
lgalabru: As discussed off band, introducing a `LspMutatingRequest` would be a good way out, introducing a path for both mutating an non mutating requests.

components/clarity-lsp/src/common/requests/completion.rs
```diff
@@ -0,0 +1,727 @@
...
+        let snippet = data.populate_snippet_with_options(
+            &"nft-mint?".to_string(),
+            &"nft-mint? ${1:asset-name} ${2:asset-identifier} ${3:recipient}".to_string(),
+        );
+        assert_eq!(
+            snippet,
+            Some("nft-mint? ${1|bitcoin-nft|} ${2:asset-identifier} ${3:recipient}".to_string())
+        );
+    }
+}
```
lgalabru, line 727: Thank you for adding this tests!

components/clarity-lsp/src/common/requests/definitions.rs
```diff
@@ -470,10 +473,16 @@ impl<'a> ASTVisitor<'a> for Definitions {
...
             for binding in binding_exprs {
-                let (name, _) = binding.match_list()?.split_first()?;
-                result.insert(name.match_atom()?.to_owned(), span_to_range(&binding.span));
+                if let Some(name) = binding
+                    .match_list()
+                    .and_then(|l| l.split_first())
+                    .and_then(|(name, _)| name.match_atom())
+                {
+                    result.insert(name.to_owned(), span_to_range(&binding.span));
+                }
```
lgalabru, line 485: This new approach is nicer to read, but not equivalent to the previous implementation. In the initial approach, the `?` would break the loop on the 1st occurence, with your refactoring, we would keep looping.
Is that what we want?

hugocaillard: Yes that intended. There's the same logic in a couple of other places in this PR.
I might have been over using `?` in methods returns `Option<>`, _especially_ in loops.
With the `.and_then`s, the code is a bit more resilient to invalid ASTs (which happens a lot while the contract is being edited). So if an expression is invalid, it's just ignore and we continue with the loops 👌

components/clarity-jupyter-kernel/Cargo.toml
```diff
@@ -8,21 +8,25 @@ readme = "README.md"
...
 clarity_repl = { package = "clarity-repl", path = "../clarity-repl" }
 lazy_static = "1.4.0"
-regex = "1.3.4"
+regex = "1.7"
 sha2 = "0.8.1"
 sha3 = "0.8.2"
 ripemd160 = "0.8.0"
-json = { version = "0.11.15" } # should use serde_json instead
-failure = { version = "0.1.5", default-features = false, features = [ "std" ] }
+json = { version = "0.11.15" }                                                # should use serde_json instead
```
hugocaillard, line 24: auto format from the toml extension i have.. do we like it 🤔

lgalabru: No strong opinion on my end.

lgalabru (approved):
Looking good, thank you for all these enhancements @hugocaillard!

PR #748, opened 2023-01-02
Fix: update pox-2

lgalabru:
The current pox-2 contract embedded in Clarinet is not the latest version (thanks @philipdesmedt).
Tested against https://github.com/arkadiko-dao/arkadiko/pull/532.


obycode (approved):
LGTM!

PR #754, opened 2023-01-06
fix: missing commands help information

### LINKED ISSUE #738: Some commands missing in the help output
While working with the `clarity-repl` I figured out that some commands are missing in the help information.

`get_epoch`
`encode`
`decode`

Is that a known issue or is it mean to be like that?

image

Cheers!

csgui:
Fix #738 .

And, when reviewing, also please check if the command wording/spelling is ok.

Cheers!

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -846,7 +846,7 @@ impl Session {
         ));
         output.push(format!(
             "{}",
-            help_colour.paint("::decode <expr>\t\t\t\tDecode a Clarity Value")
+            help_colour.paint("::decode <bytes>\t\t\tDecode a Clarity Value bytes representation")
         ));
     }
 
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -817,6 +817,10 @@ impl Session {
         ));
         output.push(format!(
             "{}",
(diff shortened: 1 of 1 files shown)


components/clarity-repl/src/repl/session.rs
```diff
@@ -836,6 +840,14 @@ impl Session {
             "{}",
             help_colour.paint("::read <filename>\t\t\tRead expressions from a file")
         ));
+        output.push(format!(
+            "{}",
+            help_colour.paint("::encode <expr>\t\t\t\tEncode an expression to a Clarity Value bytes representation")
+        ));
+        output.push(format!(
+            "{}",
+            help_colour.paint("::decode <expr>\t\t\t\tDecode a Clarity Value")
```
hugocaillard, line 849:
```suggestion
            help_colour.paint("::decode <bytes>\t\t\t\tDecode a Clarity Value bytes representation")
```

Could we make it easier to understand what is expected here?

hugocaillard (commented):
Thanks for the addition!
Small suggestion, let me know what you think

components/clarity-repl/src/repl/session.rs
csgui: Sure... better description indeed. Fixed.

hugocaillard (approved):
LGTM!
Thanks a lot for the addition!

csgui:
Failing check will be fixed by PR #757 .

PR #755, opened 2023-01-09
fix: bump stacks-node next image

hstove:
On the latest Clarinet (1.3.1), `clarinet integrate` was hanging after Bitcoin block 120. Trying to reproduce it in a clean project ran into other odd issues (like blocks not being mined after the first 2, or sometimes very slowly). I didn't see anything obvious in the logs. I checked to see if there was a new tag on quay for `stacks-node` devnet, and saw there was one. After setting that as my `stacks_node_image_url`, it's working well!


lgalabru (approved):
Thanks @hstove!
the current beta4 is already obsolete, I'll will overwrite it with a more recent version.

PR #762, opened 2023-01-10
Feat: warning when opened contract is not in manifest

hugocaillard:
I'll probably want to have a more scalable warning/error handling in the future but that's a start

Screenshot 2023-01-10 at 18 06 41


lgalabru (approved):
Great addition, thank you @hugocaillard!

PR #764, opened 2023-01-10
fix: remove warning in `expectPrintEvent`

hstove:
If you call `expectPrintEvent`, and there are other types of events before the print, then an error will be logged. This warning clutters the logs and made me think the test was failing.

### DIFF
--- a/components/clarinet-deno/index.ts
+++ b/components/clarinet-deno/index.ts
@@ -209,7 +209,7 @@ export class Chain {
       // @ts-ignore
       Deno.core.opSync("api/v1/switch_epoch", {
         sessionId: this.sessionId,
-        epoch: epoch
+        epoch: epoch,
       })
     );
     return result;
@@ -801,6 +801,7 @@ Array.prototype.expectPrintEvent = function (contractIdentifier, value) {
   for (const event of this) {
     try {
       const { contract_event } = event;
+      if (!contract_event) continue;
       if (!contract_event.topic.endsWith("print")) continue;
       if (!contract_event.value.endsWith(value)) continue;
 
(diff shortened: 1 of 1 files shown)


components/clarinet-deno/index.ts
```diff
@@ -812,8 +813,7 @@ Array.prototype.expectPrintEvent = function (contractIdentifier, value) {
         topic: contract_event.topic,
         value: contract_event.value,
       };
-    } catch (error) {
-      console.warn(error);
```
hugocaillard, line 816: Why removing the log here?

hugocaillard (commented):
Thanks a lot!
Definitely makes more sense like that.
Just one question

hugocaillard:
Hey @hstove 
Have you seen the review?
Without an answer I'm tempted to put the log back and merge the PR

components/clarinet-deno/index.ts
hstove: If you look at the other `expect` functions, they don't have a log. That's because they're built with the assumption that an error _will_ be thrown if certain properties on the event are undefined.

For example, with a fungible token event, `event.contract_event` is undefined, so calling `event.contract_event.topic` will throw an error. If you have a tx with a FT event and then a log, the warning will be printed, even though it shouldn't be.

Because I added the line at 804, then maybe we don't need the `try/catch` block at all - there should be no other errors. But overall I was just trying to follow the structure of the other functions

hstove:
Hey @hugocaillard, sorry for missing your first comment! Responded inline

components/clarinet-deno/index.ts
hugocaillard: Alright, makes sense indeed!
Yes I'd like to refacto all these `expect<Event>` method, I think it could be simplified and display better messages in case of failure. Will work on it eventually

hugocaillard (approved):

PR #769, opened 2023-01-13
VSCode extension v1.4.0

hugocaillard:


lgalabru (approved):
Looking great! Could you please update https://github.com/hirosystems/clarinet/issues/557 with the instructions required to run these tests in Github Action?

PR #775, opened 2023-01-18
fix: update epoch handling 

lgalabru:
It looks like the latest clarity vm fixed an issue related to the epoch handling that requires some updates in clarinet.
Adding a clarity 2 contract in the counter example should help us catching eventual future regressions.


hugocaillard (approved):
LGTM

hugocaillard (approved):
LGTM

PR #778, opened 2023-01-19
fix: error logs in chainhook-node

lgalabru:


hugocaillard (approved):
LGTM
Although, a description in the PR could help to review small changes like that 👍

lgalabru:
Sure, nothing too crazy here. We're removing a useless log, and reversing the order of the micro blocks before submitting them to the function in charge of detecting re-org, to avoid an error log to be displayed.

PR #779, opened 2023-01-19
fix: vscode: avoid single quote auto closing

### LINKED ISSUE #774: Make VSCode not automatically add matching single quote
In Clarity, the `'` only shows up at the beginning of a principal literal, never as a pair of quotes. Is there something that we can change in the grammar that would make it not automatically add a second `'` when you type a `'`? This is a minor annoyance, but it's a pain whenever typing in a principal literal.

hugocaillard:
Fix #774


obycode (approved):
Awesome! I was hoping this was an easy change. Thanks @hugocaillard!

Issue #462, opened 2022-02-18
CI job should still run if only Cargo.toml changes


obycode:

csgui:
Hello @obycode !

I would like to contribute with this issue. Could you please provide more info on that? When changes to `Cargo.toml` aren't triggering the CI? Is on a pull request creation? Or on a push to branches?

Thanks!

obycode:
Thanks @csgui. There have been some changes to the CI configuration since this issue was posted, so it may not be relevant anymore. Perhaps we should first start by opening a PR where only the Cargo.toml is changed and then check if the CI jobs run. When I created this issue, there had been a PR where they didn't run in that scenario, but I have a feeling it has since been fixed through unrelated changes.

csgui:
> Perhaps we should first start by opening a PR where only the Cargo.toml is changed and then check if the CI jobs run.

Hi @obycode,

I was checking the [Cargo.toml](https://github.com/hirosystems/clarinet/blob/develop/Cargo.toml) and notice that the `clarity-repl` component is missing from the members list.

We can add that to the list and check if the CI will run. What you think?

obycode:
Yes, that sounds good to me. @lgalabru - is that left out intentionally for any reason?

PR #782, opened 2023-01-19
fix: add clarity-repl component

csgui:
This PR adds a missing component `clarity-repl` to the main `Cargo.toml` file. The list of components is also re-organized in an alphabetically order.

This PR can be also used to check if the issue #462 is still relevant. If the CI runs automatically, that issue can be closed.


lgalabru (approved):
Thanks @csgui!

PR #793, opened 2023-01-23
fix: avoid devnet crash

lgalabru:
This patch is addressing an issue where a chainhook-node receiving a predicate registration before the first ingestion, would cause a crash.


hugocaillard (approved):
LGTM!

PR #797, opened 2023-01-26
fix: avoid expression ID conflict

obycode:
There is a placeholder expression ID inserted for the top-level
instruction which was previously set at 0. It seems that an expression
can have ID 0 during normal evaluation, so instead, use `u64::MAX`.


lgalabru (approved):
LGTM thanks @obycode!

PR #881, opened 2023-02-10
880 close html like tags

LakshmiLavanyaKasturi:
Fixed the html tags which were failing due to incomplete or unescaped html tags


lgalabru (approved):
Thanks @LakshmiLavanyaKasturi!

PR #883, opened 2023-02-13
Removed angular brackets for username

LakshmiLavanyaKasturi:


LakshmiLavanyaKasturi:
Docusaurus does not do anything for html reference tags. Here is the note for it https://docusaurus.io/docs/static-assets. The only way to handle it by using direct markdown tag. Fixed it in this PR

PR #895, opened 2023-02-17
fix: better syntax highlighting for function arguments

hugocaillard:
When defining a clarity function, the syntax highlighting of the function argument wasn't correct.
Here are some before / after screenshots

---
before
Screenshot 2023-02-17 at 18 04 29

---
after
Screenshot 2023-02-17 at 18 03 58


lgalabru (approved):
Thanks @hugocaillard! 
Could we also please update the repo used by linguist?

hugocaillard:
@lgalabru I need to rework this PR, it breaks other stuff. I though it was a quick fix but it's not.

And yes I'll update linguist once done 👍

PR #900, opened 2023-02-22
Fix: lsp: better handling of clarity version

hugocaillard:
The issue was that contracts with no analysis (so empty contracts or contracts with no functions defined) would default to `CLARITY_DEFAULT_VERSION`.
This default value would then be cascaded to the `active_contract_metadata` (for contracts currently opened in the editor) [here](https://github.com/hirosystems/clarinet/blob/develop/components/clarity-lsp/src/common/state.rs#L247-L249). Thus, ignoring the version specified in the manifest


obycode (approved):
LGTM!

Issue #896, opened 2023-02-19
Auto-complete: no space after function with no arguments


obycode:
When autocompleting a call to a function with no args, we get a space after the name, before the closing parenthesis, e.g. `(foo )`.

PR #901, opened 2023-02-22
fix: extra whitespace in function completion

hugocaillard:
Fix: #896 

Thanks for raising this issue @obycode


obycode (approved):
Thanks!

PR #912, opened 2023-02-27
test: improvements to the generated unit tests

obycode:
These ideas come from @whoabuddy and the CityCoins project. The TODO
assertion adds a nice visual reminder to add new tests if the user runs
`clarinet test` with the default tests. The "arrange, act, assert" is a
nice pattern for organizing these tests.


lgalabru (approved):
LGTM!

PR #915, opened 2023-03-01
fix: resolve minor issues with subnet API in devnet

obycode:


lgalabru (approved):
Thanks @obycode! Would it make sense to open an issue so we track the fact that we took a shortcut?
I will merge and address the audit issue in a subsequent PR.

PR #933, opened 2023-03-14
fix: resolve panic when generating completions

obycode:
```
$ clarinet completions bash
thread 'main' panicked at 'crate::generate should have set the bin_name', /Users/brice/.cargo/registry/src/github.com-1ecc6299db9ec823/clap_complete-3.2.5/src/shells/bash.rs:17:24
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```


obycode:
I'll also add this to the release playbook.

lgalabru (approved):
Thanks @obycode!

PR #937, opened 2023-03-15
fix: bug in dependency ordering across epochs

obycode:


lgalabru (approved):
LGTM, thanks @obycode!

PR #941, opened 2023-03-17
refactor: use constants for default settings

obycode:
There are currently two places where the default settings get used:
1. When generating the settings files for a new project
2. When creating a `NetworkManifest` object
These two should stay in sync. This commit makes the file generation use
the same constants used to initialize the object.

Note, not all values have constants defined, so this could be improved
on later.


obycode:
This change just updates the version number for the subnet-node image and does some refactoring to help ensure the default values and the generated settings file stay in sync. It might be nice to quickly release a 1.5.3 with this version number changed (I should've thought of this before releasing 1.5.2).

obycode:
Also, updated the default subnet contract.

PR #959, opened 2023-04-06
feat: allow specifying filename with test coverage

### LINKED ISSUE #329: Allow specifying a file name for `--coverage`
To work around the OOM issue in #324 we split our tests so that they [run separately in the CI](https://github.com/citycoins/contracts/blob/develop/.github/workflows/test-contract.yaml).

Each one of the tests is run with `--coverage` and creates a generic `coverage.lcov` file. Right now we've added an extra step in the CI to rename this file after each test, but it would be much cleaner if we could just append an optional file name at the end.

`clarinet test --coverage mia-coverage.lcov tests/cities/mia/`

MicaiahReid:
Fixes #329

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -25,6 +25,7 @@ use clarinet_files::{
     get_manifest_location, FileLocation, ProjectManifest, ProjectManifestFile, RequirementConfig,
 };
 use clarity_repl::analysis::call_checker::ContractAnalysis;
+use clarity_repl::analysis::coverage::parse_coverage_str;
 use clarity_repl::clarity::vm::analysis::AnalysisDatabase;
 use clarity_repl::clarity::vm::costs::LimitedCostTracker;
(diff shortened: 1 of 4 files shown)


components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -395,9 +398,13 @@ struct Integrate {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 struct Test {
-    /// Generate coverage file (coverage.lcov)
-    #[clap(long = "coverage")]
-    pub coverage: bool,
+    /// Generate coverage file, and optionally provide name of generated file (defaults to "coverage.lcov")
+    #[clap(
+        long = "coverage",
+        default_missing_value("coverage.lcov"),
```
MicaiahReid, line 404: The `clap::default_missing_value` function allows the legacy functionality to still work. A user can use just the `--coverage` flag, and it will default to `"coverage.lcov"`.

lgalabru (approved):
🥇 Looking great, thanks @MicaiahReid!

PR #963, opened 2023-04-14
feat: use realistic block times

### LINKED ISSUE #957: Realistic block times
In simnet, we simulate realistic block times between blocks, but they start at 0. It might be a good idea to start off the first block at the current time instead of 0 to provide more realistic results.

MicaiahReid:
This PR fixes #957 by setting the genesis block's `burn_block_time` field to the system's current time, in seconds.

It also updates the calculation on any given block to be in reference to the genesis block's time.


hugocaillard (approved):
LGTM

PR #969, opened 2023-04-20
chore: update subnet images and contract

obycode:
Use the latest subnet image (`0.5.0`), subnet API image (`7.1.0-subnets.1`), and subnet contract `subnet-v1-2`.

### DIFF
--- a/components/clarinet-files/src/lib.rs
+++ b/components/clarinet-files/src/lib.rs
@@ -35,8 +35,6 @@ use std::pin::Pin;
 use std::{borrow::BorrowMut, path::PathBuf, str::FromStr};
 use url::Url;
 
-pub const DEFAULT_DEVNET_BALANCE: u64 = 100_000_000_000_000;
-
 pub type FileAccessorResult<T> = Pin<Box<dyn Future<Output = Result<T, String>>>>;
 
 pub trait FileAccessor {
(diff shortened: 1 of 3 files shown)


obycode:
It looks like the 0.5.0 subnet image failed to push to docker hub, so don't merge this yet.

jbencin:
Pushed a commit removing `[[ustx_balance]]` from subnets config files. Support for this initial account balances was removed in subnets 0.5.0 and now makes the node `panic!()` on boot

This is ready to merge now

hugocaillard:
@jbencin It's super handy in clarinet to provide a few accounts with STX funds, we lost this with this PR. Would it be do it in an other way?

jbencin:
You can still start the Stacks L1 with a balance (see [here](https://github.com/hirosystems/clarinet/blob/7f0e2355a257bd90d904b76e5917a4cb9c2df8ad/components/stacks-network/src/orchestrator.rs#L946)). But this option was removed from subnets in this PR: hirosystems/stacks-subnets#249, and now the presence of `[[ustx_balance]]` in a config file will make the node crash

hugocaillard:
@jbencin It was useful to have it on the L2 as well.
For example, without it, I think we need to add a step to the [getting started](https://github.com/hirosystems/stacks-subnets/blob/master/docs/getting-started.md) because we now need to fund the L2 wallet

obycode:
The new behavior is correct, because there can be no STX minted on the subnet. They have to come from the L1. Moving STX from the L1 to the subnet should be part of the enhancements to deployment plans to support subnets.

hugocaillard (approved):
LGTM

lgalabru (approved):
LGTM!

PR #978, opened 2023-04-25
Feat/devnet epoch 2.2

lgalabru:


obycode (dismissed):
👍

obycode (approved):

components/clarinet-deployments/src/requirements.rs
```diff
@@ -122,12 +122,18 @@ pub async fn retrieve_contract(
 
 pub const MAINNET_20_START_HEIGHT: u32 = 1;
 pub const MAINNET_2_05_START_HEIGHT: u32 = 40_607;
-// TODO: This is estimated. Replace with exact height once 2.1 is activated.
-pub const MAINNET_21_START_HEIGHT: u32 = 99_564;
-pub const TESTNET_20_START_HEIGHT: u32 = 1;
+pub const MAINNET_21_START_HEIGHT: u32 = 99_113;
+pub const MAINNET_22_START_HEIGHT: u32 = 103_900;
+pub const MAINNET_23_START_HEIGHT: u32 = 104_359;
+// TODO: This is estimated. Replace with exact height once 2.4 is activated.
```
hugocaillard, line 128:
```suggestion

```

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -105,23 +106,24 @@ impl DevnetEventObserverConfig {
...
+            bitcoind_rpc_password: devnet_config.bitcoin_node_password.clone(),
+            bitcoind_rpc_url: format!("http://{}", services_map_hosts.bitcoin_node_host),
+            bitcoin_block_signaling: BitcoinBlockSignaling::Stacks("http://0.0.0.0:20443".into()),
             stacks_node_rpc_url: format!("http://{}", services_map_hosts.stacks_node_host),
             operators: HashSet::new(),
             display_logs: true,
             cache_path: devnet_config.working_dir.to_string(),
             bitcoin_network: BitcoinNetwork::Regtest,
             stacks_network: chainhook_event_observer::chainhook_types::StacksNetwork::Devnet,
+            ordinals_enabled: false,
```
hugocaillard, line 126: @lgalabru 
Could you please double-check here the default values I added? (including bitcoind_* and bitcoin_block_signaling)
Thx

components/stacks-rpc-client/src/rpc_client.rs
```diff
@@ -67,7 +67,33 @@ pub struct PoxInfo {
...
+            reward_phase_block_length: 2000,
+            reward_slots: 4000,
+            total_liquid_supply_ustx: 1368787887756275,
+            ..Default::default()
+        }
+    }
+
+    pub fn testnet_default() -> PoxInfo {
+        PoxInfo {
+            contract_id: "ST000000000000000000002AMW42H.pox-2".into(),
```
hugocaillard, line 85: Why not pox-3? cc @obycode @lgalabru

hugocaillard (commented):
Left a couple of questions

components/stacks-rpc-client/src/rpc_client.rs
obycode: I don't think these defaults are actually used anywhere. I think they can just be deleted so we don't need to keep them in sync. I'll check and then add a commit if that's right.

hugocaillard: The @obycode

obycode (approved):
LGTM! Thanks @hugocaillard. I just had one minor comment.

obycode:
I just realized, we also need to publish the stacks-node image that works with epoch 2.4 and update the default setting.

components/stacks-rpc-client/src/rpc_client.rs
lgalabru: Maybe we could have @janniks cherrypicking @MicaiahReid's fix, rebase and force push in this branch?

obycode: Oh, I missed your earlier comment. Yeah, feel free to back out or overwrite my commit that deleted this.

Cargo.toml
```diff
@@ -14,3 +14,6 @@ members = [
     "components/stacks-rpc-client",
 ]
 default-members = ["components/clarinet-cli"]
+
+# [patch.crates-io]
+# "clarity-vm" = { path = "../stacks-blockchain/clarity" }
```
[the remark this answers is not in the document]
hugocaillard: Yeah! It took me some time to understand that it only works in the root `cargo.toml`. But this is definitely something we can generalize across the repos and components 👍

obycode (dismissed):

hugocaillard (approved):

hugocaillard:
Let's merge the branch, even though it breaks the CI, it's not directly related to the VM upgrade. A PR with the fix is coming.
I'll squash to keep the history clean
cc @lgalabru

PR #985, opened 2023-05-02
fix: circular dependencies in deployments

### LINKED ISSUE #980: Clarinet check does not finish
When I add this requirement to clarinet toml file, clarinet check starts and never returns
```
[[project.requirements]]
contract_id = "SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.swap-helper-v1-03"
 ```

Code is here: https://github.com/friedger/clarity-stacking-pools/tree/feat/reward-distributions
using clarinet 1.5.4

hugocaillard:
Fix #980 

In case of circular dependency in unknown dependencies, like there is between [.alex-vault](https://explorer.hiro.so/txid/0x68e84359d36530fb71b47573796a4732651dcc5b766d2818e6f6b3e3d593e3e4?chain=mainnet) and [.fixed-weight-pool](https://explorer.hiro.so/txid/0x608734b7a5b09f90780ebd8a141f3254e5c06ea160c27c7f5157d50236a6cf90?chain=mainnet), we run into an infinite loop to handle requirements. 

The issue says that it used to work before, but I wasn't able to find at which point we broke it.

This is a naive implementation to patch it

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -357,9 +357,10 @@ pub async fn generate_default_deployment(
             queue.push_front((contract_id, epoch, clarity_version));
         }
 
+        let mut already_added_contracts: Vec<QualifiedContractIdentifier> = vec![];
+
         while let Some((contract_id, epoch, clarity_version)) = queue.pop_front() {
-            // Extract principal from contract_id
-            if requirements_deps.contains_key(&contract_id) {
+            if already_added_contracts.contains(&contract_id) {
                 continue;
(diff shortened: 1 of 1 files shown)


friedger:
LGTM

lgalabru (approved):
Thanks @hugocaillard!

friedger:
:shipit:

hugocaillard:
We wanted to release today but it introduced some regression (especially impacting subnets and maybe other projects like citycoins). Investigation in progress

PR #992, opened 2023-05-04
fix: default genesis time to 0 in wasm

hugocaillard:
`std::time` is not available in the browser.
It was introduced in https://github.com/hirosystems/clarinet/pull/963
`panicked at 'time not implemented on this platform'`
We missed the error because it happens at runtime (and not at compile time)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -721,9 +721,9 @@ dependencies = [
 
 [[package]]
 name = "chrono"
-version = "0.4.23"
+version = "0.4.24"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "16b0a3d9ed01224b22057780a37bb8c5dbfe1be8ba48678e7bf57ec4b385411f"
+checksum = "4e3c5919066adf22df73762e50cffcde3a758f2a848b113b586d1f86728b673b"
 dependencies = [
  "iana-time-zone",
  "js-sys",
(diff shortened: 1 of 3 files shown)


MicaiahReid:
Great find! And my bad for introducing this bug.

I don't love the idea of having different behavior for wasm and non-wasm use cases. Is that something that we already do elsewhere?

I wonder if there are any wasm-compatible time libraries we could use instead? I'm down to look into this, or just to approve as-is if you think this is a non-issue.

hugocaillard:
@MicaiahReid 
Np, it used to happen a lot. 
Hence why we introduced [this check](https://www.notion.so/hiropbc/Duplicate-this-entry-for-starting-new-release-f3a6a62cde984d9e897125e3ab8d4c0c?pvs=4#fad48af8962640b89a5415e7deb72fa7) in the release process. But this one's happening at runtime. I guess because some wasm platform handle std:time.

We do it quite a few times (you can search `#[cfg(not(feature = "wasm"))]` or `#[cfg(feature = "wasm")]`)
This is something I hope we'll be able to reduce in the future.

But you right that maybe I shouldn't implement a different behaviour between both env. I don't think there a library that we can install and do the work for us. But I could rely on wasm-bindgen to get the current time.

So we would still have 2 different code path for wasm and native but with the same behaviour

hugocaillard:
@MicaiahReid Actually, it's look like `chrono` can handle both. And we are already using it. I'll give it a try!

lgalabru (approved):
Thanks @hugocaillard!

MicaiahReid (approved):
Tested, looks good!

PR #997, opened 2023-05-05
fix: handle clarity-version in the dependency detector

hugocaillard:
In Clarity 1 it was possible to have duplicated functions in trait declarations. It now illegal in Clarity 2.

But when requiring a trait from mainnet that does exploit this bug, clarinet has to handle it.
In the dependency detector, we thought it would be safe to default to Clarity2 for requirements. In this case (requiring a trait with duplicated function), it isn't.

Ideally, we'll want to do the same the EPOCH (avoid default epoch in dependency detector) (to be done in this PR or a future one)

This PR also revers the changes from #985


hugocaillard:
Thanks @lgalabru.

I wasn't able to reproduce the failing behavior in the unit test. It's not trivial since it has to be a dependency of a required contract.
I'll add an examples in `components/clarinet/examples` in the upcoming PR 👌

PR #1021, opened 2023-05-19
ci: force brew bump formula

hugocaillard:
As we are currently investigating CI issues on the homebrew side that we fail to reproduce anywhere else, I'd like to have this `force` flag. Afaik it only forces PRs even if we already have one opened on the homebrew-core repo (which can be detected as duplicated). 

Note: The 2 other lines are automatic formatting


CharlieC3 (approved):
:shipit: 
That CI issue is a real head-scratcher 😕

hugocaillard:
@CharlieC3 Yep..

PR #1026, opened 2023-05-30
feat: ship stacks-network package as a binary

MicaiahReid:
PR #1028 is the base for this PR, which will be the feature branch for this change. The feature branch has already removed most Clarinet packages so that reviewing this PR is a bit easier. However, it's probably worth taking a look at #1028 to make sure I didn't delete anything of importance.

The goal of this PR is to be able to package the crucial components of `stacks-network` to allow them to be used as a docker image. This required bundling `stacks-network` as a binary (rather than just as a library, which it was previously). This necessitated the addition of `stacks-network/src/main.rs`.

This PR also removes the deployment of docker images from the `stacks-network` component.

Currently, this branch depends on a local, patched version of `chainhook-event-observer`.


MicaiahReid:
@lgalabru, I think this is ready for review. Obviously there's a lot more we can do, but I don't want these reviews to get too big, so I can try to make issues for the remaining tasks.

components/stacks-network/src/lib.rs
```diff
@@ -48,15 +48,23 @@ where
...
 }
 
+const BITCOIND_CHAIN_COORDINATOR_SERVICE: &str = "bitcoind-chain-coordinator-service";
+const STACKS_NODE_SERVICE: &str = "stacks-node-service";
+const STACKS_API_SERVICE: &str = "stacks-api-service";
+
+const BITCOIND_RPC_PORT: &str = "18443";
+const STACKS_NODE_RPC_PORT: &str = "20443";
+const STACKS_API_PORT: &str = "3999";
+const STACKS_API_POSTGRES_PORT: &str = "5432";
```
lgalabru, line 58: Should we add a TODO here?

components/stacks-network/src/lib.rs
```diff
@@ -86,15 +94,30 @@ pub async fn do_run_devnet(
...
+            stacks_explorer_host: "localhost".into(), // todo (micaiah)
+            bitcoin_explorer_host: "localhost".into(), // todo (micaiah)
+            subnet_node_host: "localhost".into(), // todo (micaiah)
+            subnet_api_host: "localhost".into(), // todo (micaiah)
+        };
+        devnet.set_services_map_hosts(hosts.clone());
+        hosts
+    };
 
+    println!("bitcoind host {}", ip_address_setup.bitcoin_node_host);
```
lgalabru, line 114: Could we use the logger coming from `ctx`, instead of `printf`?

components/stacks-rpc-client/src/rpc_client.rs
```diff
@@ -67,7 +67,33 @@ pub struct PoxInfo {
...
+            reward_phase_block_length: 2000,
+            reward_slots: 4000,
+            total_liquid_supply_ustx: 1368787887756275,
+            ..Default::default()
+        }
+    }
+
+    pub fn testnet_default() -> PoxInfo {
+        PoxInfo {
+            contract_id: "ST000000000000000000002AMW42H.pox-2".into(),
```
lgalabru, line 85: Could we update to `pox-3`?

lgalabru (approved):
Looking great! Thanks for this 1st PR!

lgalabru (approved):
Looking great! Thanks for this 1st PR!

components/stacks-network/src/lib.rs
MicaiahReid: That print statement shouldn't have been there anyways, thanks!

components/stacks-rpc-client/src/rpc_client.rs
MicaiahReid: Done! I think? 🤔 

I got my info from: http://seed.testnet.hiro.so:20443/v2/pox and http://seed.mainnet.hiro.so:20443/v2/pox.

Can you confirm that what I entered here is correct?

components/stacks-network/src/lib.rs
MicaiahReid: Would you be in favor of removing the `const`s and just having these values directly in the map of hosts? The service names and ports should always stay the same, and the only thing that should change from one devnet to another is the namespace.

I think my preference was to have these as constants so that these numbers actually mean something to me (i.e., `18443` is the bitcoind rpc port), but I don't care too much.

Let me know how you think it's best to proceed!

hugocaillard:
@MicaiahReid Can we merge?

PR #1044, opened 2023-06-13
tests: update example tests

hugocaillard:


MicaiahReid (approved):
All those `let` statements in the original examples make me think these were written by rust developers 🤔 😆

hugocaillard:
@MicaiahReid Yes haha.
The main point was to upgrade the dependencies version but I also fixed the deno warnings.
Ideally we would update the tests templates to generate tests without these warnings, but considering what will happen to the tests set up, I'll save it for later

PR #1045, opened 2023-06-13
Broken CI follow up

lgalabru:


components/stacks-rpc-client/src/rpc_client.rs
```diff
@@ -66,6 +66,53 @@ pub struct PoxInfo {
     pub next_cycle: PoxCycle,
 }
 
+impl PoxInfo {
+    pub fn mainnet_default() -> PoxInfo {
```
hugocaillard, line 70: @lgalabru 
I think Brice removed this default methods in a previous PR.
Do we want to keep it?

lgalabru: yes, this is useful (and used) but was maybe not obvious because of the previous `Default` macro.

hugocaillard (approved):
LGTM 🚀 
Thanks

hugocaillard (approved):

PR #1046, opened 2023-06-13
refactor: better stacks-rpc crate

lgalabru:
In this PR, we are moving some code to the stacks-rpc crate.
Ultimately we get rid of all the on chain logic present in the deployment plan crate (onchain.rs), and have deployment plans relying on this lib instead.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -672,7 +672,7 @@ dependencies = [
  "serde_cbor",
  "serde_derive",
  "serde_json",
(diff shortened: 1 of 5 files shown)


MicaiahReid:
>  get rid of all the on chain logic present in the deployment plan crate (onchain.rs), and have deployment plans relying on this lib instead

Will the changes on the deployment plan side be in a different PR? Or is this one still a work in progress?

hugocaillard (approved):
LGTM

PR #1050, opened 2023-06-15
docs: revert homebrew release warning (#1039)

hugoclrd:
After more than a month of issues to deploy on homebrew, the v1.7.0 has been deployed. Which was almost unexpected.

So this warning is actually not needed.

cc @LakshmiLavanyaKasturi

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -57,7 +57,7 @@ winget install clarinet
 
 For more information on how to install Clarinet on Windows, please see the [Setting Up Your Clarity Environment (Windows)](https://www.youtube.com/watch?v=r5LY1J5oACs) video walkthrough.
 
-### Install from pre-built binary
+### Install from a pre-built binary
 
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -63,7 +57,7 @@ winget install clarinet
 
 For more information on how to install Clarinet on Windows, please see the [Setting Up Your Clarity Environment (Windows)](https://www.youtube.com/watch?v=r5LY1J5oACs) video walkthrough.
 
-### Install from a pre-built binary
+### Install from pre-built binary
```
MicaiahReid, line 60:
```suggestion
### Install from a pre-built binary
```
🤷 I guess while we're changing it, why not leave this?

hugocaillard: Thanks for noticing @MicaiahReid

MicaiahReid (approved):

Issue #1052, opened 2023-06-15
Clarinet unit tests revamp POC


hugocaillard:
### Context

Some initial context in [this discussion](https://github.com/hirosystems/clarinet/discussions/1022)

TL;DR
- Clarinet currently embeds Deno to run `clarinet tests`
- This dependency is heavy, painful to maintain, and comes with build issues
- We decided to explore another approach for Clarinet test

### The POC

Clarinet will expose a JS library (through wasm) that will allow to test smart contracts in any JS/WASM environment (Node, Deno, Web Browers, etc).

- The package could be called `@stacks/clarinet-sdk/unit-tests` (introducing the clarinet SDK package #1034)
- The POC will be mostly tested with [Node 20 test runner](https://nodejs.org/api/test.html)

- here as a working template representing what's feasible at the end of this POC

```js
import { main } from "../../../../hiro/clarinet/components/clarinet-sdk/dist/index.js";
import { before, describe, it } from "node:test";
import assert from "node:assert/strict";
import { Cl } from "@stacks/transactions";

let sender = "ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5";

describe("test counter", () => {
  let session;

  before(async () => {
    session = await main();
    await session.initSession(process.cwd(), "./Clarinet.toml");
  });

  it("gets counter value", () => {
    const res = session.callReadOnlyFn("counter", "get-counter", [], sender);
    assert.deepEqual(Cl.int(0), res.result);
    assert.equal(session.blockHeight, 1);
  });

  it("increments counter value", () => {
    const incrRes = session.callPublicFn("counter", "increment", [], sender);
    assert.deepEqual(incrRes.result, Cl.ok(Cl.bool(true)));
    assert.equal(session.blockHeight, 2);

    let res = session.callReadOnlyFn("counter", "get-counter", [], sender);
    assert.deepEqual(res.result, Cl.int(1));
    assert.equal(session.blockHeight, 2);
  });

  it("add any value", () => {
    const addRes = session.callPublicFn("counter", "add", [Cl.int(10)], sender);
    assert.deepEqual(addRes.result, Cl.ok(Cl.bool(true)));
    assert.equal(session.blockHeight, 3);

    let res = session.callReadOnlyFn("counter", "get-counter", [], sender);
    assert.deepEqual(res.result, Cl.int(11));
    assert.equal(session.blockHeight, 3);
  });
});
```

### Going further

[...]

PR #1060, opened 2023-06-26
feat: Add `--allow-read` flag to allow filesystem read access in `clarinet test`

jbencin:
Add `--allow-read` to `clarinet test` subcommand in order to allow filesystem reads in tests. This allows us to share configuration between tests by reading data from JSON/YAML files.

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -445,6 +445,9 @@ struct Test {
     /// Allow network access
     #[clap(long = "allow-net")]
     pub allow_net: bool,
+    /// Allow read access to project directory
+    #[clap(long = "allow-read")]
+    pub allow_disk_read: bool,
     /// Specify optional Typescript config file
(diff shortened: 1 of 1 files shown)


MicaiahReid (approved):
Looks good, thanks @jbencin!

hugocaillard:
Thanks Jeff.
I want to point out that the current clarinet test framework (with deno embedded in clarinet) will be removed in Q3 and clarinet will adopt another way of unit testing contracts (#1052). So I don't know if it's worth merging and releasing this addition. How important critical is it for you @jbencin?

jbencin:
> So I don't know if it's worth merging and releasing this

I'm guessing it will be a while until that alternative method is ready, and there will be more Clarinet releases before this happens. This change is only really necessary in CI (because a user can just press 'y' to allow reads), so you don't have to rush a release, I just need an image with it available on DockerHub. Also, this is a pretty trivial change which isn't going to break anything, so I don't see any reason *not* to merge it.

Issue #1043, opened 2023-06-12
Crash in `clarinet check` and LSP


obycode:
I accidentally discovered that this unusual but valid Clarity code causes a crash on the command line or in the LSP:

```clarity
(at-block 0x9f5c0df2f04da020ad0283bbbb365ae177fc32593037f29727eb97a03ba4340e (get-block-info? id-header-hash u108171))
```

```
panicked at 'No height stored for current chain tip', components/clarity-repl/src/repl/datastore.rs:223:9
```

hugocaillard:
What is the expected behavior?

Do we only want to prevent the crash (and return some default value?)

obycode:
Yes, I'd say definitely prevent the crash and then recover by providing as much functionality as possible with a failing build.

PR #1061, opened 2023-06-28
fix: repl: panic at get_current_block_height for unexisting block

hugocaillard:
Fix: #1043 

The fix prevent the crash.
At the clarinet level, it's not really feasible to catch this error and return an error or warning on the diagnostic.


obycode (approved):
Makes sense to me 👍

PR #1064, opened 2023-07-03
feat: allow stacks-network to be run as standalone chain coordinator

MicaiahReid:


components/clarinet-cli/src/chainhooks/types.rs
```diff
@@ -1,433 +0,0 @@
-use serde::{Deserialize, Serialize};
```
MicaiahReid, line 1: This file wasn't used, so I took the opportunity to delete it

components/stacks-network/src/chainhooks.rs
```diff
@@ -1,14 +1,11 @@
+use chainhook_sdk::chainhook_types::{BitcoinNetwork, StacksNetwork};
```
MicaiahReid, line 1: This `clarinet-cli/src/chainhooks/mod.rs` was only used in `stacks-network`, so I moved it

MicaiahReid: When I first made this change, it allowed for some improved code reusability, but I ended up not needing it after all.

So, if anyone doesn't like this being moved I can put it back.

hugocaillard (dismissed):
LGTM
I don't have a lot of experience on stacks-network and stacks-devnet so I think @lgalabru should have a look

components/clarinet-cli/src/chainhooks/types.rs
lgalabru: Good call, thanks!

dockerfiles/components/stacks-network.dockerfile
```diff
@@ -0,0 +1,12 @@
...
+
+COPY . ./
+
+RUN cargo build --release --manifest-path ./components/stacks-network/Cargo.toml
+
+# prod stage
+FROM gcr.io/distroless/cc
+COPY --from=builder target/release/stacks-network /
+
+ENTRYPOINT ["./stacks-network"]
```
lgalabru, line 12: I'm not familiar with distroless, is this a better approach than what we have in the other dockerfiles?

components/stacks-network/src/orchestrator.rs
```diff
@@ -151,7 +160,29 @@ impl DevnetOrchestrator {
...
+            bitcoin_node_host: format!(
+                "bitcoind-chain-coordinator-service.{namespace}.svc.cluster.local:18443"
+            ),
+            stacks_node_host: format!("stacks-node-service.{namespace}.svc.cluster.local:20443"),
+            postgres_host: format!("stacks-api-service.{namespace}.svc.cluster.local:5432"),
+            stacks_api_host: format!("stacks-api-service.{namespace}.svc.cluster.local:3999"),
+            stacks_explorer_host: "localhost".into(), // todo (micaiah)
+            bitcoin_explorer_host: "localhost".into(), // todo (micaiah)
+            subnet_node_host: "localhost".into(),     // todo (micaiah)
+            subnet_api_host: "localhost".into(),      // todo (micaiah)
```
lgalabru, line 177: Could we please create an issue for tracking these todos? Thanks!

lgalabru (commented):
Changes look great, thanks @MicaiahReid! 
Could you also please tell us more about the current status of `clarinet integrate` and `stacks-devnet-js`? Are these 2 features still functional?
Thanks!

dockerfiles/components/stacks-network.dockerfile
MicaiahReid: No idea! I've changed to be the same as what we use for other dockerfiles

components/stacks-network/src/orchestrator.rs
MicaiahReid: Done: https://github.com/hirosystems/clarinet/issues/1113

MicaiahReid:
@lgalabru, I've confirmed that both `clarinet integrate` and `stacks-devnet-js` still work as expected!

CAGS295:
Hi there! I'm interested in using the event channels from devnetjs and plugging in my own devnet node instead of the orchestrator instantiating it. Basically what I'm trying to figure out is if I can instantiate a node with `clarinet integrate` and then use devnetJS to listen and push events to the node, e.g. mining empty blocks. if you could add a description to this PR, it would help me a lot. Judging from the title and file diffs it looks like this PR may be related to what I would like to implement.

MicaiahReid:
Hey @CAGS295! Usually `clarinet integrate` spins up a few docker containers (bitcoin node, stacks node, stacks api, etc), what we call the network orchestrator (which starts up these processes), and a chain coordinator (which acts as a middleman between the docker containers by forwarding requests between them and sending transactions to fund wallets and run you deployment plan).

This PR adds the ability to _only_ start up the chain coordinator, allowing you to provide a configuration of where the chain coordinator can find the stacks/bitcoin/stacks api nodes. You also provide the chain coordinator with a deployment plan, which it will deploy to those nodes.

Currently, however, this is all very specific to those nodes (that we previously ran as docker containers) being run in kubernetes. We have the host of all of the nodes hardcoded as kubernetes host urls.

This could be made to handle more situations, but it would take some work that isn't currently on the docket for us, but PRs would be welcome once this one is merged!

lgalabru (approved):
Looking great, thanks @MicaiahReid!

PR #1068, opened 2023-07-05
feat: initialize clarinet-sdk with the core js lib

hugocaillard:
Resolve: #1110 

```sh
$ cd components/clarinet-sdk/
$ npm ci
$ npm build
```


components/clarinet-sdk/src/unit_tests.rs
```diff
@@ -0,0 +1,279 @@
...
+        session.advance_chain_tip(1);
+
+        Ok(res)
+    }
+
+    // #[wasm_bindgen(js_name=mineBlock)]
+    // pub fn mine_block(&mut self, js_params: JsValue) -> JsValue {
+    //     // @todo
+    //     // re-implement chainhooks logic
+    //     // https://github.com/hirosystems/clarinet/blob/c6d65974f489606c0a8c57a6d0f01cfb993eb79b/components/clarinet-cli/src/runner/api_v1.rs#L718-L811
```
lgalabru, line 253: We can get rid of this feature.

hugocaillard: Do you mean `mine_block` or the chainhook logic @lgalabru?

lgalabru: Chainhook logic.

lgalabru: We'll still need a `mine_block` like approach, but I was wondering if we should keep the same struct, or use the types that we're using in https://github.com/hirosystems/clarinet/blob/develop/components/stacks-devnet-js/lib/index.ts.
Any take?

hugocaillard:
@MicaiahReid @csgui 
I'm marking this PR as ready for review.

The SDK won't be released right after the PR is merged, the idea to keep PRs at reasonable size.

There is still work to do that's not blocking for reviewing/merging
- testing
- bundling

components/clarinet-sdk/src/utils.rs
```diff
@@ -0,0 +1,222 @@
...
+    use super::value_to_string;
+    use clarity_repl::clarity::vm::types::{
+        ASCIIData, CharType, ListData, ListTypeData, OptionalData, ResponseData, SequenceData,
+        SequencedValue, TupleData, TypeSignature, UTF8Data, NONE,
+    };
+    use clarity_repl::clarity::vm::{ClarityName, Value};
+    use std::convert::TryFrom;
+
+    #[test]
+    fn test_value_to_string() {
```
MicaiahReid, line 157: Definitely not crucial, but I've been loving the [test-case](https://github.com/frondeus/test-case) crate for this sort of thing. You can reuse the same test code, and generate multiple test cases that are run against the test. This way, if the first case fails, the rest are still run and it doesn't look like everything is broken. For this test, it could look something like this:

```rust

    #[test_case(Value::Int(42) => is equal_to  "42".to_string(); "for Int type")]
    #[test_case(Value::UInt(12345678909876) => is equal_to  "u12345678909876".to_string(); "for UInt type")]
    #[test_case(Value::Bool(true) => is equal_to  "true".to_string(); "for Bool type")]
    #[test_case(Value::buff_from(vec![1, 2, 3]).unwrap() => is equal_to  "0x010203".to_string(); "for buff type")]
    fn test_value_to_string(val: Value) -> String {
        value_to_string(&val)
    }
```

MicaiahReid (changes requested):
Awesome work! I've got a few small comments.

components/clarinet-sdk/README.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Core testing lib
+
+This a very early preview.
+Expect many breaking changes in the API.
+
+```js
+// @ts-check
+
+import { main } from "obscurity-sdk";
```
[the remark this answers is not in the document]
hugocaillard: I think it won't be published on `@hirosystems` before September so I'd like to keep it like that until then (for the early testers)

components/clarinet-sdk/src-ts/index.ts
```diff
@@ -0,0 +1,110 @@
...
+  result: ClarityValue;
+  events: { event: string; data: { [key: string]: any } }[];
+};
+type GetDataVar = (contract: string, dataVar: string) => ClarityValue;
+type GetMapEntry = (contract: string, mapName: string, mapKey: ClarityValue) => ClarityValue;
+type GetAssetsMap = () => Map<string, Map<string, bigint>>;
+type GetAccounts = () => Map<string, string>;
+
+// because we use a proxy around the test session,
+// the type need to be hardcoded
```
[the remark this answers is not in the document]
hugocaillard: thx

components/clarinet-sdk/webpack.config.js
```diff
@@ -0,0 +1,85 @@
...
+    new WasmPackPlugin({
+      crateDirectory: path.resolve(__dirname, "./"),
+      extraArgs: "--release --target=web",
+      outDir: path.resolve(__dirname, "./src-ts/sdk"),
+    }),
+  ],
+};
+
+module.exports = [configESM];
+// module.exports = [configESM, configCJS];
```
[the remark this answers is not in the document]
hugocaillard: Fixed the CJS build

components/clarinet-sdk/src/utils.rs
```diff
@@ -0,0 +1,222 @@
...
+            event: "ft_burn_event".into(),
+            data: data.json_serialize(),
+        },
+    }
+}
+
+pub fn to_raw_value(value: &Value) -> String {
+    let mut bytes = vec![];
+
+    value.consensus_serialize(&mut bytes).unwrap();
```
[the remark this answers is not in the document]
hugocaillard: A message would be nice yes 👍

components/clarinet-sdk/src/utils.rs
hugocaillard: Indeed this looks great! I'll give it a try in the next phases of this project

MicaiahReid (approved):
🚀

PR #1074, opened 2023-07-12
Draft: Supporting PoX locks in Clarinet

kantai:
One of the issues with testing PoX in Clarinet is that Clarinet doesn't apply account locks when PoX methods are invoked (these require defining a special case handler).

This PR adds this to Clarinet, but it does it in a little bit of clumsy way.

There's kind of two or three questions in my mind about adding this support to Clarinet:

1. Is adding PoX lock handling to Clarinet a useful feature generally or is that something just people working on PoX care about? I could imagine this would make contract writing for things like pools easier. If it's not generally useful, then perhaps allowing a dynamic link here would help.
2. Does it make sense for this to be configurable in some way? Currently, this applies the locks to the boot contracts (which Clarinet does publish). But I guess I could imagine this being a toml configured option to match different Contract IDs. This would allow a "new" PoX contract to be tested in the same way any other clarity contract would.
3. Ideally, this code wouldn't need to be re-implemented between the stacks-blockchain repo and this repo. This code is actually fairly easy to separate from the rest of the stacks codebase (as this PR sort of demonstrates). So maybe this isn't so much a question for the Clarinet folks, because really, this could be another library output from the stacks-blockchain repo. So, I guess my question is, do you all think this is worthwhile enough to engage with refactoring on the stacks-blockchain side?

### DIFF
--- a/components/clarity-repl/src/repl/datastore.rs
+++ b/components/clarity-repl/src/repl/datastore.rs
@@ -1,3 +1,4 @@
+use crate::repl::special;
 use sha2::{Digest, Sha512_256};
 
 use clarity::types::chainstate::BlockHeaderHash;
@@ -285,6 +286,10 @@ impl ClarityBackingStore for Datastore {
     #[cfg(not(feature = "wasm"))]
     fn get_side_store(&mut self) -> &::clarity::rusqlite::Connection {
         panic!("Datastore cannot get_side_store")
+    }
+
+    fn get_cc_special_cases_handler(&self) -> Option<::clarity::vm::database::SpecialCaseHandler> {
+        Some(&special::handle_contract_call_special_cases)
     }
 }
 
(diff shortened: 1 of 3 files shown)


lgalabru:
I'm a supporter of having PoX being supported in clarinet!

kantai:
> I'm a supporter of having PoX being supported in clarinet!

Great. I'll try to update this PR so that it pulls the locking logic directly from a stacks-node library (see https://github.com/stacks-network/stacks-blockchain/pull/3793)

hugocaillard:
Hey @kantai 
Thx for this PR, I rebased it and converted it to draft.
What's the status of this PR? Do you think you'll have time to pursue the effort?

hugocaillard:
Superseded by #1375
Thanks @kantai for starting the work here and showing us the way 🙌

Issue #1079, opened 2023-07-15
Clarinet stop working when adding requirement SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.amm-swap-pool-v1-1


pradel:
When I add `SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.amm-swap-pool-v1-1` as a requirement of my project, clarinet stop working, clarinet check is never responding and vscode stop reporting clarity issues for my contract.

## To reproduce

1. Create a new project `clarinet new test`
2. Create an empty contract `clarinet contract new mycontract`
3. Add the requirement `clarinet requirements add SP3K8BC0PPEVCV7NZ6QSRWPQ2JE9E5B6N3PA0KBR9.amm-swap-pool-v1-1`
4. See clarinet freeze `clarinet check`

hugocaillard:
The for reporting the issue @pradel, the fix will be released this week

hugocaillard:
Thanks @pradel 
clarinet v1.71. and clarity for vscode v1.7.2 have been released with a fiw

pradel:
@hugocaillard thanks!

PR #1080, opened 2023-07-17
fix: clarity version of nested requirements

hugocaillard:
### Description

Fix: #1079 

Follow up of https://github.com/hirosystems/clarinet/pull/997
With the previous fix, the order of the requirements would matter, now we get the right clarity version for each requirement

### Checklist

Note: I didn't add unit tests for project with nested dependencies. But I added the Alex project (which also revealed the bug prior to 997) to the release playbook.

I think this PR makes this part of code slightly simpler but I still think it can be revamped to be simpler and more robust

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -1,7 +1,6 @@
 use clarity_repl::clarity::stacks_common::types::StacksEpochId;
-use clarity_repl::clarity::ClarityVersion;
-use clarity_repl::repl::DEFAULT_EPOCH;
 use clarity_repl::repl::{ClarityCodeSource, ClarityContract, ContractDeployer};
+use clarity_repl::repl::{DEFAULT_CLARITY_VERSION, DEFAULT_EPOCH};
 
 extern crate serde;
 
@@ -322,11 +321,7 @@ pub async fn generate_default_deployment(
                     ))
                 }
             };
-            queue.push_front((
-                contract_id,
-                StacksEpochId::Epoch21,
-                ClarityVersion::Clarity2,
-            ));
+            queue.push_front((contract_id, Some(DEFAULT_CLARITY_VERSION)));
         }
     }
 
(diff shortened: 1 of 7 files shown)


hugocaillard:
@obycode Adding you as an (optional) reviewer since your the original author you might something we could miss

components/clarinet-deployments/src/lib.rs
```diff
@@ -350,14 +345,10 @@ pub async fn generate_default_deployment(
                     ))
                 }
             };
-            // Download the code
-            let (_source, epoch, clarity_version, _contract_location) =
-                requirements::retrieve_contract(&contract_id, &cache_location, &file_accessor)
```
hugocaillard, line 355: the purpose of this PR is to call `retrieve_contract` for each item in the queue and not just here

components/clarinet-deployments/src/lib.rs
```diff
@@ -322,11 +321,7 @@ pub async fn generate_default_deployment(
                     ))
                 }
             };
-            queue.push_front((
-                contract_id,
-                StacksEpochId::Epoch21,
-                ClarityVersion::Clarity2,
-            ));
+            queue.push_front((contract_id, Some(DEFAULT_CLARITY_VERSION)));
```
obycode, line 324: Do we not need the epoch anymore with this change?

hugocaillard: @obycode It was already not need, it was only passed from items to items in the queue, but except for that, the value was never used

obycode: Ah, I see. Thanks!

obycode (dismissed):
LGTM!

lgalabru (approved):
thanks @hugocaillard! this part of the code is a bit sensitive and under documented.
what would be your take on adding a project in the `examples/`  folder, with some contracts outlining the scenario being fixed? The CI is running `clarinet check` on all this projects, and we would have an end to end test helping us with regressions.

hugocaillard:
Done @lgalabru

lgalabru (dismissed):
Thanks @hugocaillard 🙏

PR #1081, opened 2023-07-18
chore: update subnet node to 0.8.1

obycode:
### Description

Also update subnet API to use the latest image, which is now subnet-compatible.

#### Breaking change?

No

### Example

This just changes default values in Devnet.toml.

---

### Checklist


hugocaillard (approved):
LGTM 👍

PR #1082, opened 2023-07-18
chore: update default epoch to 2.4

obycode:
When a new contract is added with `clarinet contracts new`, this will set the epoch to `2.4` instead of the current `2.1`.


hugocaillard (approved):
LGTM
thanks @obycode

PR #1084, opened 2023-07-18
Added epoch config to the clarinet toml file

LakshmiLavanyaKasturi:
- Added clarinet version and epoch configuration to the Clarinet.toml file.


obycode (approved):
👍

Issue #768, opened 2023-01-12
Coverage: extra DA:0 in output file


AveZorgen:
When i use `test --coveraage`  on code `counter.clar`
```Clarity
(define-public (hello-world) 
 (ok (print { msg: "h!", tip: tx-sender })))
```
and test `counter_test.ts`
```Clarity
import { Clarinet, Tx, Chain, Account, types } from 'https://deno.land/x/clarinet@v1.3.1/index.ts';
import { assertEquals } from 'https://deno.land/std@0.170.0/testing/asserts.ts';

Clarinet.test({
    name: "Ensure that no da",
    async fn(chain: Chain, accounts: Map<string, Account>) {
        let sender = accounts.get('deployer')!.address;
        let block = chain.mineBlock([
            Tx.contractCall('counter', 'hello-world', [], sender)
        ]);
    },
});

```
it generates `coverage.lcov` with an extra `DA:0`
```lcov
TN:Ensure that no da:0
SF:/home/ave-zorgen/code/counter/contracts/counter.clar
FN:1,hello-world
FNDA:1,hello-world
FNF:1
FNH:1
DA:0,1
DA:2,4
end_of_record
TN:__analysis__
SF:/home/ave-zorgen/code/counter/contracts/counter.clar
FN:1,hello-world
FNF:1
FNH:0
end_of_record
```
I am new to Clarity and Rust but I guess this is because nested in `print` `SymbolicExpression` with expr `tuple` creates with `Span::zero()` that means start_line equals 0

hugocaillard:
Thanks for reporting @AveZorgen 
I'm bumping the priority to medium. This invalid line number break some coverage renders (such the extensions metioned here https://github.com/hirosystems/clarinet/issues/1062)

Issue #1062, opened 2023-06-30
Support for code coverage visualization in VSCode


lgalabru:
See https://marketplace.visualstudio.com/items?itemName=ryanluker.vscode-coverage-gutters for developer experience.
We could have a mode where the coverage every-time the developer save a file, having coverage in real time would be pretty awesome.

hugocaillard:
I'll test if it already works. See with @MicaiahReid about the lcov format compatibility issue

hugocaillard:
@lgalabru 
It seems to be working fine (I probably need to test on bigger project).
The only requirement would be to rename the lcov file into `lcov.info` 


<img width="678" alt="Screenshot 2023-07-03 at 19 54 17" src="https://github.com/hirosystems/clarinet/assets/911307/fbbc57a7-e32d-41d6-8c97-067f0d0aff85">
fo`

lgalabru:
That's really cool? So is this feature native to VScode, or do we have to install the extra extension? would it make sense to augment our extension to support this out of the box?

lgalabru:
We should probably document this! cc @LakshmiLavanyaKasturi

hugocaillard:
This is with the extension you linked.

> would it make sense to augment our extension to support this out of the box?

Why reinvent the wheel? Or it would be a very low priority task

PR #1085, opened 2023-07-18
fix: safety check on invalid line number in coverage

hugocaillard:
### Description

Fix: #768 

Invalid line number in lcov files (<= 0) can break some coverage renderers (such as the Coverage Gutters extension mentioned here https://github.com/hirosystems/clarinet/issues/1062).

This PR doesn't fix the root cause (we shouldn't have executable line with a span start at line 0) but at least it prevents the renderers to crash. The root cause should be fixed at the VM level (some cases have already been fixed in the past https://github.com/stacks-network/stacks-blockchain/pull/3421) 

#### Breaking change?

/

### Example

/

---

### Checklist

/


lgalabru (approved):
Thanks @hugocaillard!

Issue #1078, opened 2023-07-13
Chore: Revamp CI/CD


hugoclrd:
Our current CI/CD config

- is slow (we probably build useless stuff on each PR, we might also gain a lot of time by removing deno)
- only releases clarinet (the cli), other artifacts are published manually (clarinet, vscode extension, clarinet-sdk, others?)
- is maybe too complicated / lacks best practices and standards.

Some of the tasks are labelled as _short terms_ 🟢, the others are future improvements.
**Help from DevOps teams is wanted to validated the spec and work on the short term issues.**

#### Use semantic release? 

- Does it work with monorepo with multiple target (npm, crates, vscode, homebrew, etc)
- We would like to publish multiple tags depending on which component we want to release (clarinet, clarity-vscode, clarinet-sdk)

### Todo

- New GitHub action from scratch 🟢
- Ideally, we only want to build the components that have been modified

### Split in 3 flows 🟢

- Development flow:
    - Focus: speed
    - Trigger: a new PR is opened against `develop` branch
    - Actions:
        - Worker 1: Static analysis
             - Check code format (P0) 🟢
             - Audit dependencies (P1)
        - Worker 2: Native Linux builds (ideally debian bulleyes)
             - Build clarinet (P0) 🟢
             - Test (P0) 🟢
             - Code coverage (P1)
        - Worker 3: Wasm builds
             - Build clarinet-sdk (P0) 🟢
- Release Playbook flow:
    - Trigger: a new PR is opened against `main` branch
    - Actions:
        - Steps from the **Development flow** 🟢
        - Build and publish a docker image of clarinet, tagged with `develop` 🟢
        - Build and run tests on Windows 🟢
        - Build and run tests on macOs 🟢        
- Release flow:
    - Trigger: a new tag is being pushed (tag per component `clarinet`, `clarity-vscode`, `clarity-sdk`)
    - Actions:
        - Publish on crates.io
        - Publish on docker hub
              - clarinet 🟢 
        - Publish on NPM
              - clarity-sdk
              - stacks-devnet-js
        - Publish on Brew  
              - clarinet 🟢
        - Publish on Winget 
              - clarinet 🟢
        - Publish on VSCode marketplace
              - clarity-vscode 🟢

lgalabru:
> Many of these actions already exist, such as publishing to brew, or code formatting. Can you clarify what you mean by this?

[...]

PR #1104, opened 2023-07-27
ci: revamp

CharlieC3:
Related: https://github.com/hirosystems/clarinet/issues/1078

* Reduces CI build time for clarinet-cli from ~75 minutes to ~10 minutes
* Splits audit and formatting job into separate workflow to reduce duplicate configs later when we have more component-specific workflows
* Improves Rust caching to speed up builds with no depencency changes since last run
* Dists will now be built depending on what's triggering the CI run. `x86_64-unknown-linux-gnu` dist will always get created, others will only run on pushes to `main` and `develop` branches.
  * This is configurable in the [ci_clarinet_cli_dist_matrix.json](https://github.com/hirosystems/clarinet/blob/ci/revamp/.github/workflows/config/ci_clarinet_cli_dist_matrix.json) file
  * This cuts down on CI time for PRs considerably
* Unit tests in the `dist` job will be skipped for the x86_64-unknown-linux-gnu target, since this target should already be covered in the code coverage job.
  * This cuts down on CI time for PRs considerably
* Docker image now reuses the pre-built dist instead of rebuilding from scratch
* Sets up initial scaffolding to build a ARM64 Docker image on pushes to the `main` and `develop` branches
  * This is configurable in the [ci_clarinet_cli_docker_matrix.json](https://github.com/hirosystems/clarinet/blob/ci/revamp/.github/workflows/config/ci_clarinet_cli_docker_matrix.json) file
  * I have this disabled at the moment because the arm64 binary isn't able to execute in an arm64 docker image for some reason, and it will require a larger time investment to resolve. This can probably be put into a separate task, and may require [something like this to work](https://github.com/stacks-network/stacks-blockchain/blob/master/build-scripts/Dockerfile.macos-arm64#L23)

### DIFF
--- a/.github/workflows/audit-format.yaml
+++ b/.github/workflows/audit-format.yaml
@@ -1,0 +1,52 @@
+name: Audit & Format
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+      - develop
+      - rc/next
+      - ci/revamp
+  workflow_dispatch:
+
+# Cancel previous runs for the same workflow
+concurrency:
+  group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
+  cancel-in-progress: true
+
+jobs:
+  audit:
+    name: Audit and format
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Install Rust toolchain
+        run: |
+          rustup set profile minimal
+          rustup toolchain install stable --component rustfmt
+          rustup override set stable
+
+      - name: Cache cargo
+        id: cache-cargo
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cargo/bin/
+            ~/.cargo/registry/index/
+            ~/.cargo/registry/cache/
+            ~/.cargo/git/db/
+            target/
+          key: ${{ runner.os }}-cargo-audit-${{ hashFiles('**/Cargo.lock') }}
+
+      - name: Install dependencies
+        if: steps.cache-cargo.outputs.cache-hit != 'true'
+        run: cargo install cargo-audit
+
+      # This can only be ran in root dir
+      - name: Run audit
+        run: cargo audit --ignore RUSTSEC-2021-0076 --ignore RUSTSEC-2022-0028 --ignore RUSTSEC-2020-0071 --ignore RUSTSEC-2022-0090 --ignore RUSTSEC-2023-0018
+
+      - name: Run rustfmt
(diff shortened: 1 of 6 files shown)


hugocaillard:
This is awesome, thanks @CharlieC3 🙌 
If I understand well, this PR doesn't impact our release workflow right?

CharlieC3:
@hugocaillard Yup, that's the idea! I'm grabbing the latest version of the clarinet-cli component tag [here](https://github.com/hirosystems/clarinet/pull/1104/files#diff-ed5b3be8788f14bb3472c8fc10151fe96e556a8b5f3af2cf1458171da1b1dc7bR66), then comparing it against the version found in components/clarinet-cli/Cargo.toml [here](https://github.com/hirosystems/clarinet/pull/1104/files#diff-ed5b3be8788f14bb3472c8fc10151fe96e556a8b5f3af2cf1458171da1b1dc7bR75-R77). If there's no match, a new tag will be created in the format of `clarinet-cli-v<Cargo.toml version>`.

So just like before, to create a new clarinet-cli version, you have to [bump this line](https://github.com/hirosystems/clarinet/blob/main/components/clarinet-cli/Cargo.toml#L3) and commit/merge it to `main`.

However, we likely won't get good results if we merge this CI workflow in while we keep the original CI workflow active. It's up to you on how you'd like to handle this, but I would suggest disabling the new workflow until you have workflows for all components set up and ready before removing the old CI workflow.

CharlieC3:
@hugocaillard Not sure if you saw my comment above. Any thoughts. How would you like to proceed?

hugocaillard:
I need to check what was the old CI doing that the new doesn't.
And I don't think there's a big difference so maybe we could switch the new one pretty easily

hugocaillard:
@CharlieC3 Ah right.
So you suggest merging this PR but disabling the CI?

CharlieC3:
Since more workflows are needed for a complete picture, yes.

hugocaillard:
@CharlieC3
I manually disabled the "Audit and Format" and "CI Clarinet CLI" actions 👌 

Will the changes on clarinet dockerfile have impacts on the current CI?

CharlieC3:
@hugocaillard Ah good catch, I've updated the old workflow to use the `dev` Dockerfile, since it's pretty much the same as the old Dockerfile.

smcclellan:
@hugocaillard @CharlieC3 What's next on this PR?

hugocaillard (approved):
LGTM 
This new workflow has been disabled for now.
We'll keep using the old one until the workflows have been adapted for other components (mainly stacks-devnet).
The release of clarinet 2 later this year will be a good time to activate it

PR #1133, opened 2023-08-18
feat: clarity to wasm

hugocaillard:
> This PR won't be merged and I'll open a new one when getting back on this topic but I'd like to keep it around until then

We only care about this commit: 588fe0be21a13ae7e28117494250faa1f69bac5f


sabbyanandan:
@hugocaillard: 🤯 👏🏽 
image

hugocaillard:
@sabbyanandan the "-33000" is actually from the commit we don't care too much about in this PR 😅 (at least for now)

But soon!

sabbyanandan:
Much respect to you, sir, @hugocaillard!

components/clarity-repl/src/repl/session.rs
```diff
@@ -657,14 +662,60 @@ impl Session {
         Ok((execution, contract_identifier))
     }
 
+    fn run_wasm_and_clarity(
+        &mut self,
+        snippet: String,
+        eval_hooks: Option<Vec<&mut dyn EvalHook>>,
+        cost_track: bool,
+    ) -> Result<ExecutionResult, Vec<Diagnostic>> {
+        // @todo: we should take another approach than a regex to parse it
```
obycode, line 671: The parser could be useful here.

obycode: Instead though, it is probably better for the wasm runtime to provide an interface that would just allow you to pass it the snippet and run that.

hugocaillard:
This PR won't be merged and I'll open a new one when getting back on this topic but I'd like to keep it around until then

hugocaillard:
Closing for now. Will revisit soon, see #1155

PR #1116, opened 2023-08-03
add devnet package command

csgui:
This PR adds the `$ clarinet devnet package` command.

The command should run inside a Clarinet project folder.

Issue: #1102

PR #1144, opened 2023-09-06
fix: clarinet-deployments wasm build

hugocaillard:
The WASM build broke with #1116
This is a naive fix proposal

Note: the wasm build will be checked in the upcoming version of the CI


hugocaillard:
Thanks @MicaiahReid!
I hope to make things easier to not break once we get rid of deno and use the new CI

PR #1146, opened 2023-09-07
ci: use rust toolchain v1.71.0

hugocaillard:
The recent issue in the CI is caused by the fact that our version of v8 doesn't build with the latest rustc version (1.72.0 that became the stable release 2 weeks ago).

Let's fix the version to 1.71.0 (latest version able to build clarinet) both in the CI and locally (rust-toolchain file).

In the future, I think we should always use stable (both in CI and locally).
This will be possible once we remove deno and v8 from the dependencies (in the coming weeks)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,16 +6,16 @@ on:
       - develop
       - rc/next
     paths-ignore:
-      - '**/CHANGELOG.md'
-      - 'docs/**'
+      - "**/CHANGELOG.md"
+      - "docs/**"
   push:
     branches:
       - main
       - develop
       - rc/next
     paths-ignore:
-      - '**/CHANGELOG.md'
-      - 'docs/**'
+      - "**/CHANGELOG.md"
+      - "docs/**"
 
   workflow_dispatch:
 
@@ -77,7 +77,7 @@ jobs:
       - name: Install Rust toolchain
         uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
+          toolchain: 1.71.0
           profile: minimal
           components: rustfmt
           override: true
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -6,16 +6,16 @@ on:
       - develop
       - rc/next
     paths-ignore:
-      - '**/CHANGELOG.md'
-      - 'docs/**'
+      - "**/CHANGELOG.md"
```
hugocaillard, line 9: auto format

dockerfiles/components/clarinet.dockerfile
```diff
@@ -2,6 +2,8 @@ FROM debian:bookworm-slim
 
 RUN apt update && apt install -y libssl-dev
 
+RUN rustup update 1.71.0 && rustup default 1.71.0
```
hugocaillard, line 5: This file is not used _yet_
It was introduced in #1104 and rightfully used v1.67.0, reflecting what was set in rust-toolchain at the time

dockerfiles/components/clarinet-dev.dockerfile
```diff
@@ -4,7 +4,7 @@ WORKDIR /src
 
 RUN apt update && apt install -y ca-certificates pkg-config libssl-dev
 
-RUN rustup update 1.67.0 && rustup default 1.67.0
+RUN rustup update 1.71.0 && rustup default 1.71.0
```
hugocaillard, line 7: This file is not used _yet_
It was introduced in #1104 and rightfully used v1.67.0, reflecting what was set in rust-toolchain at the time

MicaiahReid (approved):

Issue #1109, opened 2023-07-31
Build JS Clarinet test lib on top the core SDK


hugocaillard:
The clarinet core JS SDK allows to load a clarinet session and interact with it (call functions, get contracts interfaces - ABI, get devnet.toml accounts and get the assets per addresses/contracts).

While the core SDK will allow to interact with a session in many ways, the test kit should provide a nice and easy way to test clarity contracts. It will be the recommended way to test clarinet projects.

- The test kit library expose a more developer friendly testing API thanks to helpers and wrappers around  `core`
- Should handle code coverage out of the box in a an easy way
- Some assertion features for clarity values and events.
- We should also provide a **clarinet 1 compatibility layer** (`test-compat`) on top of `core` and `test-kit` that will implement the exact same API as the Deno testing library (`Clarinet.test`, `String.prototype.expect*`) to ease the transition of existing projects from the clarinet 1 / deno testing framework to the new one.

sabbyanandan:
@hugocaillard: Could you update the description with some context, please? Curious.

Issue #913, opened 2023-03-01
`[clarity-repl]` Expose method to construct JSON abi from contract source


aulneau:
I'm working on an index of clarity contracts, and I want the ability to convert a string representation (standard .clar contents as string) of a contract into a JSON ABI, similar to the results from the stacks-blockchain-api/node. 

There are lots of use cases I have planned for it, but generally it's so I'm able to construct different views of a contract or find similarities between many.

hugocaillard:
Hey @aulneau,
Is think the idea would be to expose `build_contract_interface` in the repl,
Something like
```rs
/* components/clarity-repl/src/lib.rs */

/* ... */

#[cfg(feature = "wasm")]
#[wasm_bindgen]
pub fn get_contract_interface(name: &str) {
    let mut session = unsafe {
        match WASM_GLOBAL_CONTEXT.session.take() {
            Some(session) => session,
            None => return (),
        }
    };
    build_contract_interface(session. /* get analysis ... */);
    ()
}
```

I could try to dig a bit deeper over the weekend or early next week

smcclellan:
@aulneau  Is this still needed?

hugocaillard:
Just moved this issue in progress, it will be available in the v1.0.0 of the clarinet-sdk #1065 

For context, @aulneau a new clarinet-sdk will be released next week on npm (release candidate 1). So if you're still interested, let's talk 🙌

PR #1151, opened 2023-09-11
[sdk]: vitest config and custom matchers

hugocaillard:
Fix: #1109 
Fix: #913

This PR introduces the Vitest helpers and tools included in the clarinet-sdk (https://github.com/hirosystems/clarinet/pull/1068).

Including the Vitest custom matchers for clarity values (eg `expect(value).toBeUint(1)`)

More docs and examples to come as we prep for the 1.0.0 release.

Once this PR is merged, we are ready to release the SDK v1.0.0-rc1


components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
```diff
@@ -0,0 +1,377 @@
...
+      assert(isCV);
+    } catch (e: any) {
+      return errorToAssertionResult.call(this, e);
+    }
+
+    const expected = Cl.buffer(expectedRaw);
+    return {
+      pass: this.equals(actual, expected, undefined, true),
+      // note: throw a simple message and rely on `actual` and `expected` to display the diff
+      message: () => `the received Buffer does ${this.isNot ? "" : "not "}match the expected one`,
```
janniks, line 317: isNot switched? Could use the `notStr` helper. (and some following cases)

components/clarinet-sdk/vitest-helpers/src/formatCV.ts
```diff
@@ -0,0 +1,151 @@
...
+  BufferCV,
+  ClarityType,
+  ClarityValue,
+  ListCV,
+  TupleCV,
+  principalToString,
+} from "@stacks/transactions";
+
+let byteToHex: string[] = [];
+for (let n = 0; n <= 0xff; ++n) {
```
janniks, line 19: byteToHex could be imported from stacks/common if needed

components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
janniks: Maybe I'm reading this wrong, but the notStr is the opposite, and the tests cases seem to cover these, so idk

janniks (dismissed):

components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
hugocaillard: Yes because the notStr helpers is used for this kind of sentence:

```
expected "x" to be "y"
```

Which could also be
```
the received "y" is not equal to "x"
```

So the "not" isn't used in the same way.

Maybe I should rephrase
```
the received Buffer does not match the expected one
into:
expected the received Buffer to match the expected one"
```

It would be more standard

components/clarinet-sdk/vitest-helpers/src/formatCV.ts
hugocaillard: Ah, I didn't know that, will do.
Also, what do you think about moving formatCV to stacks.js at some point?

components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
janniks: Ah right, makes sense. This is not the "must" style here. 👍🏻

components/clarinet-sdk/vitest-helpers/src/formatCV.ts
janniks: Let's do it 🏎️ `Cl.toString` 😅

hugocaillard: I also explored the possibility of having a Cl.toJSON().

Or a custom JSON replacer. And I realize that I actually pushed it but it's dead code -> https://github.com/hirosystems/clarinet/blob/9a4c82bae11fb22c684c7c875678341570970c20/components/clarinet-sdk/vitest-helpers/src/formatCV.ts#L142-L151)

(It has some flaws as it is, and i'll remove it anyway)

So you can do `JSON.stringify(ClarityValue, cvJSONReplacer, 2)`
```
{
  "id": "u1",
  "messageAscii": "\"hello world\"",
  "items": [
    {
      "id": "u1",
      "owner": "(some 'ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG)",
    },
  ]
}
```

And it outputs valid JSON while taking care of the indentation (and it would have avoided me to handle the `indent & depth` when formatting lists and tuples.

But what we actually want isn't valid JSON but valid Clarity values (without `""` everywhere and with `(list )` instead of `[]`). Anyway, not sure there is a use for it but interesting to explore.

janniks: Very interesting!

MicaiahReid (changes requested):
I haven't tested yet, but I've started reviewing!

components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
```diff
@@ -0,0 +1,377 @@
...
+// @ts-ignore
+BigInt.prototype.toJSON = function () {
+  return this.toString();
+};
+
+function notStr(isNot: boolean) {
+  return isNot ? "not " : "";
+}
+
+function formatMessage(this: any, received: string, expected: string) {
```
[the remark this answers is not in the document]
hugocaillard: Indeed, thanks

components/clarinet-sdk/vitest-helpers/src/formatCV.ts
```diff
@@ -0,0 +1,130 @@
...
+  const hexOctet = n.toString(16).padStart(2, "0");
+  byteToHex.push(hexOctet);
+}
+
+function formatSpace(space: number, depth: number, end = false) {
```
janniks, line 24: We could use these from stacks.js if you want

janniks (approved):
Looks great!

components/clarinet-sdk/vitest-helpers/src/formatCV.ts
hugocaillard: Yep, I will use it in v1 now that it's available. But v0.1 was already released like that so.. let's merge the PR as it is.
My process of releasing before it's merge isn't super conventional but well, it's a v0

components/clarinet-sdk/README.md
```diff
@@ -1,20 +1,33 @@
...
 ## Core
 
+```
+npm install @hirosystems/clarinet-sdk
+```
+
 ### Usage
 
 ```ts
+import { initVM } from "clarinet-sdk";
```
hugocaillard, line 21:
```suggestion
import { initVM } from "@hirosystems/clarinet-sdk";
```

MicaiahReid (approved):

hugocaillard:
Thanks @MicaiahReid @janniks 
Let's goo!

PR #1162, opened 2023-09-25
feat: sdk: get contract ast

hugocaillard:
### Description

Slightly change the Rust Struct of `SDK` to hold the whole contracts analysis (instead  of `contracts_sources: Hashmap, contract_asts: Hashmap` etc.

The `ContractAST` typescript type is still very basic (can't easily be transform from Rust to TypeScript because of cyclic properties) and can be greatly improved. For now, I wanted to have to recursive nature of `expr.List`.

The project with missing a prettier config so the formatting could get inconsistent. Add one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1038,6 +1038,7 @@ dependencies = [
  "serde",
  "serde-wasm-bindgen 0.6.0",
  "serde_json",
+ "typescript-type-def",
  "wasm-bindgen",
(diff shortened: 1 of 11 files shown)


MicaiahReid:
This isn't really important, but I think if you want to include a scope on a conventional commit message, it would be: `feat(sdk): get contract ast`

MicaiahReid (approved):
Very nice!

Issue #749, opened 2023-01-02
Output cost data from tests to file


whoabuddy:
While working on the CityCoins protocol, there were a few revisions that were entirely based around reducing contract costs.

The output of `clarinet test --costs` is very useful, but comparing this against two points in time is difficult, and for now we're dumping the information to a text file and using a manual diff for comparisons.

![image](https://user-images.githubusercontent.com/9038904/210280655-600c201b-9287-4bc6-9629-ab73e96ab519.png)

This works for now - but got me thinking it would be great if the data used to generate the costs table presented there was output in another format, maybe JSON?

hugocaillard:
This feature will be available in @hirosystems/clarinet-sdk@0.5.0

PR #1165, opened 2023-09-27
Feat/sdk improvements

hugocaillard:
### Description

Three additions:

- add an 'options" argument to `deployContract()` it currently allows to specify the clarity_version, but I made it an options object so that we can add more options in the futures with limited or no breaking changes
- ability to get cost reports and and save it to a json file in tests
- change the vitest-helper with an `initBeforeEach` options controlling when the vm state should be re-initialized

Plus multiple small improvements

Fix: #749


components/clarinet-sdk/src-ts/vitest/index.ts
```diff
@@ -1,3 +1,4 @@
+import { bool } from "@stacks/transactions/dist/cl";
```
janniks, line 1: Is this used?

janniks (approved):
Looking good 😎

components/clarinet-sdk/src-ts/vitest/index.ts
hugocaillard: Ah, my bad, I removed it but in the follow up PR: https://github.com/hirosystems/clarinet/pull/1171/files#diff-c26bfb8a8ec5511c86cae0ab7e151b28db319eccdc06e2992fb679b4c5a8fb23L1

Marking as resolve for this one

hugocaillard:
Thanks @janniks!

PR #1171, opened 2023-10-02
chore(sdk): new package structure, splitting into two packages

hugocaillard:
### Description

The clarinet SDK was bundled with wasm-pack, which was kind of useful as a poc in the development phase. But it doesn't make sense to ship the lib as one big bundle.
Because of that, the .wasm file was duplicated on NPM (for esm and cjs) so that was 2 times 7mb.

With this new struct, we directly publish the wasm-pack lib under `@hirosystems/clarinet-sdk-wasm`.
The existing `@hirosystems/clarinet-sdk` only depends on it


package.json
```diff
@@ -0,0 +1 @@
+{}
```
janniks, line 1: Should this have content? Do we need it? (same for outermost lock file)

janniks (dismissed):

package.json
hugocaillard: This file shouldn't exist, my bad, thanks @janniks

hugocaillard:
Looking good, I'm merging this branch that was previously approved (had to dismiss review to rebase on develop and remove dead code)

PR #1172, opened 2023-10-02
fix(sdk): deserialize events values as clarity values

hugocaillard:
### Description

Clarity event values should be transform to Stack.js values just like contract call responses


janniks (dismissed):

MicaiahReid (approved):

hugocaillard:
Thanks @janniks and thanks @MicaiahReid for the reviews

PR #1175, opened 2023-10-03
Update faq.md: FAQ's -> FAQs

saralab:
FAQs NOT FAQ's


hugocaillard (approved):
LGTM 
Thanks Sarala

hugocaillard:
@aravindgee Is it possible that the actions don't run because it's a .md file, but then it prevents us from merging? 🤔

aravindgee:
@hugocaillard , yes, [this is because of the filter](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/troubleshooting-required-status-checks#handling-skipped-but-required-checks) to skip the docs directory. I've merged the PR with bypassing branch rules.

hugocaillard:
@aravindgee Ah. But can we make it work without bypassing branch protection?
I'd like to have it working for future PRs as well

aravindgee:
Hi @hugocaillard happy to assist but I need to investigate this a bit. would you mind creating an issue on the devops board? if this isn't a priority item, we will get to it during a future planning iteration.

hugocaillard:
hey @aravindgee 
Thanks for you help. 
I tried to find a solution and it looks like a solution is to create a script in the workflow that'll call the github API and check if the PR contains changes on .md files. 

But it might be overkill for now.

I'll remember to open an issue on the devops board eventually. We'll have some work to do on our CI in the coming weeks/months anyway 👍 

Thx!

PR #1176, opened 2023-10-03
Update introduction.md: Clean up

saralab:
"What is Clarinet?" right under Clarinet Overview is redundant.  The section also dives right into what Clarinet does instead of defining it


hugocaillard (approved):
LGTM

PR #1180, opened 2023-10-04
Refactor(sdk): run vitest-helpers tests as part of clarinet-sdk tests

hugocaillard:
### Description

The vitest-helpers directory contained its own package.json and node modules to run the tests. While it makes more sense to run the tests as part of the clarinet-sdk package.
It will also be more maintainable


hugocaillard:
Merging this one without the process to speed up the process. No impact on the published package but nice to have the pr included in the next version

PR #1181, opened 2023-10-04
refactor(sdk): rename vm to simnet

hugocaillard:
### Description

In the the clarinet-sdk, the naming `vm` as in `initVM` and `ClarityVM` where never satisfying. But we kept it lacking better alternative.

After talking with Ludo, we agreed that `simnet` was a better option
- it'll explicit tell developers what the `simnet` (it's basically what you interact with when using the SDK)
- it's not super intuitive, but `vm` wasn't either
- `simnet.mineBlock()` makes wayyy more sense than `vm.mineBlock()` (which doesn't make any)

This PR only changes the JS layer since the rust code doesn't make any mention of the `VM` wording in this context.

The Rust layer (components/clarinet-sdk-wasm) exposes a SDK constructor (which could be improved but is kind of okay. And isn't visible to end user since it's abstracted by the JS layer). It exposes an `init_session` methods that matches the rest of the clarinet code base, such as `initiate_session_from_deployment`). This method isn't meant to be used by end developers either

### DIFF
--- a/components/clarinet-sdk/README.md
+++ b/components/clarinet-sdk/README.md
@@ -18,19 +18,19 @@ npm install @hirosystems/clarinet-sdk
 ### Usage
 
 ```ts
-import { initVM } from "@hirosystems/clarinet-sdk";
+import { initSimnet } from "@hirosystems/clarinet-sdk";
 import { Cl } from "@stacks/transactions";
 
 async function main() {
-  const vm = await initVM();
+  const simnet = await initSimnet();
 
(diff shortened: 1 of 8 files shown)


lgalabru (dismissed):
Reading the code makes more sense to me - love it 😍
Thanks @hugocaillard!

lgalabru (approved):
🚀

csgui:
Makes more sense, indeed! 🚀

PR #1197, opened 2023-10-09
Update clarinet-js-sdk.md

max-crawford:
Grammar fixes

### DIFF
--- a/docs/feature-guides/clarinet-js-sdk.md
+++ b/docs/feature-guides/clarinet-js-sdk.md
@@ -43,7 +43,8 @@ async function main() {
 main();
 ```
 
-By default, the SDK will look for a Clarinet.toml file in the current working directory. It's also possible to provide the path to the manifest like so:
+By default, the SDK will look for a Clarinet.toml file in the current working directory.
+It's also possible to provide the path to the manifest like so:
 
 ```ts
 const simnet = await initSimnet("./path/to/Clarinet.toml");
--- a/docs/feature-guides/clarinet-js-sdk.md
+++ b/docs/feature-guides/clarinet-js-sdk.md
@@ -165,7 +165,8 @@ const counter = simnet.getDataVar("counter", "count");
 getMapEntry(contract: string, mapName: string, mapKey: ClarityValue): ClarityValue
 ```
 
-Get the value of a map entry by its key. Note that it will always return an optional value (`(some <value>)` or `none`). Just like Clarity `map-get?`.
+Get the value of a map entry by its key.
+Note that it will always return an optional value (`(some <value>)` or `none`). Just like Clarity `map-get?`.
 
 Given a contract with the following definition:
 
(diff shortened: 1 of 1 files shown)


docs/feature-guides/clarinet-js-sdk.md
```diff
@@ -43,23 +43,21 @@ async function main() {
 main();
 ```
 
-By default, the SDK will look for a Clarinet.toml file in the current working directory.  
-It's also possible to provide the path to the manifest like so:
+By default, the SDK will look for a Clarinet.toml file in the current working directory. It's also possible to provide the path to the manifest like so:
```
hugocaillard, line 46:
```suggestion
By default, the SDK will look for a Clarinet.toml file in the current working directory.
It's also possible to provide the path to the manifest like so:
```

docs/feature-guides/clarinet-js-sdk.md
```diff
@@ -43,23 +43,21 @@ async function main() {
...
 
 ### `initSimnet`
 
 ```ts
 initSimnet(manifestPath?: string): Promise<Simnet>
 ```
 
-The `initSimnet` function takes the manifest path (`Clarinet.toml`) as an optional argument. By default, it'll look for a manifest in the current working directory.  
-It will often be the first function to call when using the SDK.
+The `initSimnet` function takes the manifest path (`Clarinet.toml`) as an optional argument. By default, it'll look for a manifest in the current working directory. It will often be the first function to call when using the SDK.
```
hugocaillard, line 60:
```suggestion
The `initSimnet` function takes the manifest path (`Clarinet.toml`) as an optional argument. By default, it'll look for a manifest in the current working directory.
It will often be the first function to call when using the SDK.
```

docs/feature-guides/clarinet-js-sdk.md
```diff
@@ -165,8 +163,7 @@ const counter = simnet.getDataVar("counter", "count");
 getMapEntry(contract: string, mapName: string, mapKey: ClarityValue): ClarityValue
 ```
 
-Get the value of a map entry by its key.  
-Note that it will always return an optional value (`(some <value>)` or `none`). Just like Clarity `map-get?`.
+Get the value of a map entry by its key. Note that it will always return an optional value (`(some <value>)` or `none`). Just like Clarity `map-get?`.
```
hugocaillard, line 166:
```suggestion
Get the value of a map entry by its key.
Note that it will always return an optional value (`(some <value>)` or `none`). Just like Clarity `map-get?`.
```

docs/feature-guides/clarinet-js-sdk.md
```diff
@@ -202,8 +199,7 @@ callReadOnlyFn(
 ): ParsedTransactionRes
 ```
 
-Call read-only functions exposed by a contract. Returns an object with the result of the function call as a Clarity Value.  
-It takes function arguments in the form in Clarity Values, available in the package `@stacks/transactions`.
+Call read-only functions exposed by a contract. This method returns an object with the result of the function call as a Clarity Value. It takes function arguments in the form in Clarity Values, available in the package `@stacks/transactions`.
```
hugocaillard, line 202:
```suggestion
Call read-only functions exposed by a contract. This method returns an object with the result of the function call as a Clarity Value.
It takes function arguments in the form in Clarity Values, available in the package `@stacks/transactions`.
```

hugocaillard (changes requested):
@max-crawford 

I might have put too many line breaks in this docs.
Still requesting some change to improve the markdown formatting. The impacted lines will still be rendered on 1 line, but more readable in markdown.

--- 

In markdown, a normal line break isn't be rendered, it just helps keeping line shorts, making future changes and reviews easier.

This will be rendered on the same line:
```md
First line.<line-break>
End of first line
```

This will be rendered on two lines but in 1 paragraph:
```md
Beginning of paragraph.<space><span><line-break>
End of paragraph
```

And we actually need to line breaks to make two paragraphs
```md
First paragraph.<line-break><line-break>
Second paragraph
```

ryanwaits (dismissed):

PR #1201, opened 2023-10-11
docs: add comment to the vitest config template

hugocaillard:


components/clarinet-sdk/templates/vitest.config.js
```diff
@@ -3,13 +3,34 @@
 import { defineConfig } from "vite";
 import { vitestSetupFilePath, getClarinetVitestsArgv } from "@hirosystems/clarinet-sdk/vitest";
 
+/*
+  In this file, we configure Vitest so that it works well with Clarinet and the Simnet.
```
hugocaillard, line 7:
```suggestion
  In this file, Vitest is configured so that it works seamlessly with Clarinet and the Simnet.
```

PR #1202, opened 2023-10-11
replace . with _ for contract names

### LINKED ISSUE #1031: `clarinet contracts` allows creating contracts with illegal names
I created a contract by typing

```sh
clarinet contracts new subnet-v2.0.0
```

This created the `subnet-v2.0.0.clar` file and updated `Clarinet.toml`, but `clarinet integrate` failed to start because this is not a valid contract name (I guess `.` is not a legal character)

Clarinet should check for illegal characters, and if present, should either fail or substitute legal characters like `-` or `_`

tippenein:
### Description

fixes #1031 by replacing `.` with `_` in contract file names

#### Breaking change?

N/A

### Example

Screenshot 2023-10-11 at 4 01 54 PM

---

### Checklist

If there are applicable tests to add, I'm glad to do that. I was unable to find them tho, so I added the screenshots above

### DIFF
--- a/components/clarinet-cli/src/generate/contract.rs
+++ b/components/clarinet-cli/src/generate/contract.rs
@@ -20,7 +20,7 @@ impl GetChangesForNewContract {
     ) -> Self {
         Self {
             manifest_location,
-            contract_name,
+            contract_name: contract_name.replace(".", "_"),
             source,
             changes: vec![],
         }


tippenein:
Still looking for documentation on what characters are allowed for clarity file names, but will update if I find anything beyond what was reported in the issue

tippenein:
> Still looking for documentation on what characters are allowed for clarity file names, but will update if I find anything beyond what was reported in the issue

It looks like you can get around the check if you use `"` but this seems outside of plausible use. 

```
cargo run contracts new char."?"
```

will create a new contract called `char_?` but won't be able to run. Seems like anyone doing this would expect trouble though..

hugocaillard:
I think it makes sense as it is.

For other unexpected use cases, we probably want to throw an error instead. It can be addressed in a future PR.

Can you sign CLA please?

hugocaillard (approved):
LGTM, thanks a lot for your contribution @tippenein

Issue #1164, opened 2023-09-27
session.deployContract should not allow new Clarity Version to be deployed on old Epoch


hugocaillard:
**Describe the bug**

The method [`deployContract`](https://github.com/hirosystems/clarinet/blob/90095061037e0abfe1043443d8a4e5e5bc3d2dff/components/clarity-repl/src/repl/session.rs#L545) currently allows to deploy a contract with ClarityVersion 2 on Epoch 2.0, which should be illegal

PR #1203, opened 2023-10-12
fix(sdk): Epoch and clarity version mismatch

tippenein:
fix: #1164 
 
 - deploy_contract now returns an Err if it's called with Clarity v2 and Epoch 2.0

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -551,13 +551,16 @@ impl Session {
         test_name: Option<String>,
         ast: &mut Option<ContractAST>,
     ) -> Result<ExecutionResult, Vec<Diagnostic>> {
-        if contract.epoch == StacksEpochId::Epoch20 && contract.clarity_version == ClarityVersion::Clarity2 {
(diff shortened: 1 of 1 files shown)


components/clarity-repl/src/repl/session.rs
```diff
@@ -551,6 +551,15 @@ impl Session {
         test_name: Option<String>,
         ast: &mut Option<ContractAST>,
     ) -> Result<ExecutionResult, Vec<Diagnostic>> {
+        if contract.epoch == StacksEpochId::Epoch20 && contract.clarity_version == ClarityVersion::Clarity2 {
```
hugocaillard, line 554: Good catch!
We have the same logic to handle clarity version and epoch that are set in the Clarinet.toml.
https://github.com/hirosystems/clarinet/blob/30e6807997be3071cfde6dff1f1851724eadd847/components/clarinet-files/src/project_manifest.rs#L163-L167

Could you add a similar condition?
It should help handling future epochs and clarity versions

hugocaillard (changes requested):
Thanks for your second PR @tippenein.
Made a suggestion

hugocaillard (approved):
LGTM! Thanks @tippenein

PR #1205, opened 2023-10-13
fix: detect dependencies defined in const

hugocaillard:
### Description

A dependency to a contract can exist in a constant.

eg:

This type of dependency wasn't detected before

```clarity
(define-constant abtc-token .token-abtc)

(define-public (test)
  (try! (contract-call? abtc-token ...))
)
```

### DIFF
--- a/components/clarity-repl/src/analysis/ast_dependency_detector.rs
+++ b/components/clarity-repl/src/analysis/ast_dependency_detector.rs
@@ -621,12 +621,12 @@ impl<'a> ASTVisitor<'a> for ASTDependencyDetector<'a> {
     fn visit_dynamic_contract_call(
         &mut self,
         expr: &'a SymbolicExpression,
-        trait_ref: &'a SymbolicExpression,
+        callable_expr: &'a SymbolicExpression,
         function_name: &'a ClarityName,
         args: &'a [SymbolicExpression],
     ) -> bool {
-        let trait_instance = trait_ref.match_atom().unwrap_or(&DEFAULT_NAME);
-        if let Some(trait_identifier) = self.get_param_trait(trait_instance) {
+        let callable = callable_expr.match_atom().unwrap_or(&DEFAULT_NAME);
+        if let Some(trait_identifier) = self.get_param_trait(callable) {
             let dependencies = if let Some(trait_definition) = self.defined_traits.get(&(
                 &trait_identifier.contract_identifier,
                 &trait_identifier.name,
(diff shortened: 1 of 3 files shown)


components/clarity-repl/src/analysis/ast_dependency_detector.rs
```diff
@@ -667,6 +667,20 @@ impl<'a> ASTVisitor<'a> for ASTDependencyDetector<'a> {
...
+    fn visit_define_constant(
+        &mut self,
+        expr: &'a SymbolicExpression,
+        name: &'a ClarityName,
+        value: &'a SymbolicExpression,
+    ) -> bool {
+        if let Some(Value::Principal(PrincipalData::Contract(contract_principal))) =
+            value.match_literal_value()
+        {
+            self.add_dependency(self.current_contract.unwrap(), contract_principal);
```
obycode, line 680: This is technically overly cautious, as you could store the contract principal in a constant and never use it, or only use it as a principal, but not use it as the target of a `contract-call?`. If we're okay taking this overly cautious approach, then I think the whole dependency detector could be simplified to just an implementation of `visit_literal_value` that looks for a contract principal.

hugocaillard: Ah yes of course.
So I should check that the constant is effectively used in a `call-contract?`

obycode: Yeah, and I just realized you might have to. I think I remember ALEX or Arkadiko having an unusual situation where contract A references B, but B calls A, so that would report a circular dependence if we tried the simplification.

hugocaillard: Ah you mean they kind of bypass the circular dependency detection by storing contract in constants. 
That might be an issue indeed. The repo I'm currently testing against is a new Alex set up

obycode: Right, like they need to pass the contract principal as an argument, but not to actually call it. Something like that.

hugocaillard: Ah ok I see what you mean. 

My worry now is that we were probably missing circular dependencies (because we were missing dependencies). So some projects might break with the fix. 
But I guess it makes more sense

obycode: If there were real circular dependencies, they wouldn't have been able to deploy any way, so I'd say it's fine if you break those.

hugocaillard:
Thanks @obycode 
Let's release it and work on the follow up PR 👌

PR #1207, opened 2023-10-15
feat(sdk): allow setting env var for telemetry acceptance globally

tippenein:
feat: #205 

This is one of the proposals from #205 

I wasn't sure where to put the documentation, but FAQ seemed like a decent fallback

Issue #205, opened 2022-01-06
Environment variable for telemetry


obycode:
I think it would be nice to optionally have an environment variable that holds my selection for opting in/out of telemetry, since developers are likely to choose the same option for every project and it is a little bit annoying to answer the question every time you create a new project.

My UI suggestion:
```
$ clarinet new demo
Send usage data to Hiro.
Help Hiro improve its products and services by automatically sending diagnostics and usage data.
Only high level usage information, and no information identifying you or your project are collected.
Enable [Y/n]?y

Telemetry enabled. Thanks for helping to improve clarinet!
To skip this prompt next time and always enable:
export CLARINET_TELEMETRY=1

Creating directory demo
Created file demo/contracts
Created file demo/settings
...
```

csgui:
Hi @obycode and @lgalabru !

The idea to use environment variables to control some aspects of the `clarinet` is valid. Great idea... but the environment variable is not automatically persisted in the shell environment, it will work only once, unless the shell configuration is changed.

So - throwing an idea to discussion - what you think in adding a more robust configuration feature to `clarinet`? I mean a config file like the `.gitconfig` (or any other format) where `clarinet` can leverage of global and local (per project) configurations.

Cheers.

obycode:
Hi @csgui. Thanks for the input. It does seem likely that we will have more customization options as time goes on, so you're right, it could be nice to create a config file. 

Right now, I believe we just have the variable `CLARINET_DISABLE_HINTS` and this proposed new one.

lgalabru:
Thanks @obycode, yeah good idea.
Good point @csgui, I think a global config file would be a good move.
It is actually going to be essential for the "Clarinet and beyond" project.

csgui:
Hey @lgalabru and @obycode !

How the work of a feature like that - a config file - would be planned? I'd love to help on that. :-)

lgalabru:
Hey @csgui! 
You could look at how we're dealing with the `Clarinet.toml` manifest files: we essentially have a `ProjectManifestFile` where most of the fields are optional, then we have a `ProjectManifest` struct where all the fields optional in the 1st struct are required. `ProjectManifest` can be constructed from a `ProjectManifestFile`, and the values absent from the file are completed with default values.
By default, clarinet could be trying to load this file from `~/.clarinet/Settings.toml`. I'd like to setup a directory (vs just a file) because that directory could be reused for other things, like caching for instance.
What do you think?

csgui:
[...]

PR #1208, opened 2023-10-15
feat(sdk): allow global ~/.clarinet/Settings.toml

tippenein:
feat: #205 

This adds a global file for clarinet settings

```
cat ~/.clarinet/Settings.toml 
disable_hints = true
enable_telemetry = true
```
If the file doesn't exist, it picks the defaults

```
GlobalSettings {
  disable_hints: false,
  enable_telemetry: None
};
```

Any thoughts welcome!

I can add documentation wherever seems appropriate

(This should nullify #1207)


components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -579,7 +584,26 @@ pub fn main() {
...
     let hints_enabled = env::var("CLARINET_DISABLE_HINTS") != Ok("1".into());
+    let home_dir = dirs::home_dir().expect("Unable to get home directory");
+    let path = home_dir.join(".clarinet/Settings.toml");
+
+    let global_settings: GlobalSettings = if path.exists() {
+        let mut file = File::open(&path).expect("Unable to open the file");
+        file.read_to_string(&mut global_settings)
+            .expect("Unable to read the file");
+
+        let result = toml::from_str(&global_settings).expect("Unable to parse the TOML file");
```
hugocaillard, line 602: the `result` variable can be omitted

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -579,7 +584,26 @@ pub fn main() {
...
+    };
+    // This is backwards compatible with ENV var setting as well as the new ~/.clarinet/Settings.toml
     let hints_enabled = env::var("CLARINET_DISABLE_HINTS") != Ok("1".into());
+    let home_dir = dirs::home_dir().expect("Unable to get home directory");
+    let path = home_dir.join(".clarinet/Settings.toml");
+
+    let global_settings: GlobalSettings = if path.exists() {
+        let mut file = File::open(&path).expect("Unable to open the file");
+        file.read_to_string(&mut global_settings)
+            .expect("Unable to read the file");
```
hugocaillard, line 600: I think it would be better to log an warning instead of panicking whenever something unexpected happens.
Also, the error message could be more explicit: `"the file"` -> `"the ~/.clarinet/Settings.toml file"`

hugocaillard (changes requested):
Thanks for your contribution @tippenein 
Small improvements requested

components/clarinet-cli/src/frontend/cli.rs
tippenein: Used format_warn for this
Screenshot 2023-10-24 at 6 05 15 PM

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -579,7 +584,26 @@ pub fn main() {
...
     };
 
+    let mut global_settings = String::new();
+    let global_settings_default = GlobalSettings {
+        disable_hints: false,
+        enable_telemetry: None,
+    };
+    // This is backwards compatible with ENV var setting as well as the new ~/.clarinet/Settings.toml
     let hints_enabled = env::var("CLARINET_DISABLE_HINTS") != Ok("1".into());
+    let home_dir = dirs::home_dir().expect("Unable to get home directory");
```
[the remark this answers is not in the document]
tippenein: Changed this to just use the `Option<PathBuf>`. Didn't test it but it's nested int here with the rest of the err cases

hugocaillard (approved):
LGTM @tippenein 

Thanks a lot for this contribution and quick responses!

PR #1220, opened 2023-10-24
fix: missing coverage in clarity iterator functions

hugocaillard:
The first argument of `map`, `filter` and `fold` doesn't trigger `will_eval_hook`. Leading to false negative in coverage reports.
But it's safe to assume that this expr is hit of the map, filter or fold is hit as well.


lgalabru (approved):
😍 and thank you for adding the tests!

PR #1223, opened 2023-10-24
refactor: clippy

hugocaillard:
Set up clippy on the repo.

Fix around 500 or 600 hundred clippy warnings and a few clippy errors.

Lots of the fixes are very minors. A few of them may lead to some optimization.
Overall, using clippy makes the code more consistent and safe. It also help while refactoring code or working on new features

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -447,6 +447,9 @@ pub async fn generate_default_deployment(
             // Extract the known / unknown dependencies
             match dependencies {
                 Ok(inferable_dependencies) => {
+                    if inferable_dependencies.len() > 1 {
+                        println!("warning: inferable_dependencies contains more than one entry");
+                    }
                     // We submitted a HashMap with one contract, so we have at most one result in the `inferable_dependencies` map.
                     // We will extract and keep the associated data (source, ast, deps).
                     if let Some((contract_id, dependencies)) =
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -42,7 +42,7 @@ pub fn setup_session_with_deployment(
     deployment: &DeploymentSpecification,
     contracts_asts: Option<&BTreeMap<QualifiedContractIdentifier, ContractAST>>,
 ) -> DeploymentGenerationArtifacts {
-    let mut session = initiate_session_from_deployment(&manifest);
+    let mut session = initiate_session_from_deployment(manifest);
     update_session_with_genesis_accounts(&mut session, deployment);
     let results = update_session_with_contracts_executions(
         &mut session,
@@ -61,12 +61,9 @@ pub fn setup_session_with_deployment(
         match res {
             Ok(execution_result) => {
                 diags.insert(contract_id.clone(), execution_result.diagnostics);
-                match execution_result.result {
-                    EvaluationResult::Contract(contract_result) => {
-                        asts.insert(contract_id.clone(), contract_result.contract.ast);
-                        contracts_analysis.insert(contract_id, contract_result.contract.analysis);
-                    }
-                    _ => (),
+                if let EvaluationResult::Contract(contract_result) = execution_result.result {
+                    asts.insert(contract_id.clone(), contract_result.contract.ast);
+                    contracts_analysis.insert(contract_id, contract_result.contract.analysis);
                 }
             }
             Err(errors) => {
@@ -76,23 +73,23 @@ pub fn setup_session_with_deployment(
         }
     }
 
-    let artifacts = DeploymentGenerationArtifacts {
+    DeploymentGenerationArtifacts {
         asts,
         deps,
         diags,
         success,
         session,
         analysis: contracts_analysis,
-    };
-    artifacts
+    }
 }
 
 pub fn initiate_session_from_deployment(manifest: &ProjectManifest) -> Session {
-    let mut settings = SessionSettings::default();
-    settings.repl_settings = manifest.repl_settings.clone();
-    settings.disk_cache_enabled = true;
-    let session = Session::new(settings);
-    session
+    let settings = SessionSettings {
+        repl_settings: manifest.repl_settings.clone(),
+        disk_cache_enabled: true,
+        ..Default::default()
+    };
+    Session::new(settings)
 }
 
 pub fn update_session_with_genesis_accounts(
@@ -135,7 +132,7 @@ pub fn update_session_with_contracts_executions(
             (None, Some(min_epoch)) => std::cmp::max(min_epoch, DEFAULT_EPOCH),
             _ => DEFAULT_EPOCH,
         };
-        session.update_epoch(epoch.clone());
+        session.update_epoch(epoch);
         for transaction in batch.transactions.iter() {
             match transaction {
                 TransactionSpecification::RequirementPublish(_)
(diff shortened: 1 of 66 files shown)


components/clarinet-deployments/src/lib.rs
```diff
@@ -454,14 +449,23 @@ pub async fn generate_default_deployment(
...
-                        break;
                     }
+                    // for (contract_id, dependencies) in inferable_dependencies.into_iter() {
+                    //     for dependency in dependencies.iter() {
+                    //         queue.push_back((dependency.contract_id.clone(), None));
+                    //     }
+                    //     requirements_deps.insert(contract_id.clone(), dependencies);
+                    //     requirements_data.insert(contract_id.clone(), (clarity_version, ast));
+                    //     break;
+                    // }
```
hugocaillard, line 468: @lgalabru 
I left that commented for now, do you think the new version works the same way?
It looks like it would always break after the first iteration right?
Or did I miss something?

components/clarinet-deployments/src/lib.rs
```diff
@@ -452,15 +447,16 @@ pub async fn generate_default_deployment(
...
+                        inferable_dependencies.into_iter().next()
+                    {
                         for dependency in dependencies.iter() {
                             queue.push_back((dependency.contract_id.clone(), None));
                         }
                         requirements_deps.insert(contract_id.clone(), dependencies);
                         requirements_data.insert(contract_id.clone(), (clarity_version, ast));
-                        break;
                     }
                 }
```
hugocaillard, line 461: @lgalabru Can you double check that I didn't break anything here?

Instead of doing only one iteration in a for loop, do `inferable_dependencies.into_iter().next()`

components/stacks-network/src/orchestrator.rs
```diff
@@ -2266,63 +2251,73 @@ events_keys = ["*"]
...
 
         if let Some(ref bitcoin_node_container_id) = self.bitcoin_node_container_id {
             let _ = docker
                 .kill_container(bitcoin_node_container_id, options.clone())
                 .await;
             ctx.try_log(|logger| slog::info!(logger, "Terminating bitcoin-node"));
-            let _ = docker.remove_container(bitcoin_node_container_id, None);
+            let _ = docker
+                .remove_container(bitcoin_node_container_id, None)
+                .await;
```
hugocaillard, line 2276: The await here and the lines below could fix the case where a container isn't properly removed at the end of `clarinet integrate` (leading to conflict when running integrate again)

components/clarinet-deployments/src/lib.rs
lgalabru: The code seems equivalent yes. Should we log `warn` if `dependencies.len` > 1?

hugocaillard: Will do

lgalabru: Or should we use `inferable_dependencies.first()` instead?

components/clarinet-files/src/network_manifest.rs
```diff
@@ -919,9 +912,9 @@ impl NetworkManifest {
                     .docker_platform
                     .unwrap_or(DEFAULT_DOCKER_PLATFORM.to_string()),
             };
-            if !config.disable_stacks_api && config.disable_stacks_api {
-                config.disable_stacks_api = false;
-            }
+            // if !config.disable_stacks_api && config.disable_stacks_api {
+            //     config.disable_stacks_api = false;
+            // }
```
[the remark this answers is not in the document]
hugocaillard: Ah yes! I wanted to add a comment on github about this one. I'll remove it!

components/clarity-repl/src/repl/debug/cli.rs
```diff
@@ -129,7 +129,6 @@ impl CLIDebugger {
                         "{}:{}:{}",
                         contract_id, expr.span.start_line, expr.span.start_column
                     );
-                    return;
```
lgalabru, line 132: That seems suspect? Why is the `return` removed here?

hugocaillard: There is no other code path after this `None` branch

components/clarity-repl/src/repl/tracer.rs
```diff
@@ -156,8 +156,8 @@ impl EvalHook for Tracer {
             .flat_map(|b| &b.events)
             .collect::<Vec<_>>();
         if emitted_events.len() > self.emitted_events {
-            for i in self.emitted_events..emitted_events.len() {
-                let event = emitted_events[i];
+            for event in emitted_events.iter().skip(self.emitted_events) {
+                // let event = emitted_events[i];
```
lgalabru, line 160: Readability is not improved in my opinion, but if we're fine with that, let's get rid fo line 160?

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -679,14 +679,7 @@ async fn handle_bitcoin_mining(
...
     let stop_miner = Arc::new(AtomicBool::new(false));
-    loop {
-        let command = match mining_command_rx.recv() {
-            Ok(cmd) => cmd,
-            Err(_e) => {
-                // TODO(lgalabru): cascade termination
-                break;
-            }
-        };
+    while let Ok(command) = mining_command_rx.recv() {
```
lgalabru, line 682: I think I'd keep the initial version and add a log in the `Err` code path (and keep that todo)

lgalabru (approved):
Thanks @hugocaillard! It's a lot of changes indeed. It's looking good overall.
I am not informed on the current state of `develop` (and feel free to disregard if you think it's overkill), but how would you feel about running the release playbook on this PR? that way if we have regressions, our scope of investigation is reduced.

hugocaillard:
@sabbyanandan 
Yes, the PR basically fixes >500 clippy warnings!

@lgalabru 

I already ran a light version on the playbook 👍 
Also tested the vscode extension build and the [sdk unit test](https://github.com/hirosystems/clarinet/blob/bbe26ccf3eaf79e18949c96e6f90523a325f97ca/components/clarinet-sdk/tests/index.test.ts)

I'll address your comments, and will do some manual testing on `clarinet integrate`.

Then I feel pretty safe to merge on develop.
Tomorrow I'll carefully run the full playbook, if anything odd seems related to this PR and hard to debug. I'll probably revert it before 2.0 (which I really want to get ready by tomorrow)

components/clarinet-deployments/src/lib.rs
hugocaillard: `inferable_dependencies` is a BTreeMap and `BTreeMap::first_entry()` works a bit differently.
With `into_iter().next()` we stay very close to the initial implementation

components/clarity-repl/src/repl/tracer.rs
hugocaillard: Agree on readability.

I disabled the lint rule `clippy::needless_range_loop` in a few places where it made things worth.
This one didn't seems to make a big difference to me (both version are a bit difficult)

I renamed `emitted_events` into `nb_of_emitted_events` to make it a bit better

components/clarinet-deployments/src/lib.rs
hugocaillard: Added a log

hugocaillard:
Bypassing branch protection to merge. The approval was dismissed by my review commit

PR #1225, opened 2023-10-25
refactor: improve global settings management

hugocaillard:
Follow up #1208 

@tippenein I might have approved your previous PR a bit too quickly.
In your implementation, if telemetry is disabled globally, it would still ask for each project to enable telemetry Y/n, while i think it shouldn't.
I took this liberty to work on the follow-up PR to make sure we can ship it in 2.0 👌 
I also updated the setting file to be `~/.clarinet/clarinetrc.toml`, this way, we can more easily provide a json schema file for it in the future if we want to (vs `Settings.toml` that seems to generic)

This PR also updates the hints messages

Fixes: #205

### DIFF
--- a/components/clarinet-cli/src/frontend/clarinetrc.rs
+++ b/components/clarinet-cli/src/frontend/clarinetrc.rs
@@ -1,0 +1,53 @@
+use std::fs::{self};
+
+use std::env;
+
+#[derive(Serialize, Deserialize, Default)]
+pub struct GlobalSettings {
+    pub enable_hints: Option<bool>,
+    pub enable_telemetry: Option<bool>,
+}
+
+impl GlobalSettings {
+    pub fn get_settings_file_path() -> &'static str {
+        "~/.clarinet/clarinetrc.toml"
+    }
+
+    pub fn from_global_file() -> Self {
+        let home_dir = dirs::home_dir();
+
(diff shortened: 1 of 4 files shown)


hugocaillard:
@tippenein Are you on the Stacks discord?

components/clarinet-cli/src/frontend/clarinetrc.rs
```diff
@@ -0,0 +1,53 @@
...
+
+impl GlobalSettings {
+    pub fn get_settings_file_path() -> &'static str {
+        "~/.clarinet/clarinetrc.toml"
+    }
+
+    pub fn from_global_file() -> Self {
+        let home_dir = dirs::home_dir();
+
+        if let Some(path) = home_dir.map(|home_dir| home_dir.join(".clarinet/clarinetrc.toml")) {
```
lgalabru, line 19: Note for future 2.0 release manager: could be problematic on windows machines.

lgalabru (approved):
LGTM, thanks @hugocaillard!

components/clarinet-cli/src/frontend/clarinetrc.rs
hugocaillard: I haven't tested myself but apparently `PathBuf::join` handles the separator correctly on any OS.
I'll leave it as it is

tippenein:
> @tippenein Are you on the Stacks discord?

I am now!

Thanks for the refactor, very nice 👌

PR #1241, opened 2023-11-09
ci: add coverage

hugocaillard:

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,11 @@
+coverage:
+  status:
+    patch: false
+    project:
+      default:
+        # threshold: 0.005  # 0.5 %
+        threshold: 15%
+        if_not_found: failure
+
+github_checks:
+  annotations: false
--- a/.github/workflows/ci-clarinet-cli.yaml
+++ b/.github/workflows/ci-clarinet-cli.yaml
@@ -118,20 +118,12 @@ jobs:
       - name: Generate coverage
         run: grcov . --binary-path target/debug/ -s . -t lcov --branch --ignore-not-existing --ignore "/*" -o lcov.info
 
-      # TODO(@hugocaillard): npm test
-      # Run functional tests here in addition to the other jobs so we can fail fast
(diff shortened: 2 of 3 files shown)


codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        # threshold: 0.005  # 0.5 %
+        threshold: 0.10 # 10 %
```
hugocaillard, line 6: huge threshold for now. Just to allow merging this PR without bypassing protection. Will decrease later

hugocaillard:
@CharlieC3 
I added back the the codecov report on the "old" ci.
Unfortunately, we are still not using the new, faster version (which should be even faster now that we removed Deno).

But now that this repo is in a more stable state (post clarinet v2 release), it'll make more sense to work on a clean CI and workflow

codecov.yml
```diff
@@ -0,0 +1,7 @@
+coverage:
+  status:
+    project:
+      default:
+        # threshold: 0.005  # 0.5 %
+        threshold: 0.15 # 15 %
```
hugocaillard, line 6: huge threshold for now. Just to allow merging this PR without bypassing protection. Will decrease later

codecov.yml
```diff
@@ -0,0 +1,11 @@
+coverage:
+  status:
+    patch: false
+    project:
+      default:
+        # threshold: 0.005  # 0.5 %
+        threshold: 15%
```
hugocaillard, line 7: Huge threshold for now. Just to allow merging this PR without bypassing protection. Will decrease later

CharlieC3 (approved):
Seems fine to me! :shipit:  I don't necessarily like that we have to duplicate more workflow logic. I'm not suggesting we do anything about it right now, but eventually I would like this project to migrate to the "component" workflows.

hugocaillard:
Thanks @CharlieC3 
Indeed, switching to the "new" CI should be a Q4 focus

Issue #1101, opened 2023-07-25
Create `clarinet devnet ...` commands


MicaiahReid:
I'll attempt to summarize my discussion with @lgalabru on a vision for expanding `clarinet integrate` into a few new commands. Here are the main efforts this will entail:
 - [ ] Add `clarinet devnet start` command. This be the same as `clarinet integrate`
 - [ ] [Breaking] Remove `clarinet integrate` command
 - [ ] Add the `clarinet devnet package` command
 - [ ] Add the `clarinet devnet start --package=Package.json` startup option

---

## Add `clarinet devnet start` command / Remove `clarinet integrate`
The `clarinet integrate` command isn't really self-explanatory in terms of what it actually does, and it doesn't fit the usual pattern for clarinet commands: `clarinet noun verb`. Changing to `clarinet devnet start` fits this pattern and clearly indicates that a devnet will be started.

## Add the `clarinet devnet package` command
This idea originally came from #1059. This command will essentially package up all of the pieces of a project that are required to start a devnet and put them in one JSON file. This will make it easy for the platform team to extract the relevant data from a user's project to hand off to the devnet API. It will also make the next option possible.

_Note: We also were considering `clarinet devnet export` rather than `package`, please comment if you have any opinions on the command name_

##  Add the `clarinet devnet start --package=Package.json` startup option
This new `--package` flag will allow a user to start a devnet using only a clarinet project package (created from `clarinet devnet package`). The user can then start a devnet without having to run `clarinet integrate` from a specific directory - they just need the package file, and it has everything they need.

PR #1244, opened 2023-11-11
feat: devnet start 

tippenein:
#1101

- [x] moves `integrate` to `devnet start` but keep integrate in place for now
- [x] adds the --package flag to specify the direct contents of `devnet package`
- [x] documentation?
- [x] add warn message to `integrate` usage


components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -160,6 +160,10 @@ enum Devnet {
     /// Generate package of all required devnet artifacts
     #[clap(name = "package", bin_name = "package")]
     Package(DevnetPackage),
+
+    /// Start a local Devnet network for interacting with your contracts from your browser
+    #[clap(name = "start", bin_name = "start")]
+    DevnetStart(DevnetStart),
```
hugocaillard, line 166: To be moved around L98
Just to keep both version close to each other

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -422,6 +426,9 @@ struct Integrate {
         conflicts_with = "use-on-disk-deployment-plan"
     )]
     pub use_computed_deployment_plan: bool,
+    /// Path to Package.json
+    #[clap(long = "package")]
+    pub package: Option<String>,
```
hugocaillard, line 431: I think I'm missing some context here.
What is the package.json used for?

hugocaillard (commented):
Thanks @tippenein 
A couple of questions/remarks

components/clarinet-cli/src/frontend/cli.rs
tippenein: last bullet from #1101 

> This new --package flag will allow a user to start a devnet using only a clarinet project package (created from clarinet devnet package). The user can then start a devnet without having to run clarinet integrate from a specific directory - they just need the package file, and it has everything they need.

hugocaillard: ah yes ofc. My bad, I was thinking about a node.js `package.json` and got confused. Sounds good then, thx

hugocaillard:
Hey @tippenein 
Since you still have the "testing" task unchecked in the description, I was wondering if this is ready for review or if you'll still work on this PR?

tippenein:
It's up to you @hugocaillard. I don't have a lot of context on how to test this properly, but I could give it a shot if CodeCov needs to be appeased

hugocaillard:
@tippenein I've been working of fixing the CI and coverage report this week. Most of the coverage reports were kind of incorrect so we don't care too much about that for now. But seen we will.
If you think the PR's ready, I'll review it today or early next week 👍

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -422,6 +426,9 @@ struct Integrate {
         conflicts_with = "use-on-disk-deployment-plan"
     )]
     pub use_computed_deployment_plan: bool,
+    /// Path to Package.json
+    #[clap(long = "package")]
```
hugocaillard, line 430: It should probably conflict with the two others properties right (it shouldn't be possible to specify a package as well as manifest_path or use_computed_deployment_plan)?

hugocaillard (changes requested):
@tippenein 
Almost lgtm, just a nitpick
I'll try to get this merged soon. i also have an upocuming PR that upgrade clap to v4.x so I'd like to merge this on first

components/clarinet-cli/src/frontend/cli.rs
tippenein: True! nice catch

tippenein: I wonder too about the description. Could be better.. Any idea?

hugocaillard (approved):
LGTM!
Thanks @tippenein

PR #1258, opened 2023-11-16
Chore/test with nextest

hugocaillard:
### Description

- Add cargo aliases for test and coverage
- Try switching to `cargo-nextest` instead of the native test. It is actually faster and pleasant to use. It's also used on the stacks-core repo. Nothing prevent us for switching back to the normal `cargo test` in the future (and hoping it takes some inspiration from nextest)
- Adjust the caching strategy which was leading to many issues (all the `target/` shouldn't be cached in the case of a monorepo, plus, the key strategy was broken)
- Delete the uneeded `stacks-devnet-js/cargo.lock`. We only need to lockfile at the root

---

- [ ] still need to port the change to the "new" ci ("clarinet-ci.yaml")
- [ ] review the cahcing strategy for other actions


CharlieC3:
@hugocaillard Looks like this is still being drafted. Are you looking for reviews anyways? Or should I wait until this is out of draft mode?

hugocaillard:
Thanks for checking in @CharlieC3 
I just pushed a couple more commits and marked the PR as ready for review 👍 
cc @aravindgee

.github/workflows/ci-clarinet-cli.yaml
```diff
@@ -85,46 +85,33 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
 
       - name: Install Rust toolchain
         run: |
-          rustup set profile minimal
-          rustup toolchain install stable --component llvm-tools-preview
```
CharlieC3, line 93: I'm guessing `llvm-tools-preview` is not needed anymore? Was this ever needed?

.github/workflows/ci-clarinet-cli.yaml
```diff
@@ -85,46 +85,33 @@ jobs:
...
-      - name: Build & Unit Test
-        env:
-          RUSTFLAGS: "-C instrument-coverage"
-          LLVM_PROFILE_FILE: "${{ env.COMPONENT }}-%p-%m.profraw"
-        run: cargo test --workspace --locked --exclude clarinet-sdk-wasm --exclude clarity-jupyter-kernel
+            ~/.cargo/
+            ./target/debug/build/
+          key: ${{ runner.os }}-rust-${{ env.RUST_VERSION_HASH }}-${{ hashFiles('./Cargo.lock') }}
+
+      - name: install dependencies
```
CharlieC3, line 103: [Previously we were only installing dependencies](https://github.com/hirosystems/clarinet/pull/1258/files#diff-ed5b3be8788f14bb3472c8fc10151fe96e556a8b5f3af2cf1458171da1b1dc7bL109) if there was a cache miss. Are we no longer doing this because we're not caching the entire `target` dir anymore?

CharlieC3 (approved):
+1 to using nextest, I've also seen great things come from using it in the stacks-core repo. LGTM! Just left a couple questions.
:shipit:

.github/workflows/ci-clarinet-cli.yaml
hugocaillard: It might have been in the past, I don't know if `cargo test` / `grcov still` needs it, but it's not needed with `nextest`

.github/workflows/ci-clarinet-cli.yaml
hugocaillard: I think that this kind of dependencies are install in ~/.cargo.
But can I use `if` with `uses`?

It only takes 3sec right now, seems like the action is quite optimized already

CharlieC3: Yup, you can use `if` with any step or job! And agreed it seems fast enough to not need caching for this step.

CharlieC3 (approved):

PR #1266, opened 2023-11-29
chore: upgrade clarity-vm including epoch 3.0

hugocaillard:
### Description

- the `clarity-vm` dependency is now loaded from github instead of crates.io, it will remain the case while this is in active development
- handle epochs 2.5 and 3.0
- update some imports, especially from stacks-common
- new codec wire format

This PR also upgrades some important dependencies:
- tokio
- reqwest
- bollard
- tui -> ratatui

PR #1282, opened 2023-12-07
feat: include epochs 2.5 and 3.0 in generated Stacks.toml

obycode:
This adds support to include the start height for epochs 2.5 and 3.0 in the Stacks.toml generated for devnet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5089,20 +5089,20 @@ dependencies = [
 
 [[package]]
 name = "tracing-log"
-version = "0.1.3"
+version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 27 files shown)


obycode:
I fixed some issues in https://github.com/stacks-network/stacks-core/pull/4138.

Next step is to add some pox-4 stackers so we can start epoch 3.0!

lgalabru:
Thanks @obycode! It'd be great to ignore the 101 first blocks that I was experiencing earlier. 
It should be a one liner stacks-node but it saves a bunch of seconds of initialization time for developers (very valuable when you're in a manual, caveman debugging cycle test / retry / test / retry / ...)

hugocaillard:
@obycode I cherry-picked the commits in #1266

obycode:
Great! So we can close this one?

hugocaillard:
Yes 👍

PR #1284, opened 2023-12-11
fix(sdk): tuple diff

hugocaillard:
### Description

Because the `actual` and `expected` values are formatted with `prettyPrint`, they are just strings that vitest uses to display a diff. The properties need to be sorted so that the diff can properly be computed.

This is a screenshot of a diff _before_ this fix

Screenshot 2023-12-11 at 12 00 35

### Additional changes

- Upgrade dependencies
- export more types (requested on Discord)


janniks (approved):
👏

PR #1297, opened 2023-12-20
fix: adjust the timing of the stacking orders

obycode:
If `stack-stx` is processed during a prepare phase, then the account is not stacked for the next cycle, N, it is stacked for N+1. To deal with that, this change moves the stacking orders earlier in the cycle so that they are processed before the prepare phase.


hugocaillard (approved):
LGTM 🔥

PR #1299, opened 2023-12-23
fix: set the burnchain mode for pre-nakamoto correctly

obycode:


hugocaillard (approved):
LGTM

Thanks @obycode 
Good catch!

Issue #1305, opened 2024-01-11
Optimize bitcoind 0.25 image


hugocaillard:
Build an optimized version of our own bitcoin image > 25.0

PR #1315, opened 2024-01-19
update bitcoin image to 26.0

csgui:
This PR updates Bitcoin image to 26.0.

~It also removes the wallet support that seems needless for Clarinet use case.~

Fixes: #1305


hugocaillard (approved):
LGTM! thx @csgui

PR #1325, opened 2024-01-23
chore: upgrade clarity-vm - DO NOT MERGE

hugocaillard:
### Description

Reflect the changes made on the stacks blockchain: https://github.com/stacks-network/stacks-core/commit/12f5d9c6c3338a7ef74c3133e30d0d958273b723

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -713,7 +713,7 @@ checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 [[package]]
 name = "chainhook-sdk"
 version = "0.12.1"
-source = "git+https://github.com/hirosystems/chainhook.git?rev=54f80164c764180344cdbf25ee3847a310859132#54f80164c764180344cdbf25ee3847a310859132"
+source = "git+https://github.com/hirosystems/chainhook.git?rev=2a47192#2a4719238ded642d6961885e5b2c5ae48c727d56"
 dependencies = [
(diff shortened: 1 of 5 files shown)


hstove:
One thing with this branch is that when running `clarinet check` you get really long output, even when it succeeds. Starting with:

```
wasm error: [Diagnostic { level: Error, message: "Wasm Runtime Error: Runtime error while interpreting ST3NBRSFKX28FQ2ZJ1MAKX58HKHSDGNV5N7R21XCP.delegate-sig: Wasm(UnableToLoadModule(WebAssembly translation error\n\nCaused by:\n    Invalid input WebAssembly code at offset 5980: type mismatch: expected i64, found i32))", spans: [], suggestion: None }]
result: ExecutionResult { result: Contract(ContractEvaluationResult { result: None, contract: ParsedContract
```

I see this when running in different projects, and I don't think it's related to a specific contract.

hugocaillard:
@hstove 
So this branch include the changes to handle the new wire format of the blockchain, but is probably super late on clarity-wasm.


> you get really long output, even when it succeeds

👆 This is still a valid comment.
Clarinet tries to run clarity-wasm and the interpreter in parallel. It will compare the two result but will always the one from the interpreter.
That explains why you can see a big error from clarity-wasm but still a success from the interpreter. In a recent PR (that is not included on this branch), we increase the messaging https://github.com/hirosystems/clarinet/pull/1330

hugocaillard:
Closing in favor of #1372

PR #1330, opened 2024-01-24
feat: run clarity-wasm in a separate session

hugocaillard:
### Description

Revamp the way we run the clarity-wasm preview in `clarinet console` . We now run two session in parallel, the second one run in "wasm-mode".

- The second session (the wasm one), can be **disabled** with: 
```console
clarinet console --disable-clarity-wasm
```

- If the interpreter and clarity-wasm return different results, the results are logged and a message asks the users to open an issue
Screenshot 2024-01-24 at 20 58 11

- These PR also implements a `::toggle_timings` that give an approximate idea of how long a snippet execution took

### DIFF
--- a/components/clarity-repl/src/frontend/terminal.rs
+++ b/components/clarity-repl/src/frontend/terminal.rs
@@ -1,5 +1,6 @@
 use crate::repl::{settings::SessionSettings, Session};
 
+use clarity::vm::EvaluationResult;
 use rustyline::error::ReadlineError;
 use rustyline::Editor;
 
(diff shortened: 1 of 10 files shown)


components/clarity-repl/src/frontend/terminal.rs
```diff
@@ -73,18 +74,27 @@ fn complete_input(str: &str) -> Result<Input, (char, char)> {
...
+        let session = Session::new(session_settings);
+        let session_wasm = session_wasm_settings.map(Session::new);
+        Terminal {
+            session,
+            session_wasm,
+        }
     }
 
-    pub fn load(mut session: Session) -> Terminal {
-        session.is_interactive = true;
```
hugocaillard, line 86: Wasn't used

MicaiahReid (approved):
LGTM!

PR #1331, opened 2024-01-25
feat: improved stacking orders

### LINKED ISSUE #1302: Validate devnet config on parsing, adapt Stacking logic
**Context**
The devnet config is highly configurable in Devnet.toml.
With Nakamoto, soem settings are quite sensitive and should be carefully validated.
The check is happening [here](https://github.com/stacks-network/stacks-core/blob/f394a3b1d11954117bd66bbbcdf4f42894f255e3/testnet/stacks-node/src/config.rs#L654-L660) on the stacks-node

**Tasks**
- if epoch_3_0 height is overridden, make sure it happens during prepare phase
- validate stacking_orders
  - validate wallet names
  - makes sure there's some stacking happening before epoch 3.0 activation
  - makes sure all stacking don't happen during the same cycle, otherwise, it all stop at the same cycle and it leads to a cycle with no stacking
- detect old devnet.toml config and suggest to overwrite it

**To consider**

Is it important for the stacking orders to be customisable? Your should we hard code some robust, reliable stacking logic?

### LINKED ISSUE #1303: Implement automatic re-stacking in Devnet after the number of cycles has passed
**Context**

Because having stacked STX is important in Nakamoto, we need a robust auto stacking stategy

**Task**

For each stacking account:
```
# Send some stacking orders
[[devnet.pox_stacking_orders]]
start_at_cycle = 0
duration = 12
wallet = "wallet_1"
slots = 2
btc_address = "mr1iPkD9N3RJZZxXRk7xF9d36gffa6exNC"
```

Set up automatic re-stacking at the end of the `duration`

MicaiahReid:
Fixes #1302 
Fixes #1303


components/clarinet-files/src/network_manifest.rs
```diff
@@ -316,6 +323,14 @@ pub struct PoxStackingOrder {
     pub wallet: String,
     pub slots: u64,
     pub btc_address: String,
+    pub auto_extend: Option<bool>,
+    pub cycles_stacked: Option<u32>,
```
hugocaillard, line 327: What is `cycles_stacked` for?

components/stacks-network/src/orchestrator.rs
```diff
@@ -1343,7 +1343,7 @@ observer_port = {subnet_events_ingestion_port}
             stacks_node_p2p_port = devnet_config.stacks_node_p2p_port,
             orchestrator_port = devnet_config.orchestrator_ingestion_port,
             subnet_events_ingestion_port = devnet_config.subnet_events_ingestion_port,
-            first_burn_header_height = 0,
+            first_burn_header_height = 0, // todo: I don't think this should be 0
```
hugocaillard, line 1346: I removed this comment.
- This for the subnet config so I think we can ignore it for now
- I don't find references to this config in the stacks-repo so maybe we'll fully remove it in the future

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -572,10 +573,17 @@ pub async fn publish_stacking_orders(
...
+        let PoxStackingOrder {
+            duration,
+            start_at_cycle,
+            ..
+        } = pox_stacking_order;
+
+        if (reward_cycle_id as u32 % duration) != (start_at_cycle - 1) {
             continue;
         }
+        let extend_stacking = reward_cycle_id as u32 != start_at_cycle - 1;
```
MicaiahReid, line 586: This is waaaaay better than my attempts to see if we've already stacked; I did that in a hurry 😅 

Shouldn't we also check the `auto_extend` flag to see if the user _wants_ us to extend?

hugocaillard: Yes you right! I added a check below

hugocaillard (approved):

PR #1332, opened 2024-01-25
Update Github action format for Clarinet 2.x

whoabuddy:
In the latest version `clarinet test` is deprecated in favor of using the Clarinet SDK for unit testing.

The default steps in the README now will fail because `clarinet test` does not run, see below:

https://github.com/stacks-m2m/contracts/actions/runs/7661218176

By updating the steps to use `clarinet check` first, then install Node + run `npm test`, this version is more in line with the default setup and succeeds:

https://github.com/stacks-m2m/contracts/actions/runs/7661309610

This also updates the checkout and codecov actions to their latest versions.

You can see the file we're using here which is very similar:

https://github.com/stacks-m2m/contracts/blob/main/.github/workflows/clarinet.yml

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,13 +546,22 @@ jobs:
     name: "Test contracts with Clarinet"
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - name: "Execute unit tests"
+      - name: "Checkout code"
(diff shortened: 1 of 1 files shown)


hugocaillard (approved):
Good catch! Thanks @whoabuddy

PR #1350, opened 2024-02-07
Fix/devnet pox and stacking

hugocaillard:
### Description

Add a few fixes and improvement to be released as 2.2.1

- The publish_stacking_order formula wasn't working on some edge case
  - duration = 1
  - start_at_cycle > duration
The new formula is slightly simpler (still a bit complicated but well..) and it works in all cases

- Handle the `reward_cycle_length` property in PoxInfo and revert the regression on `reward_phase_block_length`
  - This fix requires the update on chainhook

- In the devnet UI, avoid multiple `>` in case a nakamoto block is mined at the same block height as the beginning of a new pox cycle (`eg – [12]> – [11]> – <[10] –`)
 
- use a fixed tag for the `hirosystems/stacks-blockchain-api` docker image (because `:nakamoto` is updated frequently)


MicaiahReid (approved):
LGTM!

PR #1362, opened 2024-02-26
Refactor/fix tuple items sorting in tests

### LINKED ISSUE #1354: TypeError: Object.entries(...).toSorted is not a function when using Node.js v18.15
When running the code snippet from [Unit Test for `Counter` Example](https://docs.hiro.so/clarinet/guides/test-contract-with-clarinet-sdk#unit-tests-for-counter-example) on Nodejs v18.15, the below error is thrown:

image

It is thrown from the `expect(printEvent.data.value).toBeTuple()` method from the code snippet. Upon further inspection of the definition in the `clarityValuesMatchers.ts` file, the `toBeTuple()` method (ln 345) is using [`Object.entries().toSorted`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/toSorted) method which is only supported in Nodejs v20.

image

I changed the two `.toSorted` calls to `.sort` methods and the test for "sends a print event" passes.

image
image

**Environment**
 - Nodejs v18.15
 - Clarinet v2.2.1

hugocaillard:
## Description

The clarinet-sdk wasn't properly handling tuples property sorting, leading to complex diff reading in failing tests. (It was only sorting top level tuple properties)
Stacks.js now handles tuples property sorting in pretty print so that Clarinet doesn't have (and it handle nested values out of the box)

Fix #1354

### DIFF
--- a/components/clarinet-sdk/package-lock.json
+++ b/components/clarinet-sdk/package-lock.json
@@ -1,16 +1,16 @@
 {
   "name": "@hirosystems/clarinet-sdk",
-  "version": "1.2.0",
+  "version": "1.3.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@hirosystems/clarinet-sdk",
-      "version": "1.2.0",
+      "version": "1.3.0",
       "license": "GPL-3.0",
       "dependencies": {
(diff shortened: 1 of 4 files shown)


janniks (approved):
LGTM ✨

PR #1363, opened 2024-02-27
fix: log to devnet.log file when running js devnet

### LINKED ISSUE #1361: [devnet] log file is not saved for stacks-devnet-js
**Describe the bug**
When using stacks-devnet-js, a `devnet.log` file is not saved to the user's `working_dir` like it is when using `clarinet devnet start`

MicaiahReid:
### Description

Previously when running `stacks-devnet-js`, a `devnet.log` file was not created as it is in a normal `clarinet devnet start` run. This PR creates that log file and writes devnet events to the file.

Fixes #1361


hugocaillard (approved):
LGTM 👍

PR #1365, opened 2024-02-29
Add test with contracts in parent path

friedger:
### Description

stacks-core uses a setup for Clarinet where the contract code lives in a parent folder of Clarinet.toml. The resulting coverage reports generate unreadable html due to missing css files living in the coverage folder not reachable from the source code folder.

This PR adds a test for this setup.

### Checklist


friedger:
There is no issue with the generated code if genhtml is called from the parent folder.

PR #1372, opened 2024-03-04
chore: update stacks core

hugocaillard:
### Description

- update clarity to the latest feat/clarity-wasm-next branch
  - this implies a lot of `.expect()` that should be pretty safe
- fix a new clippy warning (ptr_arg improvement) 3 occurrences

PR #1375, opened 2024-03-05
chore: handle pox-locking in simnet

hugocaillard:
### Description

- update pox-4.clar
- implement pox-locking
- I wanted to split a PR into 2, but the pox-locking unit tests were pushed to wrong one (already merged on main) -> https://github.com/hirosystems/clarinet/pull/1372/files#diff-a4b371d4141fdb859465960f65188b6a26d18a7fe2473425e4133b2648ffe0d8

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -43,7 +43,7 @@ use types::{ContractPublishSpecification, EpochSpec};
 pub type ExecutionResultMap =
     BTreeMap<QualifiedContractIdentifier, Result<ExecutionResult, Vec<Diagnostic>>>;
 
-pub struct UpdateSessionExecutioResult {
+pub struct UpdateSessionExecutionResult {
     pub boot_contracts: ExecutionResultMap,
     pub contracts: ExecutionResultMap,
 }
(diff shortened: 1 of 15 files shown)


components/clarinet-deployments/src/lib.rs
```diff
@@ -125,14 +125,14 @@ pub fn update_session_with_contracts_executions(
 ) -> BTreeMap<QualifiedContractIdentifier, Result<ExecutionResult, Vec<Diagnostic>>> {
     let boot_contracts_data = BOOT_CONTRACTS_DATA.clone();
 
-    for (_, (boot_contract, ast)) in boot_contracts_data {
-        session
+    let mut results = BTreeMap::new();
```
MicaiahReid, line 128: When I originally made this change, I wondered if this has any side effects that I hadn't considered. Can you think of any ways that adding these boot contracts to the contract executions could be problematic @hugocaillard?

MicaiahReid (approved):

components/clarinet-deployments/src/lib.rs
hugocaillard: @MicaiahReid
To avoid unwanted side effects (such has having diagnostics from the boot contracts or even a bad impact on the performances), I split the `update_session_with_contracts_executions` result to separate boot_contracts result from the project contracts

MicaiahReid (approved):

Issue #1379, opened 2024-03-11
Update stacks-node and add stacks-signer


hugocaillard:
- update stacks-node image
- update stack-stx (new arguments in pox-4 for stack-stx and extend)
- add the signer image to the devnet

sabbyanandan:
@hugocaillard: It is probably too early right now, but when the time is right, it'd be good to get a sense of the Docker daemon's compute requirements (i.e., what is a reasonable CPU, disk, and memory defaults?), including the additional resources (for instance, a new "signer image") that need to be spun up. We can add/update the pre-requisite section in docs/README. 

In the Platform, we will automatically apply and handle the dynamic allocation of new resource requirements for each user.

PR #1384, opened 2024-03-20
Feat/upgrade devnet

hugocaillard:
### Description
Fixes: #1379 

wip


components/stacks-network/src/orchestrator.rs
```diff
@@ -1263,6 +1406,180 @@ start_height = {epoch_3_0}
...
+            "7287ba251d44a4d3fd9276c88ce34c5c52a038955511cccaf77e61068649c17801",
+            "530d9f61984c888536871c6573073bdfc0058896dc1adfe9a6a10dfacadc209101",
+        ];
+
+        let signer_conf = format!(
+            r#"
+stacks_private_key = "{signer_private_key}"
+node_host = "stacks-node.{network_name}:{stacks_node_rpc_port}" # eg "127.0.0.1:20443"
+# must be added as event_observer in node config:
+endpoint = "0.0.0.0:30000" # e.g 127.0.0.1:30000
```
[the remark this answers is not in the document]
hugocaillard: Not necessary but done, to have the same behavior as the regtest env

components/clarinet-files/src/network_manifest.rs
```diff
@@ -893,6 +896,16 @@ impl NetworkManifest {
                         }
                     },
                 ),
+                stacks_signer_image_url: devnet_config
+                    .stacks_signer_image_url
+                    .take()
+                    .unwrap_or_else(|| {
+                        if devnet_config.use_nakamoto.unwrap_or(false) {
+                            DEFAULT_STACKS_NODE_IMAGE_NAKA.to_string()
```
MicaiahReid, line 904: Currently this won't work with the default naka image, right? Do we need to update this const to be beta2?

components/stacks-network/src/orchestrator.rs
```diff
@@ -559,14 +597,108 @@ impl DevnetOrchestrator {
             }
         };
 
+        // Start stacks-signer-1
```
MicaiahReid, line 600: Do we want booting the signers to be wrapped in `if use_nakamoto`?

MicaiahReid (approved):

MicaiahReid:
My main concern is - do we want to disable the signer if `use_nakamoto = false`? Has this been tested with nakamoto disabled?

hugocaillard:
@MicaiahReid
> My main concern is - do we want to disable the signer if use_nakamoto = false? Has this been tested with nakamoto disabled?

Ah yes, I might have forgot to do that 🤔 
Will do.

Testing with nakamoto disabled is part of the release playbook so I was about to 👍

components/stacks-network/src/orchestrator.rs
hugocaillard: Good catch

components/clarinet-files/src/network_manifest.rs
hugocaillard: Done! Up to beta4

MicaiahReid (approved):

PR #1394, opened 2024-03-27
Allow requirements for contracts deployed with multisig

friedger:
### Description

Calling `clarinet check` fails when the Clarinet.toml file includes requirements with contracts deployed from a multisg.

### Example

https://github.com/lisalab-io/liquid-stacking/pull/39
---

### Checklist


hugocaillard:
Good catch! Thanks for the fix @friedger 
I'll release today as Clarinet 2.4.1

hugocaillard (approved):
LGTM! Thanks @friedger!

Issue #1393, opened 2024-03-26
Epoch 3.0 handling in Devnet


hugocaillard:
With Clarinet 2.8.0, the Devnet supports Epoch 2.5 and 3.0.
3.0 is not enabled by default, this is normal, we want Clarinet to be as close as possible to Mainnet by default.

It's unclear why we fail to mine Epoch 3.0 blocks. It's likely to be a timing issue:
- pox cycles could be too fast (10 blocks – prepare phase is 4 block), let's try 20-5 as in the regtest evn
- bitcoin blocks too fast? By default a bitcoin block is mined every 30sec bit it can be configured to faster blocks. Also, while the devnet is applying a deployment plan, it mines bitcoin blocks every 1sec, which can happen in epoch 2.5 if a contract is set to be deployed in 3.0. Is it too fast for the DKG process to happen in Epoch 2.5? Let's see how fast can bitcoin blocks be until it breaks
- other time issue could simply be cause be the block height at which the stackers stack or by the configured epoch 3.0 height. Will have to try multiple values

PR #1397, opened 2024-03-27
Handle epoch 3.0

hugocaillard:
### Description

Fixes: #1393 

- fix stacks-node config and update the wire format
- adjust stacking orders and cycles timings so that we can reach epoch 3.0 quickly


hugocaillard:
Those changes were squashed and included and #1416 1416

PR #1409, opened 2024-04-09
chore: update pox-4.clar

hugocaillard:
### Description

Update pox-4.clar with the latest changes.
Update signers.clar and signer-voting.clar so that they are exactly the same as they are in stacks-core (with the extra spaces)


MicaiahReid (approved):
Incredible work. 😜

PR #1412, opened 2024-04-09
feat: enable boot contracts coverage

hugocaillard:
### Description

In some context (Stacks core developers or others power users), Clarinet can be used to unit tests boot contracts (bns, pox, signers, etc).
In this cases, Clarinet should be able to report code coverage for such contract. This is what this PR allow. As long as the user provides a path to the copy of the boot contracts that are being tested. 

See it in action here: https://github.com/hugocaillard/pox-testing-sample/pull/3


components/clarinet-sdk/tests/pox-locking.test.ts
```diff
@@ -147,12 +166,7 @@ describe("test pox-4", () => {
       btcAddr: publicKeyToBtcAddress(pubKey),
       signerPrivKey: signerPrivKey,
       signerPubKey: signerPubKey,
-      client: new StackingClient(stxAddress, {
-        ...network,
-        transactionVersion: network.version,
```
hugocaillard, line 152: This was greatly simplified in stacks.js @ 6.13.x

obycode (approved):
LGTM!

PR #1416, opened 2024-04-11
feat: devnet in 2 5 by default

hugocaillard:


components/clarity-repl/src/repl/mod.rs
```diff
@@ -25,7 +25,7 @@ use clarity::types::StacksEpochId;
 use clarity::vm::ClarityVersion;
 
 pub const DEFAULT_CLARITY_VERSION: ClarityVersion = ClarityVersion::Clarity2;
-pub const DEFAULT_EPOCH: StacksEpochId = StacksEpochId::Epoch24;
+pub const DEFAULT_EPOCH: StacksEpochId = StacksEpochId::Epoch25;
```
hugocaillard, line 28: MB, I won't include that.
Will do another upgrade once we know the epoch 2.5 height on mainnet, and also change this line

hugocaillard:
Superseded by #1422

PR #1417, opened 2024-04-12
fix: improve error message in sdk custom matchers

hugocaillard:
### Description

When using the sdk with vitest, if the custom matcher expects the wrong types, we keep error messages short
(eg: `expect(Cl.uint(1)).toBeOk(Cl.uint(1))` will throw `actual value must be a Clarity "UInt", received "ResponseOk"`)

This should be improved to display a more detailed error diff, but it requires some more work so that the error message remain readable, even when the test is handling big set of data (such as big lists or buffers).

This PR simply handle a quick fix for the common case were `expectOk` actually receive a ResponseError. We should display the received error

### DIFF
--- a/components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
+++ b/components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
@@ -120,7 +120,7 @@ function checkCVType<T extends ClarityType>(
     // for readability, the error diff is kept short if the developers uses the wrong `expect<ClarityType>`
     // ideally, we should have a way to display short message diffs even if the actual and/or expected data are big lists/tuples/buffers
 
-    // for now, we make and exception and display the full error message if the actual value is a ResponseErr
+    // for now, we make an exception and display the full error message if the actual value is a ResponseErr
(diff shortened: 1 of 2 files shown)


hugocaillard:
cc @friedger

components/clarinet-sdk/vitest-helpers/src/clarityValuesMatchers.ts
```diff
@@ -117,10 +117,17 @@ function checkCVType<T extends ClarityType>(
   const isCVWithType = isClarityValueWithType(actual, expectedType);
 
   if (!isCVWithType) {
+    // for readability, the error diff is kept short if the developers uses the wrong `expect<ClarityType>`
+    // ideally, we should have a way to display short message diffs even if the actual and/or expected data are big lists/tuples/buffers
+
+    // for now, we make and exception and display the full error message if the actual value is a ResponseErr
```
MicaiahReid, line 123:
```suggestion
    // for now, we make an exception and display the full error message if the actual value is a ResponseErr
```

MicaiahReid (changes requested):

MicaiahReid (approved):

PR #1420, opened 2024-04-15
ci: fix homebrew release

hugocaillard:
### Description

The release actions failed on the homebrew step for clarinet 2.5.0:
https://github.com/hirosystems/clarinet/actions/runs/8690598712
With the error:
```
Invalid usage: clarinet: no `--url` or `--version` argument specified!
```

I was able [to open the homebrew PR](https://github.com/Homebrew/homebrew-core/pull/169081) by running this command locally:
```console
brew bump-formula-pr \
  --no-browse \
  --no-audit \
  --force \
  --tag 22e10c2f004ebdc3d07cd475844eaa4d11ebbb75 \
  --version 2.5.0 \
  clarinet
```

Looks like there was a breaking change in the bump-formula-pr command

### DIFF
--- a/.github/workflows/pkg-version-bump.yaml
+++ b/.github/workflows/pkg-version-bump.yaml
@@ -27,6 +27,10 @@ jobs:
           HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.GH_TOKEN }}
           TAG: ${{ github.event.client_payload.tag || github.event.inputs.tag }}
         run: |
+          # Get version info
+          $VERSION=${env:TAG}.substring(1)
+
+          # Configure git configs
           git config --global user.name "${GIT_USER_NAME}"
           git config --global user.email "${GIT_USER_EMAIL}"
 
@@ -36,6 +40,7 @@ jobs:
             --no-audit \
             --force \
             --tag "${TAG}" \
+            --version ${VERSION} \
             ${{ github.event.repository.name }}
 
   winget:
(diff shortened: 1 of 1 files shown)


.github/workflows/pkg-version-bump.yaml
```diff
@@ -27,6 +27,10 @@ jobs:
           HOMEBREW_GITHUB_API_TOKEN: ${{ secrets.GH_TOKEN }}
           TAG: ${{ github.event.client_payload.tag || github.event.inputs.tag }}
         run: |
+          # Get version info
+          $VERSION=${env:TAG}.substring(1)
```
CharlieC3, line 31:
```suggestion
          VERSION=$(echo "${ENV#v}")
```

CharlieC3 (changes requested):
Thanks for starting this change! I think the homebrew job will need to be different than winget because it's using bash instead of powershell.

hugocaillard:
Thanks @CharlieC3 
I knew a copy/paste wouldn't be enough.

That part is good right? `--version ${VERSION} \` (we don't need quotes `"--version "${VERSION}" \`?)

CharlieC3:
> That part is good right? --version ${VERSION} \ (we don't need quotes "--version "${VERSION}" \?)

Hmm may be best to add the quotes to match the other arg just in case.

hugocaillard:
Alright, done @CharlieC3 
Thx!

CharlieC3 (approved):
:shipit:

hugocaillard:
thx @CharlieC3

PR #1437, opened 2024-04-24
fix: improved sdk deployment plan handling and perf improvements

### LINKED ISSUE #1435: Handle contract deletion in simnet deployment plan
## Description

The simnet tries to check if the deployment plan needs to be updated before doing so.
It'll look at existing contract deployments and other custom transactions, update the contracts deployment following the Clarinet.toml spec and keep the custom transactions.

It currently fails to handle moved/renamed/deleted contracts

hugocaillard:
### Description

Fix #1435

This PR started as a bug fix, but as I worked on it, I found that I could get massive perf improvements.

In the SDK `init_session`, some data is cached. Before this PR, just de deployment was cached.
But the **whole session** can actually be cached. It reduces file system access by a lot (in a Vitest context, we now only need to read the manifest and contract once) and reduce the time spent parsing contracts.

On my machine (M1), the cached `init_session` go from more than 100ms to a few ms (the time taken for `cache.clone()`). And knowing that `init_session` is called before each test, it takes the time to run tests in a project like pox-4 testing down from 32sec to 6secs

It also reworks a bit the session and the SDK structs

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -55,6 +55,7 @@ pub fn setup_session_with_deployment(
 ) -> DeploymentGenerationArtifacts {
     let mut session = initiate_session_from_deployment(manifest);
     update_session_with_genesis_accounts(&mut session, deployment);
+    // session.load_boot_contracts();
     let UpdateSessionExecutionResult { contracts, .. } = update_session_with_contracts_executions(
         &mut session,
         deployment,
@@ -120,7 +121,6 @@ pub fn update_session_with_genesis_accounts(
                 session.set_tx_sender(wallet.address.to_address());
             }
         }
-        session.load_boot_contracts();
     }
 }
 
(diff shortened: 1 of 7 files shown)


components/clarinet-deployments/src/lib.rs
```diff
@@ -120,7 +121,6 @@ pub fn update_session_with_genesis_accounts(
                 session.set_tx_sender(wallet.address.to_address());
             }
         }
-        session.load_boot_contracts();
```
hugocaillard, line 123: `update_session_with_genesis_accounts()` should not be responsible for calling laod_boot_contracts.

It doesn't have any impacts to remove it because every `update_session_with_genesis_accounts` calls are also followed by `update_session_with_contracts_executions` which also loads boot contracts

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -314,106 +343,54 @@ impl SDK {
...
+            .await?
+        {
+            let spec_file_content = self
+                .file_accessor
+                .read_file(deployment_plan_location.to_string())
+                .await?;
+
+            let mut spec_file = DeploymentSpecificationFile::from_file_content(&spec_file_content)?;
+
+            // the contract publish txs are management by the manifest
```
MicaiahReid, line 365:
```suggestion
            // the contract publish txs are managed by the manifest
```

MicaiahReid (approved):

PR #1441, opened 2024-04-26
fix: add options to enable or disable costs and coverage reports in sdk

hugocaillard:
### Description

To be rebased on main and merged after #1437 

The sdk was always tracking costs and coverage reports even when it wasn't enabled.
By only enabling it when needed, we get better performance

### DIFF
--- a/components/clarinet-sdk/tests/reports.test.ts
+++ b/components/clarinet-sdk/tests/reports.test.ts
@@ -1,0 +1,109 @@
+import fs from "node:fs";
+import path from "node:path";
+import { describe, expect, it, beforeEach, afterEach } from "vitest";
+
+// test the built package and not the source code
+// makes it simpler to handle wasm build
+import { Simnet, initSimnet } from "../dist/esm";
+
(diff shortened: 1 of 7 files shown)


components/clarinet-sdk/tests/reports.test.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+}
+
+beforeEach(async () => {
+  deleteExistingDeploymentPlan();
+});
+
+afterEach(() => {
+  deleteExistingDeploymentPlan();
+});
```
MicaiahReid, line 28: Do we need to do this before and after?

MicaiahReid (approved):

components/clarinet-sdk/tests/reports.test.ts
hugocaillard: `beforeEach` is definitely required to make sure the first test is properly ran.
The  `afterEach` _could_ be replaced by an `afterAll` just to clean the project after the test ran (or we could not clean it at all

MicaiahReid: 👍

PR #1442, opened 2024-04-26
fix: pox-4

hugocaillard:
### Description

This change is the right version. This is pox-4 as deployed on chain


obycode (approved):
Thanks!

PR #1444, opened 2024-04-29
Upgrade rustyline

### LINKED ISSUE #616: Escaped string-utf8 value causes the console to get stuck
Here are 2 examples;

```
~ clarinet console
note: no manifest found. Continuing with default.
clarity-repl v0.33.0
Enter "::help" for usage hints.
Connected to a transient in-memory database.
```
```clarity
>> (define-data-var variable (string-utf8 50) u"And this is an UTF-8 string \u{1f601}")
(.. 
(.. 
(.. 
(.. 
```

Another example

```clarity
>> (len u"And this is an UTF-8 string \u{1f601}")
(.. 
(.. 
(.. 

```

### LINKED ISSUE #1271: Console line ending detection error
**Describe the bug**
If you try to enter the expression `{foo: "\\"}` into the console, it thinks that the expression is not yet complete, and prompts you for more input. This is a valid Clarity expression, it's just the logic detecting whether to wait for more input that has an issue with it.

**To Reproduce**
1. Run `clarinet console`
2. Type `{foo: "\\"}` at the prompt
3. See problem `{.. ` indicating that it is waiting for the rest of the expression

**Expected behavior**
This should correctly be processed as a valid tuple.

### LINKED ISSUE #1323: UTF-8 with Unicode isn't working in complex types (tuple, lists etc)
Unicode characters, like `\u{ff}`, don't work in complex types, like lists and tuples. For instance this would work fine in clarinet console

```
(list u"ff")
```

but as soon as you put unicode character, it won't work. Following example won't return any result

```
(list u"\u{ff}")
```

**To Reproduce**
1. Open terminal
2. Run `clarinet console`
3. Execute `(list u"\u{ff}")`

**Expected behavior**
Should return result (or panic if input is not correct).

You can see the following image

image

hugocaillard:
### Description

- upgrade rustyline dependency
- fix `skip_next` behavior when parsing user input in clarinet console

Fix #1323
Fix #1271
Fix #616

### DIFF
--- a/components/clarity-repl/src/frontend/terminal.rs
+++ b/components/clarity-repl/src/frontend/terminal.rs
@@ -22,22 +22,19 @@ fn complete_input(str: &str) -> Result<Input, (char, char)> {
     let mut in_string = false;
 
     for (pos, character) in str.char_indices() {
-        let skip_current = skip_next;
+        // if the previous character was a backslash, skip this character (only in strings)
         if skip_next {
             skip_next = false;
+            continue;
         }
 
         match character {
             '\\' => {
-                if !skip_current {
-                    skip_next = true
+                if in_string {
+                    skip_next = true;
                 }
             }
-            '"' => {
-                if !skip_current {
-                    in_string = !in_string
-                }
-            }
+            '"' => in_string = !in_string,
             '(' | '{' => {
                 if !in_string {
                     brackets.push(character);
--- a/components/clarity-repl/src/frontend/terminal.rs
+++ b/components/clarity-repl/src/frontend/terminal.rs
@@ -2,7 +2,7 @@ use crate::repl::{settings::SessionSettings, Session};
 
 use clarity::vm::EvaluationResult;
 use rustyline::error::ReadlineError;
-use rustyline::Editor;
+use rustyline::DefaultEditor;
 
 const VERSION: Option<&'static str> = option_env!("CARGO_PKG_VERSION");
 const HISTORY_FILE: Option<&'static str> = option_env!("CLARITY_REPL_HISTORY_FILE");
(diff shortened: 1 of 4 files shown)


components/clarity-repl/src/frontend/terminal.rs
```diff
@@ -22,12 +22,19 @@ fn complete_input(str: &str) -> Result<Input, (char, char)> {
     let mut in_string = false;
 
     for (pos, character) in str.char_indices() {
+        let skip_current = skip_next;
```
MicaiahReid, line 25: I think some comments explaining this logic will be helpful for future Hugo!

components/clarity-repl/src/frontend/terminal.rs
```diff
@@ -230,3 +237,32 @@ pub fn print_clarity_wasm_warning() {
...
+mod tests {
+    use super::*;
+    use test_case::test_case;
+
+    #[test_case(r#"(list (list u1)"# ; "incomplete input missing closing parenthesis")]
+    #[test_case(r#"{ a: { b: 1 }"# ; "incomplete input missing closing curly brace")]
+    #[test_case(r#"(concat u")""# ; "incomplete input with closing parenthesis in string")]
+    #[test_case(r#"{ a: { b: "}" }"# ; "incomplete input with closing curly brace in strng")]
+    #[test_case(r#"{foo: "\"""# ; "incomplete input with escaped quote in string")]
+    fn test_incomplete_input(input: &str) {
```
MicaiahReid, line 251: 🔥

MicaiahReid (approved):

MicaiahReid (approved):

PR #1481, opened 2024-06-21
Fix/better new project name handling

### LINKED ISSUE #1480: Creating a new Clarinet project in the local folder breaks devnet
**Describe the bug**
When you use `clarinet new` to create a project in the current folder, it will break devnet. The reason is that it names the project `'.'` which causes the Docker network names to look like this: `stacks-node...devnet`. The resulting tool looks like this:

```toml
[project]
name = '.'
```

**To Reproduce**
```shell
mkdir broken
cd broken
clarinet new .
clarinet devnet start
```

**Expected behavior**
Everything should still work properly.

**Suggested fix**
When using a path, Clarinet should use the folder name as the project name. In the example above, it should take the name `broken`. Then we can use `clarinet new .`. Additionally, it is probably good to filter out illegal characters in the name for the Docker network names. (If it is not done already.)

hugocaillard:
### Description

Fix #1480 

- Ask users if they want to continue if the directory already exists or already contains files (when using `.`
- Allow to include path in project name `clarinet new path/to/project`
- Sanitize project name (only allow alphanum, `-`, and `_`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -85,7 +85,7 @@ jobs:
         run: cargo install cargo-audit
 
       - name: Run audit
-        run: cargo audit --ignore RUSTSEC-2022-0093 --ignore RUSTSEC-2021-0076 --ignore RUSTSEC-2022-0090 --ignore RUSTSEC-2022-002 --ignore RUSTSEC-2022-0028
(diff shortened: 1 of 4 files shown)


MicaiahReid (approved):
NICE 🎉

PR #1482, opened 2024-06-21
fix: set tenure height on epoch change if great or equal than 3.0

hugocaillard:
### Description

set tenure height on epoch change if great or equal than 3.0


obycode (approved):
LGTM!

PR #1493, opened 2024-07-01
fix: introduce create-only-if-needed functions for containers and images

### LINKED ISSUE #1455: Devnet should work offline if the Docker images exist locally
Currently, the devnet seems to always try to download the Docker images even if a version exist locally.
It should work offline

tippenein:
closes #1455 

The main addition here is `create_image_if_needed` and the container_if_needed is seemingly only needed for postgres. There is a bit of a discrepancy with what to do when we hit already existing containers. Currently it uses it, but alternatively we could drop and recreate for postgres.. 

this current functionality allows me to `clarinet devnet start` with wifi off.

### DIFF
--- a/components/stacks-network/src/orchestrator.rs
+++ b/components/stacks-network/src/orchestrator.rs
@@ -1,12 +1,13 @@
 use bollard::container::{
     Config, CreateContainerOptions, KillContainerOptions, ListContainersOptions,
-    PruneContainersOptions, WaitContainerOptions,
+    PruneContainersOptions, RemoveContainerOptions, WaitContainerOptions,
 };
 use bollard::errors::Error as DockerError;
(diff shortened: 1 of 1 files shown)


hugocaillard:
@tippenein 
Regarding the original issue (#1455), I was wondering if you could make it work with native docker behaviors, either by building images with `--pull false` (`docker build --pull=false <...>` ) or when starting the container `docker run --pull=never <....>`

hugocaillard:
@tippenein If we build our images with `--pull false`, I think it would still try to pull it it doesn't exist.
The documentation says that `--pull: _Always_ attempt to pull all referenced images`, I'm assuming that pull false only make it so docker pulls it only if it needs to

hugocaillard:
Switching to draft for now will it's still wip

PR #1500, opened 2024-07-09
Support workspaces

bradleyayers:
The location of `node_modules` is not guaranteed to be relative to the current directory (e.g. in yarn workspaces). 

A better solution here is to update package.json `exports` to include the file.

### Example

---

### Checklist

### DIFF
--- a/components/clarinet-sdk/src/vitest/index.ts
+++ b/components/clarinet-sdk/src/vitest/index.ts
@@ -1,5 +1,6 @@
 import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
+import { join } from "node:path";
 
 export function getClarinetVitestsArgv() {
   const argv = hideBin(process.argv);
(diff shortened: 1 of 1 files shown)


components/clarinet-sdk/src/vitest/index.ts
```diff
@@ -1,5 +1,6 @@
 import yargs from "yargs";
 import { hideBin } from "yargs/helpers";
+import { join } from "node:path";
```
hugocaillard, line 3: You are not using it right?

hugocaillard:
Good catch, thank you for opening this PR @bradleyayers 
I'll give it a try and need to make sure it doesn't have unwanted side effectsx

bradleyayers:
One thing to check is whether require.resolve is the correct API to use for ESM output, i'm not sure. And also the point about instead of doing a manual path concat i think it's better if the target file is made one of the exports in package.json.

hugocaillard (changes requested):
remove unused imports

hugocaillard:
Hey @bradleyayers 
Sorry we couldn't get this merged before, now there is conflict. The file has been moved, you basically need to migrate the your change in this file:  `components/clarinet-sdk/node/src/vitest/index.ts`.

Do you think you'll have time to do it or should I take over the task (and credit you in my commit)?

bradleyayers:
Please take this over that would be amazing, thank you.

hugocaillard:
Superseded by #1514

PR #1506, opened 2024-07-15
Advance chaintips changes

### LINKED ISSUE #1495: Add advance_stacks_tip and advance_burnchain_tip in simnet
In simnet, the chaintip in simnet with `advance_chain_tip`. And the equivalent in the JS SDK: `mineEmpyBlocks(n: number)`

It advances: 
- the data store height
- the burn data store height
- the tenure height

In epoch 3.0, we want a more granular control over the chain height

- advance_stacks_chaintip
  - only increases the stacks height
- advance_burn_chaintip
  - increases bitcoin, tenure, and stacks height
  - replaces the existing advance_chain_tip method/command to make it more explicit
  
  
It will allow to have precise control of the simulated chain height and handle real world scenarios in unit tests

tippenein:
updates advance_chaintip to handle burn and stacks chains separately

- closes #1495


components/clarity-repl/src/repl/session.rs
```diff
@@ -823,14 +835,41 @@ impl Session {
...
+        self.interpreter.advance_stacks_chaintip(count)
+    }
+    pub fn advance_burn_chaintip(&mut self, count: u32) -> u32 {
+        self.interpreter.advance_stacks_chaintip(count);
+        self.interpreter.advance_burn_chaintip(count)
+    }
+    // Not exposed/documented but still supported for convenience
     pub fn advance_chain_tip(&mut self, count: u32) -> u32 {
-        self.interpreter.advance_chain_tip(count)
+        self.interpreter.advance_stacks_chaintip(count)
```
hugocaillard, line 872: @tippenein This should call `advance_burn_chaintip`to keep the same behavior as before

components/clarinet-sdk/node/tests/simnet-usage-epoch3.test.ts
```diff
@@ -0,0 +1,54 @@
...
+
+function deleteExistingDeploymentPlan() {
+  if (fs.existsSync(deploymentPlanPath)) {
+    fs.unlinkSync(deploymentPlanPath);
+  }
+}
+
+beforeEach(async () => {
+  deleteExistingDeploymentPlan();
+  await fs.copyFileSync(customDeploymentPlanPath, deploymentPlanPath);
```
hugocaillard, line 29: No need to `await` here

components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -1098,12 +1098,23 @@ impl ClarityInterpreter {
...
+        let new_height = self.burn_datastore.advance_chain_tip(count);
+        let _ = self.datastore.advance_chain_tip(count);
         self.set_tenure_height();
         new_height
     }
+    pub fn advance_stacks_chaintip(&mut self, count: u32) -> Result<u32, String> {
+        let current_epoch = self.burn_datastore.get_current_epoch();
+        if current_epoch < StacksEpochId::Epoch30 {
+            Err(format!(
+                "stacks block height can't be advanced in {}",
```
hugocaillard, line 1111: I got this error message and actually find it hard to understand ("stacks block height can't be advanced in 2.4")

Can we instead have message that tells what should be done such as:
- "only burn chain height can be advanced in epoch lower than 3.0"
And in the sdk we could return "use `mineEmptyBurnBlock` or `mineEmptyBurnBlocks` in epoch lower than 3.0"

components/clarity-repl/src/repl/session.rs
```diff
@@ -1427,6 +1472,81 @@ mod tests {
         let mut session = Session::new(settings);
         session.start().expect("session could not start");
 
+        session.handle_command("::set_epoch 3.0");
```
hugocaillard, line 1475: This replaces in the test `evaluate_at_block`, which was testing `at-block` (and apparently a few other things, looks like it was really a "unit" test).
But the new test you introduced tests something else, although I'm not sure why.
In think that `evaluate_at_block` should remain untouched and you should add an other test

hugocaillard (changes requested):
A few change requests in the tests and couple of nitpicks.
Also, could you make sure that the new methods and session are covered by tests? https://app.codecov.io/gh/hirosystems/clarinet/pull/1506/blob/components/clarity-repl/src/repl/session.rs#L824 (`parse_and_advance_stacks_chaintip` & `parse_and_advance_burn_chaintip`

components/clarity-repl/src/repl/session.rs
hugocaillard: > the original is now below called evaluate_at_block_pre_epoch3

Yes, but I think it should remain `evaluate_at_block` since it tests the `at-block` clarity method, and isn't influenced by the epoch things.
Or, if we want to make sure it still works as expected in epoch3.0, I would add another test `evaluate_at_block_in_epoch_3_0` instead of changing what already existed. 


> what should this epoch 3 test be testing

idk, mayne it's redundant

hugocaillard (approved):
I don't know if we should enforce it, but it would be nice to have new lines add the end of each files

hugocaillard (changes requested):
I don't know if we should enforce it, but it would be nice to have new lines add the end of each files

components/clarity-repl/src/repl/session.rs
```diff
@@ -248,8 +248,15 @@ impl Session {
             cmd if cmd.starts_with("::get_contracts") => {
                 self.get_contracts().unwrap_or("No contract found".into())
             }
+            cmd if cmd.starts_with("::get_burn_block_height") => self.get_burn_block_height(),
+            cmd if cmd.starts_with("::get_stacks_block_height") => self.get_block_height(),
             cmd if cmd.starts_with("::get_block_height") => self.get_block_height(),
-            cmd if cmd.starts_with("::advance_chain_tip") => self.parse_and_advance_chain_tip(cmd),
```
hugocaillard, line 252: Even though it's removed from the docs, we want to keep it available (avoid breaking changes)

components/clarity-repl/src/repl/session.rs
```diff
@@ -823,14 +856,21 @@ impl Session {
...
             .to_string()
     }
 
+    pub fn advance_stacks_chaintip(&mut self, count: u32) -> Result<u32, String> {
+        self.interpreter.advance_stacks_chaintip(count)
+    }
+    pub fn advance_burn_chaintip(&mut self, count: u32) -> u32 {
+        self.interpreter.advance_burn_chaintip(count)
+    }
+    // Not exposed/documented but still supported for convenience
```
hugocaillard, line 871: Not sure this comment is needed.
It's "exposed" as a pub function and used internally.

components/clarinet-sdk/node/tests/simnet-usage-epoch3.test.ts
```diff
@@ -0,0 +1,33 @@
...
+import { Simnet, initSimnet } from "..";
+
+let simnet: Simnet;
+
+beforeEach(async () => {
+  simnet = await initSimnet("tests/fixtures/Clarinet.toml");
+  simnet.setEpoch("3.0");
+});
+
+describe("basic simnet interactions", () => {
```
hugocaillard, line 16: The describe is copy pasted from others test file

components/clarinet-sdk/node/tests/simnet-usage.test.ts
```diff
@@ -47,6 +47,11 @@ describe("basic simnet interactions", () => {
     simnet.mineEmptyBlocks(4);
     expect(simnet.blockHeight).toBe(blockHeight + 5);
   });
+  it("can not mine empty stacks block in pre-3.0", () => {
+    expect(() => simnet.mineEmptyStacksBlock()).toThrowError(
+      "use mineEmptyBurnBlock in epoch lower than 3.0"
+    );
+  })
```
hugocaillard, line 54: I think it would make more sense group all of the tests related to burn/stacks chains height in one file. Including this tests and the one from `simnet-usage-epoch3.test.ts`. (eg: `block-heights-handling.test.ts` or something else)

components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -1098,12 +1098,20 @@ impl ClarityInterpreter {
...
+    pub fn advance_burn_chaintip(&mut self, count: u32) -> u32 {
+        let new_height = self.burn_datastore.advance_chain_tip(count);
+        let _ = self.datastore.advance_chain_tip(count);
         self.set_tenure_height();
         new_height
     }
+    pub fn advance_stacks_chaintip(&mut self, count: u32) -> Result<u32, String> {
+        let current_epoch = self.burn_datastore.get_current_epoch();
+        if current_epoch < StacksEpochId::Epoch30 {
+            Err("only burn chain height can be advanced in epoch lower than 3.0".to_string())
```
hugocaillard, line 1110: I know this is an idea I suggested instead of the previous error message.
But I just tested in the clarinet console and I don't think the error message is clear for the user. 
it's misleading because in epoch lower than 3.0, both chain are advanced together (not only burn chain height is advanced).
Suggestion: maybe here we should just say that `advance_stacks_chaintip` can't be called in epoch lower than 3.0". But in session.rs, when we actually display an error to the users here https://github.com/hirosystems/clarinet/blob/1e1fc418f5c03dd023cb921e90696ee345e53987/components/clarity-repl/src/repl/session.rs#L842-L843

We should explain that this command should not be called in epoch lower than 3.0 and tell which command to use instead

hugocaillard (changes requested):

hugocaillard (approved):
LGTM 🚀

PR #1516, opened 2024-07-25
feat: follower node in devnet

hugocaillard:
Add an optional follower node in the devnet.
It's disabled by default. Most end-users will likely not use it, but not it can be useful in some situation, especially for stacks core developers

wip


vini-btc:
Hi @hugocaillard! Quick question: Would this feature allow to test reorgs?

hugocaillard:
@vini-btc No it wouldn't help

hugocaillard:
Closing this as it might not be that useful for now in clarinet

PR #1517, opened 2024-07-25
feat: add devnet start --default-settings

### LINKED ISSUE #1374: `clarinet devnet start` out of clarinet context
It'd be great to be able to spin up a devnet network locally without requiring a clarinet project.
We could introduce something like:
```console
$ clarinet devnet start --default-settings
```
- if a Clarinet.toml manifest file can not be found in the current directory, the `--default-settings` mode kicks in by default (even if not provided)
- if a Clarinet.toml manifest file can be found and the flag `--default-settings` is provided, the manifest file is ignored.

tippenein:
- closes #1374 

adds `clarinet devnet start --default-settings`


tippenein:
This was pretty far out of date and it should be easier now since a bunch of defaults have been added in the meantime

PR #1523, opened 2024-07-31
feat: add stacks_node_next_initiative_delay in devnet config

hugocaillard:
This setting has been introduced in stacks-node for nakamoto.
We should experiment around having a default of value of 3 or 4 seconds. Lower values could allow to handle faster Bitcoin blocks but can also break the blockchain.
Maybe in the future we'll warn the users that can't put a value lower than X seconds, but that ok for now, we'll communicate about epoch 3.0 support as "beta"

### DIFF
--- a/components/clarinet-cli/src/generate/project.rs
+++ b/components/clarinet-cli/src/generate/project.rs
@@ -341,7 +341,6 @@ disable_stacks_api = false
 # subnet_api_image_url = "{default_subnet_api_image}"
 # subnet_api_postgres_database = "subnet_api"
 
-# For testing in epoch 2.1 / using Clarity2
 # epoch_2_0 = {DEFAULT_EPOCH_2_0}
 # epoch_2_05 = {DEFAULT_EPOCH_2_05}
 # epoch_2_1 = {DEFAULT_EPOCH_2_1}
@@ -349,6 +348,7 @@ disable_stacks_api = false
 # epoch_2_3 = {DEFAULT_EPOCH_2_3}
 # epoch_2_4 = {DEFAULT_EPOCH_2_4}
 # epoch_2_5 = {DEFAULT_EPOCH_2_5}
+# epoch_3_0 = 144
 
 
 # Send some stacking orders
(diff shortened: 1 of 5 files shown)


components/clarinet-cli/src/generate/project.rs
```diff
@@ -341,14 +341,14 @@ disable_stacks_api = false
...
 
-# For testing in epoch 2.1 / using Clarity2
 # epoch_2_0 = {DEFAULT_EPOCH_2_0}
 # epoch_2_05 = {DEFAULT_EPOCH_2_05}
 # epoch_2_1 = {DEFAULT_EPOCH_2_1}
 # epoch_2_2 = {DEFAULT_EPOCH_2_2}
 # epoch_2_3 = {DEFAULT_EPOCH_2_3}
 # epoch_2_4 = {DEFAULT_EPOCH_2_4}
 # epoch_2_5 = {DEFAULT_EPOCH_2_5}
+# epoch_3_0 = 144
```
hugocaillard, line 351: We want to add this commented line for visibility/discoverability.
But keep it hard coded for now because the DEFAULT_EPOCH_3_0 will remain very high until epoch 3.0 lands on mainnet

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -434,7 +437,11 @@ pub async fn start_chains_coordinator(
             }
             ObserverEvent::NotifyBitcoinTransactionProxied => {
                 if !boot_completed.load(Ordering::SeqCst) {
-                    std::thread::sleep(std::time::Duration::from_secs(1));
+                    let _ = devnet_event_tx
+                        .send(DevnetEvent::info("1 - Waiting for boot mining".to_string()));
+                    std::thread::sleep(std::time::Duration::from_millis(2000));
```
hugocaillard, line 442: Need to revisit this change/value

components/stacks-network/src/ui/ui.rs
```diff
@@ -112,8 +112,7 @@ fn draw_devnet_status(f: &mut Frame, app: &mut App, area: Rect) {
         .style(Style::default().fg(Color::White))
         .borders(Borders::ALL)
         .title("Stacks Devnet");
-    let mut inner_area = block.inner(area);
-    inner_area.height = inner_area.height.saturating_sub(1);
+    let inner_area = block.inner(area);
```
hugocaillard, line 115: Small fix unrelated to this PR.
I think the "- 1" was a patch for something that has been fix in ratatui

PR #1545, opened 2024-08-24
generated from `clarinet completions zsh` command

OluTshegz:
git commit -m "Add clarinet completions for zsh"


hugocaillard:
@OluTshegz I don't think this file should be comitted.
If you think otherwise, could you add a description to the PR explaining your motivation?

hugocaillard:
Closing for now

PR #1555, opened 2024-09-05
Feat/customize number of signers in devnet

### LINKED ISSUE #1530: Make signers config customizable
## Context

The Stacks blockchain needs Signers to sign blocks in Nakamoto. For Nakamoto support, it's important that the Devnet setup has Signers and Stackers.
Two signers were setup in Clarinet in this PR https://github.com/hirosystems/clarinet/pull/1384/

At they time, the two signers were [hard coded](https://github.com/hirosystems/clarinet/blob/79ce8a417090ddabc6f3f72716c9a0c82de68de5/components/stacks-network/src/orchestrator.rs#L579-L663) and non customizable.

They use the [hard coded wallet_1 and wallet_2 private keys](https://github.com/hirosystems/clarinet/blob/79ce8a417090ddabc6f3f72716c9a0c82de68de5/components/stacks-network/src/orchestrator.rs#L1408-L1411) (default wallet in devnet.toml).

The stackers will then use one of these private keys ([hard coded again](https://github.com/hirosystems/clarinet/blob/79ce8a417090ddabc6f3f72716c9a0c82de68de5/components/stacks-network/src/chains_coordinator.rs#L688-L697)) to submit their stacking orders.

## Allow the Signers config to be customizable

Developers should be able to customize the signers config in the Devnet.toml. The config should allow 1 or multiple signers
It would also impact the stacking orders settings, we may want to add the ability to config the signer in key for the stackers.

For backward compatibility (existing Devnet.toml without signer configs), we should either provide a mechanism to add it on `clarinet devnet start` (promp the user "signers config not found, o you want to add the default one: ...") or have a default one in clarinet.

hugocaillard:
Adding a new `devnet.stacks_signers_keys` configuratoin key, that is an array of strings.
Each keys will result in a stacks signer container being booted as part of the devnet.

Stackers will also use these key to send the stacking txs.

This PR also adds a `devnet.stacks_signers_env_vars` (similar to `stacks_node_env_vars`)

resolves #1530

### DIFF
--- a/components/stacks-network/src/orchestrator.rs
+++ b/components/stacks-network/src/orchestrator.rs
@@ -1098,8 +1098,6 @@ amount = {}
                 r#"
 [[events_observer]]
 endpoint = "stacks-signer-{i}.{}:{}"
-retry_count = 255
-include_data_events = false
 events_keys = ["stackerdb", "block_proposal", "burn_blocks"]
 "#,
                 self.network_name,
@@ -1109,13 +1107,11 @@ events_keys = ["stackerdb", "block_proposal", "burn_blocks"]
 
         stacks_conf.push_str(&format!(
             r#"
-# Add orchestrator (docker-host) as an event observer
-[[events_observer]]
-endpoint = "host.docker.internal:{orchestrator_ingestion_port}"
-retry_count = 255
-include_data_events = true
-events_keys = ["*"]
-"#,
+        # Add orchestrator (docker-host) as an event observer
+        [[events_observer]]
+        endpoint = "host.docker.internal:{orchestrator_ingestion_port}"
+        events_keys = ["burn_blocks", "memtx"]
+        "#,
             orchestrator_ingestion_port = devnet_config.orchestrator_ingestion_port,
         ));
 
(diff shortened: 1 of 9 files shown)


components/stacks-network/src/orchestrator.rs
```diff
@@ -47,17 +49,6 @@ pub struct DevnetOrchestrator {
...
-
-// pub enum DevnetServices {
-//     BitcoinNode,
-//     StacksNode,
-//     StacksApi,
-//     StacksExplorer,
-//     BitcoinExplorer,
-//     SubnetNode,
-//     SubnetApi,
-// }
```
hugocaillard, line 59: dead code

components/stacks-network/src/orchestrator.rs
```diff
@@ -1144,37 +1093,28 @@ amount = {}
             ));
         }
 
-        // add the 2 signers event observers
-        stacks_conf.push_str(&format!(
-            r#"
-[[events_observer]]
-endpoint = "stacks-signer-1.{}:30001"
-retry_count = 255
-include_data_events = false
```
hugocaillard, line 1153: retry_count and include_data_events have both been removed from the observers config a while ago

PR #1558, opened 2024-09-14
chore: add disable_postgres flag and separate from stack_api disabling

tippenein:
In order to use clarinet's devnet for testing the api functionality we needed a way of disabling one or the other

`disable_postgres` field in Devnet.toml is essentially the main addition here, but also implies that `disable_stacks_api` no longer disables postgres which it _did_ do previously

### DIFF
--- a/components/stacks-network/src/orchestrator.rs
+++ b/components/stacks-network/src/orchestrator.rs
@@ -357,16 +357,6 @@ impl DevnetOrchestrator {
                 "initializing",
             );
         }
-        if !disable_postgres {
-            send_status_update(
-                &event_tx,
-                enable_subnet_node,
-                &self.logger,
-                "postgres",
-                Status::Red,
-                "initializing",
-            );
-        }
         if !disable_stacks_explorer {
             send_status_update(
                 &event_tx,
@@ -451,15 +441,6 @@ impl DevnetOrchestrator {
 
         // Start postgres container
         if !disable_postgres {
-            // Start postgres
-            send_status_update(
-                &event_tx,
-                enable_subnet_node,
-                &self.logger,
-                "postgres",
-                Status::Yellow,
-                "preparing postgres container",
-            );
             let _ = event_tx.send(DevnetEvent::info("Starting postgres".to_string()));
             match self.prepare_postgres_container(ctx).await {
                 Ok(_) => {}


components/stacks-network/src/event.rs
```diff
@@ -86,10 +86,11 @@ pub fn send_status_update(
         "stacks-node" => 1,
         "stacks-signers" => 2,
         "stacks-api" => 3,
-        "subnet-node" => 4,
-        "subnet-api" => 5,
-        "stacks-explorer" => subnet_services + 4,
-        "bitcoin-explorer" => subnet_services + 5,
+        "postgres" => 4,
```
hugocaillard, line 89: I'm not sure we want to have it here, displayed in the TUI. We didn't have it before, and it might not be relevant for common use cases. What do you think?

components/stacks-network/src/orchestrator.rs
```diff
@@ -346,30 +347,46 @@ impl DevnetOrchestrator {
...
+        if !disable_postgres {
+            send_status_update(
+                &event_tx,
+                enable_subnet_node,
+                &self.logger,
+                "postgres",
+                Status::Red,
+                "initializing",
+            );
+        }
```
hugocaillard, line 369: The postgres status updates can be removed if we don't show it in the UI right?

components/stacks-network/src/orchestrator.rs
```diff
@@ -432,14 +449,14 @@ impl DevnetOrchestrator {
...
-        if !disable_stacks_api {
+        // Start postgres container
+        if !disable_postgres {
             // Start postgres
             send_status_update(
                 &event_tx,
                 enable_subnet_node,
                 &self.logger,
-                "stacks-api",
+                "postgres",
```
hugocaillard, line 459: Same here

hugocaillard (changes requested):
TUI changes

components/stacks-network/src/orchestrator.rs
hugocaillard: Ah, I see.
Then the `send_status_update` function would need to be revamped because if no service name is matched, it does nothing I believe

hugocaillard (approved):
lgtm 👍

PR #1562, opened 2024-09-19
refactor: consolidate datastore and burn datastore data

hugocaillard:
wip


hugocaillard:
@tippenein 
Thx!
This is to address an actual issue, but ended up being a way bigger refactor than I expected.

Basically we went realistc block times in simnet between burn blocks and between stacks blocks. We were only tracking one type of blocks so far (doesn't really matter which type, since we had a 1:1 relation between burn and stacks blocks).
So now we store burn_blocks and stacks_blocks, and we store realistic times for both.
The `clarity_datastore` basically only holds the MARF keys and values now, and the `datastore` stores [burn and stacks blocks](https://github.com/hirosystems/clarinet/blob/b35e449f7d31c6df964ab0c4466ae0c5ba6b9b58/components/clarity-repl/src/repl/datastore.rs#L75-L78).
We know also have `burn_block_info` and `stacks_block_info`

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -98,107 +95,117 @@ fn height_to_id(height: u32) -> StacksBlockId {
...
+        //     .expect("ERROR: Failed to commit MARF block");
+    }
+
+    pub fn commit_to(&mut self, _final_bhh: &StacksBlockId) {
+        // println!("commit_to({})", final_bhh);
+        // self.side_store.commit_metadata_to(&self.chain_tip, final_bhh);
+        // self.side_store.commit(&self.chain_tip);
+        // self.marf.commit_to(final_bhh)
+        //     .expect("ERROR: Failed to commit MARF block");
+    }
```
krl, line 173: Why is the code in this section commented out?

krl: nvm saw your comment below. in general i prefer deleting code that is not used to reduce cognitive overhead. It's still in the history if needed.

components/clarity-repl/src/repl/clarity_values.rs
```diff
@@ -22,47 +22,52 @@ pub fn uint8_to_value(mut value: &[u8]) -> Value {
...
-            let mut data = Vec::new();
-            for (name, value) in tup_data.data_map.iter() {
-                data.push(format!("{}: {}", &**name, value_to_string(value)))
-            }
-            format!("{{ {} }}", data.join(", "))
+            let data = tup_data
+                .data_map
+                .iter()
+                .map(|(name, value)| format!("{}: {}", name, value_to_string(value)))
+                .collect::<Vec<_>>()
```
krl, line 31: this is doing a lot of allocations for no good reason, first in each format, then in the collect, and finally again in join.

You could instead setup an empty mut string and write! to it, which would only allocate the string

krl (approved):
looks good

components/clarity-repl/src/repl/datastore.rs
hugocaillard: Yes I agree, but I'm not sure why it was left there in the first. And since it's not really the point of this PR, I'll come back at it later I think

components/clarity-repl/src/repl/clarity_values.rs
hugocaillard: Addressed

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -319,193 +325,316 @@ impl ClarityBackingStore for Datastore {
...
+        let stacks_block_height = self.stacks_chain_height;
+
+        let last_stacks_block = self
+            .stacks_blocks
+            .get(&clarity_datastore.current_chain_tip)
+            .expect("current chain tip missing in stacks block table");
+        let last_burn_block = self
+            .burn_blocks
+            .get(&height_to_burn_block_header_hash(burn_chain_height))
+            .expect("burn block missing in burn block table");
```
csgui, line 403: An area to consider some improvement, if is the case:

. using `expect()` after these lookups could panic the program if the keys are not found.

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -319,193 +325,316 @@ impl ClarityBackingStore for Datastore {
...
+    ) -> u32 {
+        for _ in 1..=count {
+            let last_stacks_block = self
+                .stacks_blocks
+                .get(&clarity_datastore.current_chain_tip)
+                .unwrap();
+            let last_burn_block = self
+                .burn_blocks
+                .get(&last_stacks_block.burn_block_header_hash)
+                .unwrap();
```
csgui, line 452: The use of `unwrap()` on `self.stacks_blocks.get()` and `self.burn_blocks.get()` means that the function will panic if these lookups fail.

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -319,193 +325,316 @@ impl ClarityBackingStore for Datastore {
...
+
+            let mut next_burn_block_time = last_burn_block.burn_block_time + 600;
+            if last_stacks_block.stacks_block_time > next_burn_block_time {
+                next_burn_block_time = last_stacks_block.stacks_block_time + 10;
+            }
```
csgui, line 457: What are those hard-coded values 600 and 10 in time calculations.? These constants could be extracted as named variables or configuration parameters for clarity and maintainability.

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -319,193 +325,316 @@ impl ClarityBackingStore for Datastore {
...
                 .insert(block_info.consensus_hash, sortition_id);
-            self.store.insert(id, block_info);
+            self.stacks_blocks.insert(id, block_info);
+
+            clarity_datastore
+                .block_id_lookup
+                .insert(id, current_lookup_id);
+            clarity_datastore
+                .height_at_chain_tip
+                .insert(id, self.stacks_chain_height);
```
csgui, line 502: These operations, inserting into hashmaps and updating states, assume no failure. Error handling could be improved to handle cases where these operations fail.

csgui (commented):
I verified the `datastore` changes, and they seem to be ok. I added some remarks that could be improved, if applicable.

components/clarity-repl/src/repl/datastore.rs
hugocaillard: Good call, addressed

components/clarity-repl/src/repl/datastore.rs
hugocaillard: Addressed

csgui (approved):
Changes on datastore LGTM!

Issue #1447, opened 2024-05-03
Incorrect stx costs for deployment


friedger:
**Describe the bug**
When applying a deployment, the costs are shown as 0.001 STX instead of 0.1 STX

**To Reproduce**
1. Create a deployment
2. Provide costs of 100000 uSTX
3. Apply deployment and see preview

**Expected behavior**
Costs are converted correctly from uSTX to STX

**Screenshots**
![image](https://github.com/hirosystems/clarinet/assets/1449049/f74b2b6d-edec-4bdc-a62b-45dc1d70be40)

hugocaillard:
Good catch, thanks for reporting @friedger

hugocaillard:
Fixed by #1563
Released in 2.9.0

PR #1563, opened 2024-09-19
fix: convert correctly from uSTX to STX when applying a deployment on issue #1447

erickcestari:
### Description

This PR addresses an issue with the conversion from micro STX (uSTX) to STX during deployment. Previously, the conversion was not handled correctly, leading to incorrect STX values being applied. This fix ensures that the conversion is accurate and that the correct STX values are used.

Related issues and PRs:

Issue: #1447

#### Breaking change?

No breaking changes are introduced by this PR.

### Example

Example
Before this fix, deploying with micro STX values would result in incorrect STX values being applied. After this fix, the conversion is handled correctly, ensuring that the deployed values are accurate.


hugocaillard (approved):
LGTM!
Thx @erickcestari

PR #1568, opened 2024-09-26
feat: handle stacks reorg

hugocaillard:
Tihs is a patch for a case that should not happen under normal conditions.
More investigation to happen with block chain team.

Also, this only impacts the UI

### DIFF
--- a/components/stacks-network/src/chains_coordinator.rs
+++ b/components/stacks-network/src/chains_coordinator.rs
@@ -391,8 +391,18 @@ pub async fn start_chains_coordinator(
                     StacksChainEvent::ChainUpdatedWithMicroblocksReorg(_) => {
                         unreachable!() // TODO(lgalabru): good enough for now - code path unreachable in the context of Devnet
                     }
-                    StacksChainEvent::ChainUpdatedWithReorg(_) => {
-                        unreachable!() // TODO(lgalabru): good enough for now - code path unreachable in the context of Devnet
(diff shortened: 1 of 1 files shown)


components/stacks-network/src/chains_coordinator.rs
```diff
@@ -391,8 +391,18 @@ pub async fn start_chains_coordinator(
                     StacksChainEvent::ChainUpdatedWithMicroblocksReorg(_) => {
                         unreachable!() // TODO(lgalabru): good enough for now - code path unreachable in the context of Devnet
                     }
-                    StacksChainEvent::ChainUpdatedWithReorg(_) => {
-                        unreachable!() // TODO(lgalabru): good enough for now - code path unreachable in the context of Devnet
+                    StacksChainEvent::ChainUpdatedWithReorg(data) => {
+                        // looks like in nakamoto we have reorgs, lets try that
+                        let reorg_data = format!("{:#?}", data);
```
hugocaillard, line 396: this will be removed in the final version of the PR
just keeping it to have some logs when testing

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -391,8 +391,18 @@ pub async fn start_chains_coordinator(
...
+                        std::fs::write("reorgdata.txt", reorg_data)
+                            .expect("Failed to write reorg data to file");
```
hugocaillard, line 398: this will be removed in the final version of the PR
just keeping it to have some logs when testing

PR #1574, opened 2024-10-04
feat: add a logger hook

hugocaillard:
### Description

The current way to retrieve the result of a `(print <value>)` function call in the tests, is to retrieve the events.
But, events are not emitted in read only functions calls or in failing public functions calls. Which is when logs are the most needed.

By using an evalhook, we can retrieve the value of the printed value and return it to the sdk so that it can be logged in JS

### DIFF
--- a/components/clarinet-cli/src/generate/contract.rs
+++ b/components/clarinet-cli/src/generate/contract.rs
@@ -31,11 +31,6 @@ impl GetChangesForRmContract {
         f.append_path("tests")?;
         f.append_path(&name)?;
         if !f.exists() {
-            format!(
-                "{} tests/{} doesn't exist. Skipping removal",
-                red!("Warning"),
-                name
-            );
             return Ok(());
         }
         let change = FileDeletion {
--- a/components/clarity-repl/src/analysis/mod.rs
+++ b/components/clarity-repl/src/analysis/mod.rs
@@ -154,9 +154,8 @@ where
     F: FnOnce(&mut AnalysisDatabase) -> std::result::Result<T, E>,
 {
     conn.begin();
-    let result = f(conn).map_err(|e| {
+    let result = f(conn).inspect_err(|_| {
         conn.roll_back().expect("Failed to roll back");
-        e
     })?;
     conn.commit().expect("Failed to commit");
     Ok(result)
--- a/components/clarity-repl/src/repl/logger_hook.rs
+++ b/components/clarity-repl/src/repl/logger_hook.rs
@@ -1,0 +1,59 @@
+use clarity::vm::errors::Error;
+use clarity::vm::functions::NativeFunctions;
+use clarity::vm::representations::Span;
+use clarity::vm::types::QualifiedContractIdentifier;
+use clarity::vm::{
+    contexts::{Environment, LocalContext},
+    types::Value,
+    EvalHook, ExecutionResult, SymbolicExpression,
+    SymbolicExpressionType::List,
+};
+
+pub struct ContractLog {
+    contract_id: QualifiedContractIdentifier,
+    span: Span,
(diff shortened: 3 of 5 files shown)


components/clarity-repl/src/repl/hooks/logger.rs
```diff
@@ -0,0 +1,55 @@
...
+            return;
+        };
+        let Some(function_name) = function_name.match_atom() else {
+            return;
+        };
+
+        if let Some(NativeFunctions::Print) = NativeFunctions::lookup_by_name(function_name) {
+            let contract_name = &env.contract_context.contract_identifier.name;
+            let span = &expr.span;
+            let line_annotation = format!("({}:{})", contract_name, span.start_line);
```
jbencin, line 41: Clippy is going to complain about this now

```suggestion
            let line_annotation = format!("({contract_name}:{})", span.start_line);
```

components/clarity-repl/src/repl/remote_data/http_request/node.rs
```diff
@@ -44,9 +44,9 @@ pub fn http_request<T: DeserializeOwned>(url: &str) -> Result<T, String> {
     curl_command.push("-H \"Accept: application/json\"".to_string());
     curl_command.push("-H \"x-hiro-product: clarinet-sdk\"".to_string());
     if let Some(api_key) = &*HIRO_API_KEY {
-        curl_command.push(format!("-H \"x-api-key: {}\"", api_key));
+        curl_command.push(format!("-H \"x-api-key: {api_key}\""));
```
jbencin, line 47: How did this get skipped in #1877? I've had a feeling for a while that there are parts of the codebase that get ignored by Clippy

components/clarity-repl/src/repl/hooks/logger.rs
```diff
@@ -0,0 +1,55 @@
+use clarity::vm::contexts::{Environment, LocalContext};
+use clarity::vm::errors::Error;
+use clarity::vm::functions::NativeFunctions;
+use clarity::vm::types::Value;
+use clarity::vm::SymbolicExpressionType::List;
```
jbencin: I think it's kinda awkward to import a single enum variant. Can you make this match `tracer.rs`, import the `enum`, and reference it as `SymbolicExpressionType::List`?

jbencin (dismissed):
Looks good just some minor style comments

components/clarity-repl/src/repl/hooks/logger.rs
hugoclrd: It's actually not.
When one of the argument has to be passed as a paramaters (`span.start_line`), clippy allows consistency and have everything as parameters.
I was happy that it wasn't complaining because it does look better to me with this consistency. What do you think?

jbencin (dismissed):

components/clarity-repl/src/repl/remote_data/http_request/node.rs
hugoclrd: Yes indeed. Everything behind `target_arch = "wasm32"` conditional 

But I did run clippy fix manually on this target when opening #1877.
Maybe this on was skipped for some reason. But we can add the check #1895 (intentionally failing for now so that we can observe the error)


https://github.com/hirosystems/clarinet/blob/0b185dc3816c07fb552a0410e0b0d86ec7028f7d/components/clarity-repl/src/repl/remote_data/http_request/mod.rs#L6-L14

hugoclrd:
@jbencin that last commit dismissed your review.
It simply revert a change unrelated to this PR. I uncommented a test because i didn't remember why it was commented.
I now do, it fails on windows in the CI.
Will inspect as part of an other PR

PR #1580, opened 2024-10-09
feat: session.eval_with_hooks

hugocaillard:
wip - exploring an other interface for session hooks

With this approach:
- the Session is more aware of it's hooks
  - instead of passing `eval_hooks` to every functions (`eval`, `deploy_contract` , `call_contract`), the caller only enable an hook once
  - in the future we could have `session.enable_debug_print_hook`, `session.enable_tracer_hook`, etc
- I added `.eval_with_hooks` that keeps the same behaviour and is backward compatible with exisintg implementation. It allows to evaluated snippets with arbitrary hooks
  - Should we choose to use the implemation, I should add some documentation in the comments for the `eval_with_hooks` method

Concerns: 
- is the Session supposed to be that aware of the hooks, or is it better to keep them as parameters, so that the Sessionis mor versatile

### DIFF
--- a/components/clarity-repl/src/analysis/coverage_tests.rs
+++ b/components/clarity-repl/src/analysis/coverage_tests.rs
@@ -1,18 +1,16 @@
 use std::collections::BTreeMap;
 
-use super::coverage::CoverageHook;
 use crate::repl::session::Session;
 use crate::repl::SessionSettings;
 
 fn get_coverage_report(contract: &str, snippets: Vec<String>) -> String {
     let mut session = Session::new(SessionSettings::default());
+    session.enable_coverage();
+    session.set_test_name("test_scenario".to_string());
 
-    let mut coverage_hook = CoverageHook::new();
-    coverage_hook.set_current_test_name("test_scenario".to_string());
-
-    let _ = session.eval(contract.into(), Some(vec![&mut coverage_hook]), false);
+    let _ = session.eval(contract.into(), false);
     for snippet in snippets {
-        let _ = session.eval(snippet, Some(vec![&mut coverage_hook]), false);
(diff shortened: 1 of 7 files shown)


components/clarity-repl/src/analysis/coverage_tests.rs
```diff
@@ -721,10 +718,25 @@ fn multiple_test_files() {
...
+            "BRF:0",
+            "BRH:0",
+            "end_of_record",
+            "SF:/contract-1.clar",
+            "FN:1,add",
+            "FNF:1",
+            "FNH:0",
+            "BRF:0",
+            "BRH:0",
+            "end_of_record",
```
hugocaillard, line 739: As a sidev effect, this is the wanted behaviour for code coverage.
Because session.deploy_contract uses the session coverage hook.
This is therefore the initial analysis of the contracts.
If we wanted to pass the eval_hooks to `update_session_with_deployment_plan` ([here](https://github.com/hirosystems/clarinet/blob/6f178d8ca8a9f7a29db749c077214082462741ff/components/clarinet-sdk-wasm/src/core.rs#L455)) instead, this would make pass eval_hooks args in a whole lot more of functions (and it comes with some difficulties down the line)

hugocaillard: So this a wanted behavior for better coverage handling; It doesn't really impact code coverage % though

hugocaillard:
@tippenein This PR allows both

- enable hook at the session level, so that all of the impact method (eval, deploy_contract, call_contract and others) will use the session hooks
- leave an `eval_with_hooks` method that allows to pass arbitrary hooks to an eval call

PR #1583, opened 2024-10-11
feat: add types to support `StacksSignerMessage` serialization

rafaelcr:
Type definitions were taken directly from stacks-core.


components/stacks-codec/src/codec.rs
```diff
@@ -2774,6 +2805,93 @@ pub struct NakamotoBlockHeader {
...
+        // if tx_merkle_root != header.tx_merkle_root {
+        //     warn!("Invalid block: Tx Merkle root mismatch";
+        //         "consensus_hash" => %header.consensus_hash,
+        //         "stacks_block_hash" => %header.block_hash(),
+        //         "stacks_block_id" => %header.block_id()
+        //     );
+        //     return Err(CodecError::DeserializeError(
+        //         "Invalid block: tx Merkle root mismatch".to_string(),
+        //     ));
+        // }
```
rafaelcr, line 2889: @hugocaillard these validations added a few extra dependencies so I commented them out. Not sure if you prefer these validations to stay in place, though, I can re-add them if necessary.

PR #1586, opened 2024-10-18
refactor: improve handling of enums with big size variations

hugocaillard:
Clippy fixes for [large_enum_variant](https://rust-lang.github.io/rust-clippy/master/index.html#large_enum_variant)

In the LSP, the `Initialize` Response and Request are only sent at the initialization of the LSP, then it's only the other kinds that are used. So it really makes sense to box those.

I preferred to ignored the changes in deployments, because clarinet onchain deployments are less memory intensive (and, less used as well). So the benefits in changing those was less obvious

### DIFF
--- a/components/clarinet-deployments/src/onchain/mod.rs
+++ b/components/clarinet-deployments/src/onchain/mod.rs
@@ -198,6 +198,7 @@ pub fn encode_contract_publish(
     )
 }
 
+#[allow(clippy::large_enum_variant)]
 #[derive(Clone, Debug)]
 pub enum TransactionStatus {
     Queued,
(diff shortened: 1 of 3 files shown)


components/clarinet-deployments/src/onchain/mod.rs
```diff
@@ -198,6 +198,7 @@ pub fn encode_contract_publish(
     )
 }
 
+#[allow(clippy::large_enum_variant)]
```
csgui, line 201: Ok. Seems valid to ignore large variants in this case.

components/clarinet-deployments/src/onchain/mod.rs
```diff
@@ -221,6 +222,7 @@ pub enum TransactionCheck {
     BtcTransfer,
 }
 
+#[allow(clippy::large_enum_variant)]
```
csgui, line 225: Ok. Seems valid to ignore large variants in this case.

csgui (approved):
Code LGTM

PR #1591, opened 2024-10-24
fix: update SignerMessage codec

zone117x:
Update SignerMessage codec to match core


hugocaillard (approved):
LGTM! 
thanks a lot @zone117x

Issue #796, opened 2023-01-25
Advanced trait completion


obycode:
It could be cool if in the completion for a trait type, e.g. when I type `<` in a place where a type is expected, if in addition to suggesting traits which are defined in this contract, or already brought in with `use-trait`, it could suggest traits from other contracts in the project, and add the `use-trait` at the top for you, similar to how in Rust, it will suggest an auto-complete, then automatically add the corresponding `use` for you.

lgalabru:
Kind of related https://github.com/hirosystems/clarinet/issues/600

PR #1592, opened 2024-10-25
feat: trait autocompletion

usagi32:
addresses #600 and #796 

what this pr does:

- Introduces autocompletion and auto import of traits and aliases. When typing '<' where a type is expected, a list of available trait aliases should be visible and on selecting one the corresponding (use-trait...) syntax would be auto imported on the top or if there is already a section with use trait imports.

- Autocompletion on principal when " \' " is typed and their corresponding contracts and traits suggestions on further typing '.'.

- Trait identidfier suggestions in (impl-trait ...) and if some methods of the selected trait are not implemented before their signatures will also be auto implemented.

Some refactoring can be done.

I was not able to test this because of a dependency in stacks-core/clarity. Nevertheless, I included some manual tests because of that.

### DIFF
--- a/components/clarity-lsp/src/common/backend.rs
+++ b/components/clarity-lsp/src/common/backend.rs
@@ -286,7 +286,7 @@ pub fn process_request(
             };
 
             let completion_items = match editor_state
-                .try_read(|es| es.get_completion_items_for_contract(&contract_location, &position))
+                .try_read(|es| es.get_completion_items_for_contract(&contract_location, &position, &params.context))
             {
                 Ok(result) => result,
                 Err(_) => return Ok(LspRequestResponse::CompletionItems(vec![])),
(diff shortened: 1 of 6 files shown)


usagi32:
There is a dependency that needs to be merged in the clarity-vm [here](https://github.com/stacks-network/stacks-core/pull/5383).

hugocaillard:
Thanks for this great PR @usagi32 🙌 
At first glance it looks nice, let's wait on the clarity-vm PR to move this one forward. 

If you plan on working on other clarinet issue, can you please comment on the related issue (or open one if missing). I'll try to do same, I've been working on some lsp improvements as well lately (showing variable and functions types on hover)

hugocaillard:
Closing this PR for now given the necessary upstream work

PR #1595, opened 2024-10-25
fix: make MockSignature signature field public

rafaelcr:


hugocaillard:
Do you still need that @rafaelcr .

hugocaillard:
@rafaelcr Ok good, can you just run cargo format on the PR please?
Sorry it got lost for so long

hugocaillard (approved):
lgtm

PR #1596, opened 2024-10-26
docs(README): recent link testnet

guspan-tanadi:
### Description

Navigate a link to testnet page.

### Example

---

### Checklist


hugocaillard:
The link below is outdated as well, maybe this all seciton should just link to https://docs.hiro.so/stacks/clarinet/concepts/network-types

hugocaillard (changes requested):
The link below is outdated as well, maybe this all section should just link to https://docs.hiro.so/stacks/clarinet/concepts/network-types

hugocaillard (changes requested):
I think thie bullet points should be removed, and instead have something like

`[...] allowing you to deploy the contract to the [devnet, testnet, or mainnet](link/to/doc)` 

Since the docs now explain the diff between the network, there's no need to explain it here

hugocaillard (approved):
lgtm, thanks @guspan-tanadi

PR #1605, opened 2024-10-30
feat: support stacks.js v7

hugocaillard:
### Description

Update the clarinet js sdk to use stacks.js 7


components/clarinet-cli/src/generate/project.rs
```diff
@@ -425,11 +425,11 @@ btc_address = "mvZtbibDAAA3WLpY7zXXFqRa3T4XSknBX7"
   "license": "ISC",
   "dependencies": {{
     "@hirosystems/clarinet-sdk": "^2.3.2",
-    "@stacks/transactions": "^6.12.0",
+    "@stacks/transactions": "^7.0.0",
     "chokidar-cli": "^3.0.0",
     "typescript": "^5.3.3",
     "vite": "^5.1.4",
-    "vitest": "^1.3.1",
+    "vitest": "^2.1.4",
```
pradel, line 432: vitest released a v3, you might want to jump from 1 > 3 directly as this pr is a breaking change see https://github.com/vitest-dev/vitest/releases/tag/v3.0.0 :)

pradel:
> release the clarinet sdk as 3.x but decouple it from clarinet versionning (and keep the cli as 2.x), would make sense but bring a bit of confusion

Making clarinet and the SDK have the same version is nice, what's the issue with releasing this as clarinet v3?

hugocaillard:
> Making clarinet and the SDK have the same version is nice, what's the issue with releasing this as clarinet v3?

Some people can expext major updates or new features from a new major version.
And for developers to upgrade (and adapt to breaking changes), they have to find some value in it.

But we are doing it. Here is (almost) everything that will be included in 3.0: #1672

hugocaillard:
> Making clarinet and the SDK have the same version is nice, what's the issue with releasing this as clarinet v3?

Some people can expext major updates or new features from a new major version.
And for developers to upgrade (and adapt to breaking changes), they have to find some value in it.

But we are doing it. Here is (almost) everything that will be included in 3.0: #1672 
cc @pradel

components/clarinet-cli/src/generate/project.rs
hugocaillard: @pradel 
You can start using vitest v3 today, just make to have
- `@hirosystems/clarinet-sdk` >= 2.14.0
- `vitest-environment-clarinet` >= 2.3.0

And you'll be able to use any vitest version from 1 to 3.
We'll default to 3.0 in future versions 👍

components/clarinet-sdk/node/vitest-helpers/src/clarityValuesMatchers.ts
```diff
@@ -286,7 +290,7 @@ expect.extend({
       return errorToAssertionResult.call(this, e);
     }
 
-    const actualString = principalToString(actual);
+    const actualString = cvToString(actual, "tryAscii");
```
janniks, line 293: Why not the `Cl.stringify` aka `prettyPrint` that you added?

janniks (approved):
Looks great! 👏 Functional, clean!

components/clarinet-sdk/node/vitest-helpers/src/clarityValuesMatchers.ts
hugocaillard: If you think this is the best replacement for `principalToString` i should use it

hugocaillard: In this particular case, we don't want the `'` at the beginning of the string. 
But Cl.stringify rightfully puts it.
I'm tempting to stick to cvToString, unless you tell me it'll be deprecated in the future

hugocaillard:
```suggestion
    const actualString = Cl.prettyPrint(actual, 2).replace(/^'/, "");
```

janniks: Got it, that makes sense. No no, fine to use either. Was just curious :)

janniks (approved):

Issue #606, opened 2022-03-03
Add an opinionated formatter


obycode:
It is always useful to have a formatter to fix whitespace and other minor formatting changes in your code. Our VSCode extension should provide a formatter for Clarity files.

lgalabru:
Related https://github.com/hirosystems/clarinet/pull/664

smcclellan:
This is in the "Gathering design options" phase of this initiative. Discussion to happen in #clarity Discord channel.

PR #1609, opened 2024-11-12
Clarinet format

tippenein:
fix: #606 

Features:
Adds `format`/`fmt` to clarinet
```
  -m, --manifest-path <MANIFEST_PATH>      
  -f, --file <FILE>                        If specified, format only this file
  -l, --max-line-length <MAX_LINE_LENGTH>  
  -i, --indent <INDENTATION>               indentation size, e.g. 2
  -t, --tabs                               use tabs instead of spaces
      --dry-run                            
      --in-place 
```

remaining work

- [x] fix pre- and post-comments
- [x] top level max-line breaking
- [x] @format-ignore comments (possibly pushed off until post-mvp)
- [x] passing golden test suite

soon:
- LSP/VSCode integration (https://github.com/hirosystems/clarinet/pull/1632)


components/clarinet-format/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "clarinet-format"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+clarinet-files = { path = "../clarinet-files", default-features = false, optional = true }
+clarity-repl = { path = "../clarity-repl" }
```
hugocaillard, line 8: We might not need `clarity-repl` here, but instead we can directy depend on `clarity`.
I'm not a big fan of `clarity-repl` re-exporting all of the VM, and it creates a dependency mess.

```toml
clarity = { workspace = true }
```

Eventually, you will also have to manage the clarity feature like for native and wasm (you can see this pattern in other Clarinet components, such as the repl

```toml
[features]
default = ["cli"]
cli = [
    "clarity/canonical",
    "clarity/developer-mode",
    "clarity/devtools",
    "clarity/log",
]
wasm = [
    "clarity/wasm",
    "clarity/developer-mode",
    "clarity/devtools",
]
```

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -94,11 +95,31 @@ enum Command {
...
     /// Step by step debugging and breakpoints from your code editor (VSCode, vim, emacs, etc)
     #[clap(name = "dap", bin_name = "dap")]
     DAP,
 }
 
+#[derive(Parser, PartialEq, Clone, Debug)]
+struct Formatter {
+    /// Path to clarity files (defaults to ./contracts)
+    #[clap(long = "path", short = 'p')]
+    pub code_path: Option<String>,
```
hugocaillard, line 110: By default, the formatter should be based on the manifest.
Load the manifest like other comamnds do (looking for `./Clarinet.toml`). And an option should allow to specify the manifest path

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1180,6 +1201,28 @@ pub fn main() {
                 process::exit(1);
             }
         },
+        Command::Formatter(cmd) => {
+            let sources = get_source_with_path(cmd.code_path, cmd.file);
```
hugocaillard, line 1205: Get the contracts from the manifest

tippenein:
Giving an update here. 

1. We need a way of organizing the traversal to identify expression-blocks to wrap. Currently, you could attempt to check line lengths and wrap conditionally as you go (tried this), but it won't work easily if you need multiple wraps per block. My thought is we need a `Wrap` type that's just used while traversing the PSEs and checks the span column length and start_line diff to turn wrapping on or off.

2. We need to figure out `ignored_expr` for skipping formatting if specified. The function that exists is basically a stub but it _should_ be possible to reconstruct the source from the PSE spans..

3. There's a fair amount of nested indentation that fails within golden tests (the unit tests don't test this much). I would guess that a better way of handling indentation wrapping would ease this (rather than the current attempt of passing in `previous_indentation` and constructing the string from that.. Very error prone.

4. This is more of a refactor, but basically we're doing way too much arg passing. There's a branch ([`experiment-with-accumulator`](https://github.com/hirosystems/clarinet/commit/f619b0a6fa15749c4eb390033ca4053d9df1cd6c#diff-bdfb6363ac810ef4d419caf59fba1e9b747a6a281fbd56b6d87a70cf66c226f4R107-R149)) which i attempted to use an Accumulator struct to pass around the relevant info for constructing the result but it got a bit hairy with dereferencing and ultimately stack overflowed when I got it compiling. It's probably worth looking into because it would help fix 3 implicitly with the `indent` `dedent` and `newline` methods on `Accumulator`

components/clarinet-format/Cargo.toml
```diff
@@ -0,0 +1,31 @@
+[package]
+name = "clarinet-format"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+# clarity-repl = { path = "../clarity-repl" }
```
hugocaillard, line 7: can be removed

components/clarinet-format/tests/golden-intended/alex-transfer.clar
```diff
@@ -0,0 +1,134 @@
...
+                    (if (is-eq no-to-treasury u0)
+                            (list)
+                            (unwrap-panic
+                            (slice? owned-by-sender
+                                (- (len owned-by-sender) no-to-treasury)
+                                (len owned-by-sender)
+                            )
+                        )
+                        )
+                )
```
hugocaillard, line 60: There are some indentation issues here right? (double indent within `if` and mis-aligned closing parenthesis lines 58 and 59
```
               (ids-to-treasury
                    (if (is-eq no-to-treasury u0)
                        (list)
                        (unwrap-panic
                            (slice? 
                                owned-by-sender
                                (- (len owned-by-sender) no-to-treasury)
                                (len owned-by-sender)
                            )
                        )
                    )
                )
```

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+				index: uint
+			}
+		)
+	)
+	(let (
+		(data (get txbuff ctx))
+		(base (get index ctx))
+	)
+		(ok
+			{
```
hugocaillard, line 44: Could/should the new line before `{` be avoided?
Looks like there's this pattern in all `(ok <tuple>)`

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+					ctx: {
+						txbuff: tx,
+						index: (+ u1 ptr)
+					}
+				}
+			)
+				(if (is-eq byte u253)
+					(let (
+					;; next two bytes is the varint
+					(parsed-u16
```
hugocaillard, line 186: missing indentation?

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+							(try!
+								(read-uint64
+									{
+										txbuff: tx,
+										index: (+ u1 ptr)
+									}
+								)
+							)
+						)
+					)
```
hugocaillard, line 236: What is the rule/logic for new lines here?
I feel like I'd be happy with something like
```clarity
            (let (
							;; next eight bytes is the varint
							(parsed-u64 (try! (read-uint64 {
								txbuff: tx,
								index: (+ u1 ptr)
							})))
						)
```

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+		(result
+			(response
+				{
+					ctx: {
+						txbuff: (buff 4096),
+						index: uint
+					},
+					remaining: uint,
+					txins: (list 8
+						{
```
hugocaillard, line 357: The new line before `{` seems unnecessary to me as well here

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+				(err ERR-TOO-MANY-TXINS)
+				(fold read-next-txin (bool-list-of-len num-txins)
+				(ok
+					{
+						ctx: new-ctx,
+						remaining: num-txins,
+						txins: (list)
+					}
+				)
+			)
```
hugocaillard, line 439: Missing indentation level?

components/clarinet-format/tests/golden-intended/tuple.clar
```diff
@@ -0,0 +1,19 @@
...
+    (user principal)
+    (asset principal) ;; comment
+    (state
+      {
+        principal-borrow-balance: uint,
+        last-variable-borrow-cumulative-index: uint,
+        origination-fee: uint,
+        stable-borrow-rate: uint,
+        last-updated-block: uint, ;; comment
+        use-as-collateral: bool,      }
```
hugocaillard, line 11: Missing new line before `}`?

hugocaillard (commented):
Took a quick look at it, I left a few comments/questions in the intended golden outputs. Just some of the most visible patterns of issues that seems important to me

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
tippenein: it depends on the source. if it's over max width it'll newline it. This could be changed but it starts looking weird if you start a map on the same line as `(ok ..` and then dump all the rest of the keys on the next indented lines

```
(ok { something: 1,
  a: b,
  c: d
  }
)
```

Oh, maybe I see your point. It should be this?

```
(ok {
  something: 1,
 ```

File: components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar (hunk 8 above)
Comment (tippenein) [reply components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar:236]: It does the paren lining up here: https://github.com/hirosystems/clarinet/pull/1609/files#diff-bdfb6363ac810ef4d419caf59fba1e9b747a6a281fbd56b6d87a70cf66c226f4R779-R781

It's quite easy to stack the closing parens `)))` by not including the above logic. 

Maybe it makes sense in some situations, but this is generic inner content handling. It's also worsened by how bad tabs look here, I think..

File: components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar (hunk 10 above)
Comment (tippenein) [reply components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar:439]: interesting.. `fold` wrapped wrong. Probably `inner_content` bug.. Will check 👌

File: components/clarinet-format/tests/golden-intended/alex-transfer.clar (hunk 5 above)
Comment (tippenein) [reply components/clarinet-format/tests/golden-intended/alex-transfer.clar:60]: The double indent _is_ intended. The tricky part here is that the double indent is for situations where the `if` is nested inside something, like

```
(key (if (...)
    (list) ;; double indent here
    ...
```

but since the line was too long, it got broken up onto separate lines..

Open to whatever changes, but I'll fix the `unwrap-panic` indenting since that's clearly wrong

File: components/clarinet-format/src/formatter/mod.rs [hunk 12]
```diff
@@ -0,0 +1,1232 @@
...
+    );
+
+    for expr in list.iter() {
+        let formatted = format_source_exprs(settings, &[expr.clone()], &base_indent);
+        let trimmed = t(&formatted);
+        let expr_width = trimmed.len();
+
+        if !first_on_line {
+            // For subexpressions that would exceed line length, add newline with increased indent
+            if current_line_width + expr_width + 1 > 80 {
```
csgui, line 761: Consider changing the hardcoded value of `80` to use `settings.max_line_length` instead, so that formatting is configurable.

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1232 @@
...
+            } else {
+                // Handle non-list values (e.g., literals or simple expressions)
+                acc.push(' ');
+                acc.push_str(&display_pse(settings, value, ""));
+                acc.push(')');
+            }
+        }
+
+        acc.push('\n');
+        acc.to_owned()
```
csgui, line 244: `acc` is already an owned String. Probably you can simply return `acc` without invoking `.to_owned()`, which avoids extra allocations.

components/clarinet-format/src/formatter/ignored.rs
```diff
@@ -0,0 +1,83 @@
...
+            format!(".{}", name)
+        }
+        PreSymbolicExpressionType::SugaredFieldIdentifier(ref contract, ref field) => {
+            format!(".{}.{}", contract, field)
+        }
+        PreSymbolicExpressionType::FieldIdentifier(ref trait_id) => {
+            format!("'{}", trait_id)
+        }
+        PreSymbolicExpressionType::TraitReference(ref name) => {
+            println!("trait ref: {}", name);
```
csgui, line 73: Can that be replaced by `format!()` or do we really need to print that information to the console?

csgui (commented):
Hey @tippenein! Well done on this code formatting implementation. I have added some remarks that might be worth checking. Tks!

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
hugocaillard: Beyond just stacking closing parens, I think we don't need to introcude new line before each opening one `(`
If the value was not a tuple (or was a variable name), it would easily fit on one line
```clarity
(parsed-u64 (try! (read-uint64 value)))
```

Here, I'd say we want to split the tuple into multiple lines for readability, but it doesn't imply splitting the rust of the line
```clarity
(parsed-u64 (try! (read-uint64 {
  txbuff: tx,
  index: (+ u1 ptr)
 })))
 ```
 
 
What do you think?

File: components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar (hunk 6 above)
Comment (hugocaillard) [reply components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar:44]: Yes that was my point
```
(ok {
  k1: 1
  k2: 2
})
```

Instead of
(ok
  {
    k1: 1
    k2: 2
  }
)
```

components/clarinet-format/src/formatter/ignored.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    // Handle single-line expressions
+    if start_line == end_line {
+        // Add padding spaces before the expression
+        output.extend(std::iter::repeat(' ').take(start_col - 1));
+        output.push_str(&display_pse_unformatted(expr));
+    } else {
+        // Handle multi-line expressions
+        let expr_str = display_pse_unformatted(expr);
```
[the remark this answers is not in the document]
tippenein: yes, `format-ignore` is unimplemented as of now

hugocaillard (commented):
Some more comments/questions

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+;; Returns (err ERR-LEFTOVER-DATA) if the tx buffer contains leftover data at the end.
+(define-read-only (parse-tx
+		(tx (buff 4096))
+	)
+	(let (
+		(ctx {
+				txbuff: tx,
+				index: u0
+			}
+		)
```
hugocaillard, line 775:
```suggestion
  (ctx {
    txbuff: tx,
    index: u0
  })
```

The opening `(` and `{` are on the same indentation level, so should be the closing one

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+	)
+	(if (> (get tree-depth proof) (len (get hashes proof)))
+			(err ERR-PROOF-TOO-SHORT)
+			(ok
+			(get verified
+				(fold inner-merkle-proof-verify
+					(unwrap-panic
+						(slice? (list u0 u1 u2 u3 u4 u5 u6 u7 u8 u9 u10 u11 u12 u13) u0
+							(get tree-depth proof)
+						)
```
hugocaillard, line 976: There are some places where i think the line splitting logic could be improved.
I think it would be nice to have some sort of rule like "if a function call doesn't fit on one line, all arguments should be split on new ones", instead of the just last one like here.

So instead of
```clarity
(unwrap-panic
  (slice? (list u0 u1 u2 u3 u4 u5 u6 u7 u8 u9 u10 u11 u12 u13) u0
    (get tree-depth proof)
  )
)
```

Something like:
```clarity
(unwrap-panic
  (slice?
    (list u0 u1 u2 u3 u4 u5 u6 u7 u8 u9 u10 u11 u12 u13)
    u0
    (get tree-depth proof)
  )
)
```


--- 

I think more "one-lining" could be done, but this is already being discussed [here](https://github.com/hirosystems/clarinet/pull/1609/files#r1942697216)
```clarity
(unwrap-panic (slice?
  (list u0 u1 u2 u3 u4 u5 u6 u7 u8 u9 u10 u11 u12 u13)
  u0
  (get tree-depth proof)
))
```

hugocaillard: Also the rule "if a function call doesn't fit on one line, all arguments should be split on new ones", does have exception for control flow functions like `match`, `if` where the first arg is on the samel line, but that seems to be working well

hugocaillard (commented):
More golden tests comments

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1183 @@
...
+		(if (<= byte u252)
+				;; given byte is the varint
+				(ok
+				{
+					varint: byte,
+					ctx: {
+						txbuff: tx,
+						index: (+ u1 ptr)
+					}
+				}
```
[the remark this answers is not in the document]
tippenein: ah! Definitely agree. Didn't know that was accepted in clarity.. that's great. Probably easier for me to remove the comma special casing across all the code and then update the tests that fail

components/clarinet-format/tests/golden-intended/alex-transfer.clar
tippenein: We handled this on a call. the double indent is gone now. resolving this issue

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1220 @@
...
+                // no next expression to space out
+                between = "";
+            }
+
+            result.push_str(&format!("{current}{between}"));
+        }
+        result
+    }
+
+    fn format_constant(&self, exprs: &[PreSymbolicExpression]) -> String {
```
hugocaillard, line 215: I tried
```rs
let result = format_with_default(&String::from("(define-fungible-token hello)"));
```

But it seems to fail.
The second param is options in `define-fungible-token`

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1220 @@
...
+            ));
+        }
+
+        acc.push('}');
+        acc
+    }
+
+    // used for (tuple (n1  1)) syntax
+    // Note: Converted to a { a: 1 } style map
+    // TODO: This should be rolled into format_key_value_sugar, but the PSE
```
hugocaillard, line 511: that's alright imo. I feel like in certain context (asts manipulation), a bit a duplication is better than to much abstraction

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1220 @@
...
+
+                // Apply formatting and compare
+                let result = format_file_with_metadata(&src);
+                pretty_assertions::assert_eq!(
+                    result,
+                    intended,
+                    "Mismatch in file: {:?}",
+                    file_name
+                );
+            }
```
hugocaillard, line 1217: Ideally, integration tests should leave in the [`./tests` directory](https://doc.rust-lang.org/book/ch11-03-test-organization.html#the-tests-directory) (close the golden/golden-intended contracts actually)
We don't do a very good job at doin this in the repo, but we aim to do better.

(Only if this is not too involved to move this test)

hugocaillard (changes requested):
Looking very good, I find the code really readable and I think it will be reasonably to maintain.
Left a few comments/suggestions

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1220 @@
...
+        let agg = Aggregator::new(&self.settings, &pse, source);
+        let result = agg.generate();
+
+        // make sure the file ends with a newline
+        result.trim_end_matches('\n').to_string() + "\n"
+    }
+    /// Alias `format_file` to `format`
+    pub fn format(&self, source: &str) -> String {
+        self.format_file(source)
+    }
```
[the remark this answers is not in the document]
tippenein: it was only because of the specific file/section methods. Seemed like format_file should be the default and `format` should be a method included on this

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1220 @@
...
+                            NativeFunctions::TupleCons => {
+                                // if the kv map is defined with (tuple (c 1)) then we strip the
+                                // ClarityName("tuple") out first and convert it to key/value syntax
+                                self.format_key_value(&list[1..], previous_indentation)
+                            }
+                            NativeFunctions::If => self.format_if(list, previous_indentation),
+                            NativeFunctions::And | NativeFunctions::Or => {
+                                self.format_booleans(list, previous_indentation)
+                            }
+                            _ => {
```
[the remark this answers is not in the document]
tippenein: oooo, yea. That's a _big_ list 😆 Let's see how that looks

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1220 @@
...
+                    acc.push(' ');
+                    acc.push_str(&self.display_pse(value, ""));
+                    acc.push(')');
+                }
+            }
+
+            acc.push('\n');
+            acc
+        } else {
+            panic!("Expected a valid constant definition with (name value)")
```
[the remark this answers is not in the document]
tippenein: Yes, I should remove all these panics ideally. I had them in while i was writing it so i'd know which loop was failing

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -0,0 +1,1220 @@
...
+
+        // make sure the file ends with a newline
+        result.trim_end_matches('\n').to_string() + "\n"
+    }
+    /// Alias `format_file` to `format`
+    pub fn format(&self, source: &str) -> String {
+        self.format_file(source)
+    }
+    /// for range formatting within editors
+    pub fn format_section(&self, source: &str) -> String {
```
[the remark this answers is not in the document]
tippenein: We should be able to format selected sections of code with this. I believe we'll need to pre-populate the `previous_indentation` but I'l' at least add a note for that now.

components/clarinet-format/src/formatter/mod.rs
tippenein: Didn't realize that. easy fix. It uses this dumb function called `name_and_args` which i should really remove because it's just asking for trouble

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
tippenein: Ok, I did some hacking around this and it's hard to get something that works both deterministically and with paired closing brackets/parens.

For example we can use this basic idea to match the closing at the end of an expression, but it relies on span information
```
        let mut break_lines = false;
        let start_line = list[0].span().start_line;
        for expr in list.iter() {
            if expr.span().start_line != start_line {
                break_lines = true;
            }
```

That means that we lose determinism.

For example:

```
(try! (ok (list .... ))
;; formatted
(try!
  (ok (list 
    1 2 3...
    )
))
```

will be max-line-length closed differently than 

```
(try! 
  (ok (list 1 2 3..))
)
;; formatted
(try!
  (ok (list 
    1 2 3...
  ))
)


  
  ```

Comment (hugocaillard) [reply components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar:775]: Could there be another approach where we don't consider the spans of the input at all. Or, said differently, we always one-line the input whatever it is
```clarity
;; input 1
(try!
  (ok (list 1 2 3..))
)
;; input 2
(try!
  (ok (list 1 2 3..)))
;; input 3
(try!  (ok (list 1 2 3..)))

 ;; would all be considered like so:
(try! (ok (list 1 2 3..)))
```

Then, the job of the formatter is to insert breaklines where it's needed
And, by keeping track of the parenthesis pairs, put the closing parens on the same indentation level as the opening one.

Although, maybe I'm over simplying this, or what I'm suggesting doesn't really fit in the current design of the formatter.

--- 

Note: in your comment i don't really understand this example
```clar
(try! (ok (list .... ))
;; formatted
(try!
  (ok (list 
    1 2 3...
    )
))
```
Why is the first `)` indented twice?
Why is the second `)` not intended? 
It sounds to me that they are both wrong (when with an extra indent and one with a missing one)

tippenein: Ya, sorry. Bad example.

```
(try! (ok 
  (list 1 2 3
    4 5 6...
  )
))
```

Point was really just about the span checking messing up determinism. We can figure out the closing parens in the case the max-line-length isn't reached, but in that case we have an issue with the current wrapping.

I think you're right though. The current wrapping logic is too simplistic

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1199,6 +1224,33 @@ pub fn main() {
                 process::exit(1);
             }
         },
+        Command::Formatter(cmd) => {
```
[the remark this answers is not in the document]
tippenein: > This command is in beta. Feedback is welcome!

components/clarinet-format/tests/golden-intended/alex-transfer.clar
```diff
@@ -0,0 +1,115 @@
...
+                            ))
+                    ))
+                (ids-to-recipient
+                    (if (is-eq no-to-recipient u0)
+                        (list)
+                        (unwrap-panic
+                            (slice? new-available-ids
+                                (- (len new-available-ids) no-to-recipient)
+                                (len new-available-ids)
+                            ))
```
hugocaillard, line 71:
```suggestion
                            )
                        )
```

The unwrap-panic closing parens should have 1 less indentation level

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -0,0 +1,1119 @@
...
+            (as-max-len?
+              (unwrap! (slice? data base (+ base u1)) (err ERR-OUT-OF-BOUNDS))
+              u1
+            ))
+        ),
+        ctx: {
+          txbuff: data,
+          index: (+ u1 base),
+        },
+      })
```
hugocaillard, line 55: The opening `(` has one less indentation level.
Although it would work to have both on the same

hugocaillard, line 55:
```suggestion
    (ok {
			uint8: (buff-to-uint-le
				(unwrap-panic
					(as-max-len?
						(unwrap! (slice? data base (+ base u1)) (err ERR-OUT-OF-BOUNDS))
						u1
					))
			),
			ctx: {
				txbuff: data,
				index: (+ u1 base),
			},
		})
```

hugocaillard (commented):
Remarks on intended-golden

hugocaillard (approved):
Let's ship it! 🚀

hugocaillard:
Thanks @csgui.
We merged a bit quickly to get it ready for the release 👌

Issue #1503, opened 2024-07-11
Mainnet Execution Simulation


hugocaillard:
## Mainnet Execution Simulation

This feature allows developers to fork the Mainnet state in Clarinet Simnet.

Use cases:

- When Clarinet projects rely on deployed contracts, such as DeFi protocols or oracles, it can be challenging to populate these protocols with real-world data. This feature enables the use of Mainnet data in unit tests or the Clarinet console.
- It also enables the simulation of a Mainnet transaction before actually running it.

This feature aims to be data-efficient and only fetches the necessary marf data and metadata.

This specification outlines the work required in the following three main areas:

- [Clarinet](#clarinet)
- [Stacks RPC API](#new-stacks-rpc-endpoints)
- [DevOps for API keys and caching](#stacks-hiro-api-rate-limiting-and-caching)

### Clarinet

This feature impacts the Clarinet datastore, which implements the [ClarityBackingStore trait](https://github.com/stacks-network/stacks-core/blob/02b6a05b1bda21dc7fe490645aa9cd277898842e/clarity/src/vm/database/clarity_store.rs#L62). When the Simnet attempts to read data from the store, it first checks if it is available locally. If not, it fetches the data from the network and writes the result locally. The desired chaintip (or block hash) will be set in the project manifest, allowing for consistent data retrieval from the network with efficient caching.

Clarinet will rely on an API key or auth token set in `~/.clarinet/clarinetrc.toml` or an environment variable. This can also be set using the `clarinet login` command and integration with the Platform. However, this can be done at a later stage.

New manifest settings (Clarinet.toml) (naming to be confirmed):

- `[simnet]`
  - `fork_network: boolean (false)`
  - `network_api_url: string (https://api.hiro.so)`
  - `chaintip: number (none)`

New command:

- `clarinet simnet enable-mainnet-fork` to configure the simnet settings and set the chaintip to the current height

### New Stacks RPC endpoints

To enable the simnet Datastore to read data from a specific Stacks network, the Stacks nodes need to expose two new RPC endpoints:

- `/v2/clarity_marf_value/:clarity_marf_key`
  - Takes an optional `tip` query parameter
  - Takes an optional `proof` query parameter (defaults to `false` in this context)

- `/v2/clarity_metadata/:principal/:contract_name/:clarity_metadata_key`
  - Does not take a `tip` query parameter
  - Does not take a `proof` query parameter

### Stacks Hiro API rate limiting and caching

[...]

PR #1622, opened 2024-12-06
feat: mainnet execution simulation

hugocaillard:
### Description

Resolve: #1503 

### Example

---

### Checklist


components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -22,6 +22,7 @@ use clarinet_deployments::types::{DeploymentGenerationArtifacts, DeploymentSpeci
 use clarinet_deployments::{
     get_default_deployment_path, load_deployment, setup_session_with_deployment,
 };
+use clarinet_files::clarinetrc::ClarinetRC;
```
hugocaillard, line 25: Simply renamed `GlobalSettings` into `ClarinetRC`

components/clarinet-deployments/src/lib.rs
```diff
@@ -106,6 +106,9 @@ fn update_session_with_genesis_accounts(
     deployment: &DeploymentSpecification,
 ) {
     if let Some(ref spec) = deployment.genesis {
+        let addresses: Vec<_> = spec.wallets.iter().map(|w| w.address.clone()).collect();
+        session.interpreter.save_genesis_accounts(addresses);
```
hugocaillard, line 110: `genesis_accounts` are saved in the store. It allows to skip remote_data fetching for those addresses (big perf gain on session start)

components/clarinet-deployments/src/lib.rs
```diff
@@ -339,21 +344,28 @@ pub async fn generate_default_deployment(
     let mut requirements_data = BTreeMap::new();
     let mut requirements_deps = BTreeMap::new();
 
+    let mut repl_settings = manifest.repl_settings.clone();
+    repl_settings.remote_data.enabled = false;
```
hugocaillard, line 348: Might need to find a better way to do it.
But we basically want to fully disable remote_data fetching `clarinet check` and other static analyses.
We basically only want remote_data in the repl context (`clarinet console`, js sdk sessions, etc)

components/clarinet-sdk/node/tests/remote-data.test.ts
```diff
@@ -0,0 +1,70 @@
...
+  deleteExistingDeploymentPlan();
+});
+
+describe("simnet remote interactions", async () => {
+  const simnet = await getSDK();
+
+  it("can call a remote contract", async () => {
+    await simnet.initEmptySession({
+      enabled: true,
+      api_url: "https://api.testnet.hiro.so",
```
hugocaillard, line 39: In this PR we are running some tests again testnet... Which doesn't sounds like a very good idea. But it is convenient.
Other options would be to
- test against a local devnet (but we know that's annoying to do in github actions)
- test against mainnet might actually be better, since tesnet can be reset at anytime

components/clarity-repl/src/uprint.rs
```diff
@@ -0,0 +1,17 @@
+// universal print macro
```
hugocaillard, line 1: print help that works in wasm and non-wasm

components/clarinet-sdk/node/tests/remote-data.test.ts
```diff
@@ -0,0 +1,70 @@
...
+      initial_height: 57000,
+    });
+    const { result } = simnet.callReadOnlyFn(
+      counterAddress,
+      "get-count-at-block",
+      [Cl.uint(56230)],
+      sender,
+    );
+    expect(result).toStrictEqual(Cl.ok(Cl.uint(0)));
```
obycode, line 68: This might be more interesting to check at block 56233 or somewhere that it is not zero.

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -351,6 +352,17 @@ struct Console {
         conflicts_with = "use_on_disk_deployment_plan"
     )]
     pub use_computed_deployment_plan: bool,
+
+    /// Enable remote data fetching from mainnet or a testnet
+    #[clap(long = "enable-remote-data", short = 'r')]
+    pub enable_remote_data: bool,
+    /// Set a custom Hiro API URL for remote data fetching
```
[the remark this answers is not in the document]
hugocaillard: It needs to be "Hiro's Stacks Blockchain API" and not just the stacks-node RPC endpoints.
But you're right, we just call it the "Stacks Blockchain API", I'll drop the Hiro

components/clarinet-sdk/node/tests/remote-data.test.ts
hugocaillard: I thout it was interested to check somewhere where the value is different than the one at the initial height (57000), but I'll check both 👍

obycode: Nice. Yeah, I think it's interesting to check both.

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -328,9 +328,13 @@ impl ClarityDatastore {
     }
 
     fn fetch_clarity_marf_value(&mut self, key: &str) -> Result<Option<String>> {
-        let key_hash = TrieHash::from_key(key);
+        let key_hash = TrieHash::from_key(dbg!(key));
```
hugocaillard, line 331: Shouldn't have pushed that, will remove

obycode (approved):
This is great! 🙌

tippenein (approved):

PR #1640, opened 2025-01-14
chore: Apply Clippy lints  `redundant_clone` and  `clear_with_drain`

jbencin:
### Description

I've been doing some code cleanup on stacks-core, so I thought I'd try it on this repo also. This repo is already configured to use Clippy, which is great, but by default Clippy does not apply it's "nursery" (experimental) lints. Some of these can be useful and are worth applying. Specifically, I checked the following:

```
clear_with_drain
collection_is_never_read
needless_collect
iter_with_drain
mutex_integer
set_contains_or_insert
trivial_regex
useless_let_if_seq
unused_rounding
unnecessary_struct_initialization
tuple_array_conversions
```

Only `redundant_clone` and `clear_with_drain` resulted in any changes

#### Breaking change?

No

### Example

This can make the code more clear and improve performance by removing unnecessary function calls. For example, one change I made was:

```rust
fn level_to_string(level: &Level) -> String {
    match level {
        Level::Note => blue!("note:").to_string(),
        Level::Warning => yellow!("warning:").to_string(),
        Level::Error => red!("error:").to_string(),
    }
}
```

to

```rust
fn level_to_string(level: &Level) -> String {
    match level {
        Level::Note => blue!("note:"),
        Level::Warning => yellow!("warning:"),
        Level::Error => red!("error:"),
    }
}
```

### Checklist

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -510,7 +510,7 @@ pub fn main() {
                 }
             };
             clap_complete::generate(cmd.shell, &mut app, "clarinet", &mut file);
-            println!("{} {}", green!("Created file"), file_name.clone());
+            println!("{} {}", green!("Created file"), file_name);
             println!("Check your shell's documentation for details about using this file to enable completions for clarinet");
         }
         Command::New(project_opts) => {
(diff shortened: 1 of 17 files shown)


hugocaillard:
@jbencin thaznks for trying it on Clarinet as well 🙌 💯 

What are the benefits/risks of running the experimental lints?
I see that you didn't change any config, so I assume you ran a custom command to get the lints? Should also change a config that would run clippy experimental features so that we don't keep introducing code that break any of these rules?

jbencin:
> What are the benefits/risks of running the experimental lints?

They are just less reliable, so sometimes they miss things, or find false positives. They are still useful, it's just more of a manual process instead of `cargo clippy --fix`

> Should also change a config that would run clippy experimental features so that we don't keep introducing code that break any of these rules?

Personally I wouldn't recommend it because of the false positives. There are still a couple incorrect `redundant_clone` violations that Clippy finds (I was using Rust/Clippy 1.83)

hugocaillard (approved):
lgtm! thx @jbencin

PR #1649, opened 2025-01-23
fix: Windows specific path problem

### LINKED ISSUE #1634: Absolute manifest path cannot be used to initialize a `simnet` session on Windows
**Describe the bug**
An unexpected error is thrown while attempting to initialize the `simnet` using an absolute manifest path on Windows. The error message is _"unable to mutate url"_.

**To Reproduce**
1. On a Windows machine, create a minimal Node.js app having the following `index.js`:
```js
import { initSimnet } from "@hirosystems/clarinet-sdk";
import { resolve } from "path";

const main = async () => {
  const manifestDir = process.argv[2];

  const manifestPath = resolve(manifestDir, "Clarinet.toml");

  console.log(manifestPath);
  await initSimnet(manifestPath);
};

main();
```
2. Add a Clarinet project called `example` at the root of the node app.
3. Run `index.js` using the following command:
```sh
node index.js example
```
4. See error:
```
$ node index.js example
C:\Users\...\bug-demo\example\Clarinet.toml
node:internal/process/promises:389
      new UnhandledPromiseRejection(reason);
      ^

UnhandledPromiseRejection: This error originated either by throwing inside of an
 async function without a catch block, or by rejecting a promise which was not h
andled with .catch(). The promise rejected with the reason "unable to mutate url
".
    at throwUnhandledRejectionsMode (node:internal/process/promises:389:7)
    at processPromiseRejections (node:internal/process/promises:470:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:96:3
2) {
  code: 'ERR_UNHANDLED_REJECTION'
}

Node.js v20.18.1
```

 - OS: Windows

**Additional Context**
The same setup works on Linux, macOS, and Debian.

tippenein:
fixes #1634

I need to test this on a windows machine, but I believe if we normalize the windows format before getting the path we should be ok. Added some tests which may be useful..

### DIFF
--- a/components/clarinet-files/src/lib.rs
+++ b/components/clarinet-files/src/lib.rs
@@ -101,6 +101,13 @@ impl FileLocation {
     }
 
     pub fn from_url_string(url_string: &str) -> Result<FileLocation, String> {
+        // Handle Windows-style paths specially
+        if url_string.contains('\\') || url_string.contains(':') {
+            let path = PathBuf::from(url_string);
+            if path.exists() {
+                return Ok(FileLocation::FileSystem { path });
+            }
+        }
         let url = Url::from_str(url_string)
             .map_err(|e| format!("unable to parse {} as a url\n{:?}", url_string, e))?;
 
@@ -129,15 +136,26 @@ impl FileLocation {
                 path.extend(&path_to_append);
             }
             FileLocation::Url { url } => {
+                // Handle Windows paths first
+                if url.scheme() == "file" && url.path().starts_with("/C:") {
+                    let new_path = url.path().trim_start_matches('/').to_string();
+                    url.set_path(&new_path);
+                }
                 let mut paths_segments = url
                     .path_segments_mut()
                     .map_err(|_| "unable to mutate url".to_string())?;
+
+                // Clear existing empty segments at the end
+                paths_segments.pop_if_empty();
+
                 for component in path_to_append.components() {
                     let segment = component
                         .as_os_str()
                         .to_str()
                         .ok_or(format!("unable to format component {:?}", component))?;
-                    paths_segments.push(segment);
+                    if !segment.is_empty() {
+                        paths_segments.push(segment);
+                    }
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


components/clarinet-files/src/lib.rs
```diff
@@ -101,6 +101,13 @@ impl FileLocation {
     }
 
     pub fn from_url_string(url_string: &str) -> Result<FileLocation, String> {
+        // Handle Windows-style paths specially
+        if url_string.contains('\\') || url_string.contains(':') {
+            let path = PathBuf::from(url_string);
+            if path.exists() {
+                return Ok(FileLocation::FileSystem { path });
+            }
+        }
```
hugocaillard, line 110: I think there's an issue with this code.
`from_url_string` kind of assume that we are passing a valid `url` and that we should return a `FileLocation:Url`.
If not, `from_path_string` is called.
And Url should never be "windows style"

Is there an issue in how we handle it [here](https://github.com/hirosystems/clarinet/blob/faa63a89749e714ba33770bd8598a0198c1cbfdd/components/clarinet-deployments/src/types.rs#L981-L985)?

One other approach could be to have a generic `From<String>` that just sort of handle it.
```
impl From<String> for FileLocation {
    fn from(location_string: String) -> Self {
        if let Ok(location) = FileLocation::from_url_string(&location_string) {
            return location;
        }
        if let Ok(location) = FileLocation::from_path_string(&location_string) {
            return location;
        }
        panic!("Invalid location string: {}", location_string);
    }
}
```
Could be a good opportunity to implement `From<PathBuf>` and `From<Url>` and get a bit more idomatic (instead of the custom functions that we have now)


The point is that Path adn PathBuf should be platform agnostic and we shouldn't have to deal with it, looks like it will get even more error prone.

hugocaillard (changes requested):
I think we may be able to avoid (as much as possible) to write specific code for the different platform (macos/linux/windows) – dealing with native/wasm is already being error prone.
Also, the original issue #1634 seems to highlight that it also break in a wasm context (otherwise, Clarinet would be fully broken on Windows).

I udpated the CI so that we run the SDK unit tests on all platforms (#1650) - still working on that.

It should allow use to see the error in #1634 (or we should add a test that triggers it).

And we should print the actual error [here](https://github.com/hirosystems/clarinet/blob/1f454d3c57056448e2d05dbd06d18fae663beb9b/components/clarinet-files/src/lib.rs#L134) and [there](https://github.com/hirosystems/clarinet/blob/1f454d3c57056448e2d05dbd06d18fae663beb9b/components/clarinet-files/src/lib.rs#L269) to get a better understanding of what's happening.

Still, I think this PR is heading in the right direction, FileLocation needs some improvements to be more robust and less error prone

PR #1668, opened 2025-02-06
fix: burn-block-height in at-block at epoch 3.x

hugocaillard:
### Description

Preserve burn-block-height behavior in 3.x
https://github.com/stacks-network/stacks-core/pull/5524

### DIFF
--- a/components/clarity-repl/src/repl/datastore.rs
+++ b/components/clarity-repl/src/repl/datastore.rs
@@ -989,6 +989,15 @@ impl BurnStateDB for Datastore {
     }
 
     fn get_tip_burn_block_height(&self) -> Option<u32> {
+        // preserve the 3.0 and 3.1 special behavior of burn-block-height
+        // https://github.com/stacks-network/stacks-core/pull/5524
+        if matches!(
+            self.current_epoch,
+            StacksEpochId::Epoch30 | StacksEpochId::Epoch31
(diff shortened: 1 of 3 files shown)


components/clarity-repl/src/repl/datastore.rs
```diff
@@ -989,6 +989,15 @@ impl BurnStateDB for Datastore {
     }
 
     fn get_tip_burn_block_height(&self) -> Option<u32> {
+        // preserve the 3.0 and 3.1 special behavior of burn-block-height
+        // https://github.com/stacks-network/stacks-core/pull/5524
+        if matches!(
+            self.current_epoch,
+            StacksEpochId::Epoch30 | StacksEpochId::Epoch31
```
jbencin, line 996: Given that this is not fixed in stacks-core yet, this should be:

```rust
self.current_epoch >= StacksEpochId::Epoch30
```

Assuming you have `PartialEq` implemented for `StacksEpoch`

hugocaillard: I was assuming that it would be fixed in 3.2 😁
But that's fair. I'll do what Brice said on the other PR, a match with no wildcare, so I'll must check this code path when a new epoch is added

jbencin (approved):
LGTM

PR #1675, opened 2025-02-13
refactor: remove sync-request from sdk

hugocaillard:
### Description

wip


components/clarity-repl/src/repl/remote_data/http_request/browser.rs
```diff
@@ -0,0 +1,42 @@
+use serde::de::DeserializeOwned;
+use web_sys::XmlHttpRequest;
+
+pub fn http_request<T: DeserializeOwned>(url: &str) -> Result<T, String> {
+    let xhr = XmlHttpRequest::new().map_err(|e| format!("Failed to create XHR: {:?}", e))?;
+
+    xhr.open_with_async("GET", url, false)
```
tippenein, line 7: so weird that `open_with_async` is sync if you pass the 3rd false 🤯

tippenein (approved):
Looks good for a replacement hack 👍

components/clarity-repl/src/repl/remote_data/http_request/browser.rs
hugocaillard: Yes that's how rust bindings have to deal with JS APIs with optional params. In the case of xhr.open() the third argument is optional, and defaulting to true

PR #1678, opened 2025-02-14
ci: run sdk tests in node 20 and 22

hugocaillard:
### Description

Run clarinet-sdk tests in node 20 and 22


hugocaillard:
Confirmed that the tests fail in Node 20.
Removing this version from this PR to re-injtroduce in #1675

PR #1700, opened 2025-02-27
formatter caching

tippenein:
simple caching to speed up benchmarks

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -7,7 +7,15 @@ use clarity::vm::functions::{define::DefineFunctions, NativeFunctions};
 use clarity::vm::representations::{PreSymbolicExpression, PreSymbolicExpressionType};
 use helpers::{name_and_args, t};
 use ignored::ignored_exprs;
-use std::fmt;
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::{fmt, slice};
+
+// A thread-local cache to avoid RefCell borrowing issues
+thread_local! {
+    static EXPR_CACHE: RefCell<HashMap<(usize, String), String>> = RefCell::new(HashMap::new());
+    static LIST_CACHE: RefCell<HashMap<(usize, String), String>> = RefCell::new(HashMap::new());
+}
 
 pub enum Indentation {
     Space(usize),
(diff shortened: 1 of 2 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -281,16 +300,12 @@ impl<'a> Aggregator<'a> {
                             DefineFunctions::Constant
                             | DefineFunctions::PersistedVariable
                             | DefineFunctions::FungibleToken
+                            | DefineFunctions::ImplTrait
+                            | DefineFunctions::UseTrait
                             | DefineFunctions::NonFungibleToken => {
                                 self.constant(list, previous_indentation)
                             }
                             DefineFunctions::Map => self.format_map(list, previous_indentation),
-                            DefineFunctions::UseTrait | DefineFunctions::ImplTrait => {
```
tippenein, line 288: these were the same as `constant` anyway, so lumped em together

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -376,36 +395,29 @@ impl<'a> Aggregator<'a> {
...
+        let mut acc = format!("({func_type} ");
         let indentation = &self.settings.indentation.to_string();
-        let space = format!("{}{}", previous_indentation, indentation);
+        let space = format!("{}{}", indentation, previous_indentation);
+        acc.push_str(&self.format_source_exprs(slice::from_ref(&exprs[1]), previous_indentation));
+        let mut iter = exprs[2..].iter().peekable();
+        while let Some(expr) = iter.next() {
+            let trailing = get_trailing_comment(expr, &mut iter);
 
-        if let Some((name, args)) = name_and_args(exprs) {
```
tippenein, line 384: this was overcomplicated and error-prone. I replaced it with this boilerplate-y mess, but we can abstract this away later

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -843,22 +877,28 @@ impl<'a> Aggregator<'a> {
...
-                if last_line.trim() == ")" || last_line.trim() == "}" {
-                    // Don't add a newline if we're already looking at a closing paren/bracket
-                    false
-                } else {
-                    true
-                }
-            } else {
-                true
-            }
+        let break_lines = if !result.contains('\n') {
```
tippenein, line 880: no substantive change, just more clear and concise

hugocaillard (commented):
Awesome work on the perf 🙌 

I'm wondering if the CACHEs need to be top level RefCell?
Could it be stored in the Aggregator directly?
```
pub struct Aggregator {
    ...,
   expr_cache: HashMap<(usize, String),
   list_cache: HashMap<(usize, String),
}
```

Or what are benefits with the refcell approach here?

tippenein:
@hugocaillard simplified the cache a bit. Also removed the redundant segment of raw source from clarity-bitcoin since we already test that.

hugocaillard (approved):
lgtm ⚡

PR #1702, opened 2025-02-28
fix: make sbtc funding on devnet more robust

hugocaillard:
### Description

The previous logic was picking an arbitrary bitcoin block as an argument to pass to the mint call.
It was very likely that whatever this height was, it was skipped (because a lot of bitcoin blocks are skipped around epoch 3.0 switch).
The PR just takes the latest bitcoin block, instead of an arbitrary one

### DIFF
--- a/components/stacks-network/src/chains_coordinator.rs
+++ b/components/stacks-network/src/chains_coordinator.rs
@@ -898,12 +898,20 @@ fn fund_genesis_account(
             let funding_result = stacks_rpc.post_transaction(&tx);
             deployer_nonce += 1;
 
-            if let Err(e) = funding_result {
-                let _ = devnet_event_tx_moved.send(DevnetEvent::error(format!(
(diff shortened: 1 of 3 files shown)


components/stacks-network/src/chains_coordinator.rs
```diff
@@ -817,14 +815,23 @@ fn fund_genesis_account(
 
     let _ = hiro_system_kit::thread_named("sBTC funding handler").spawn(move || {
         while !boot_completed_moved.load(Ordering::SeqCst) {
-            std::thread::sleep(std::time::Duration::from_secs(1));
+            std::thread::sleep(std::time::Duration::from_secs(3));
```
hugocaillard, line 818: This is probably not bullet prof, but I found 3 seconds to work reliably

tippenein (dismissed):
lgtm

PR #1704, opened 2025-03-03
refactor: rework wasm conditional compilation

hugocaillard:
## Context

Because it compiles to both native and wasm platform, clarinet uses conditional dependency management and compilation.
I could make dependency management mrpoe difficult than it should be

## Description

Instead of managing conditional dependencies with cargo feature, we can use cargo target dependencies instead.

So instead of this
```toml
# package-a
[dependencies]
clap = { optional = true }

[features]
default = ["cli"]
cli = ["clap"]
wasm  = []

##########
#package-b
[dependencies]
package-a = { path = "../package-a", optional = true }

[features]
default = ["cli"]
cli = ["package-a/cli"]
wasm  = ["package-a/cli"]
```

We now do this

```toml
# package-a
[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
clap = { optional = true }

##########
#package-b
[dependencies]
package-a = { path = "../package-a" }
```

With this feature, we remove 100 of toml config.

## Future work

The idea of simplifying this depedency management topic on the repo, is mainly to famalirise myself with it so that I can use it in stacks-core to have a clean of handling clarity-vm compilation to wasm

### DIFF
--- a/components/clarinet-deployments/Cargo.toml
+++ b/components/clarinet-deployments/Cargo.toml
@@ -5,42 +5,27 @@ edition = "2021"
 
 [dependencies]
 colored = { workspace = true }
+base64 = "0.21.3"
 serde = "1"
 serde_json = "1.0.123"
 serde_derive = "1"
 serde_yaml = "0.8.23"
 rand = "0.8.5"
-
-clarity-repl = { path = "../clarity-repl", default-features = false, optional = true }
-clarinet-files = { path = "../clarinet-files", default-features = false }
-stacks-rpc-client = { path = "../stacks-rpc-client", optional = true }
-clarinet-utils = { path = "../clarinet-utils", optional = true }
-stacks-codec = { path = "../stacks-codec", optional = true }
-
-# CLI
 reqwest = { workspace = true }
-bitcoin = { version = "0.31.2", optional = true }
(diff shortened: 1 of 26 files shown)


components/clarinet-deployments/Cargo.toml
```diff
@@ -5,42 +5,27 @@ edition = "2021"
...
-base64 = "0.21.3"
-libsecp256k1 = { version = "0.7.0", optional = true }
-
 clarity = { workspace = true }
 
-[features]
-default = ["cli"]
-cli = ["clarity-repl/sdk", "clarinet-files/cli", "stacks-codec", "onchain"]
-wasm = ["clarity-repl/wasm", "clarinet-files/wasm"]
-onchain = [
```
hugocaillard, line 35: onchain was never ever used with cli.
And wouldn't even compile without it
```sh
$ cargo check -p clarinet-deployments --no-default-features --features=onchain 
> errors
```

tippenein (approved):

hugocaillard:
@tippenein 
The unreachables were network issues in the unit tests (possibly rate limits) because some tx simulation integration tests call the hiro testnet api. (and also because it badly handles network errors)
I'll add an API key to increase the rate limit in ci

PR #1710, opened 2025-03-13
fix: improve devnet sbtc handling

hugocaillard:
### Description

Because the sbtc contract can be deployed at the end of a deployment plan, is possible that the `boot_completed` is set to true while the sbtc contracts are still in the mempool. So when they are deployed (after boot is completed), the sbtc funding tx were not triggered.
This PR fixes it.

it also tweaks some devnet timing value, that are promising to get a faster and stable devnet start to epoch 3.0


tippenein (approved):
LGTM

PR #1713, opened 2025-03-13
chore: update dependencies

hugocaillard:
### Description

Update dependencies.
- Removed prettytable in favor of comfy table (actively maintained, better api)
- Updated dependencies that were bringing old versions of nix or windows-sys
- Update many other dependencies

On mac, dependency number goes from 713 to 658 (mostly due to dedup)
Eventually, will we get this number to go down as we avoid unecessary dependencies and switch to more mature dependencies

No change in behavior.


tippenein (approved):
lgtm

PR #1715, opened 2025-03-14
feat: LSP formatting

tippenein:
- [x] adds document formatting and section formatting
- [x] changes to the `format_section`

https://github.com/user-attachments/assets/4595a7bd-fe5c-4f94-8241-d9a3a09e9f2d

By default format-on-save is disabled. The clarity-lsp config section can be changed to enable it

Screenshot 2025-03-28 at 9 36 45 AM


hugocaillard (changes requested):
@tippenein 
The native_bridge should have the same capabilities.
Even though it's not well maintained is has some stabiltiy issues, we don't want it to fall too far behind

hugocaillard:
@tippenein 
If this PR enables formatting on save, we might want to wait for the 3.0 release to broadly announce it

PR #1716, opened 2025-03-17
fix extra useless newlines and add test

tippenein:
let simple expressions continue on same line as opposed to newlining every expr break.

Test case added to show the intended format from here: https://github.com/hirosystems/clarinet/discussions/1689#discussioncomment-12328285


components/clarinet-format/tests/golden-intended/nested_map.clar
```diff
@@ -13,19 +13,18 @@
...
         namespace: namespace,
-      },
-      id: id-to-be-minted,
-      properties: (map-get? name-properties
-        {
-          name: name,
-          namespace: namespace,
-        }),
-    })
+      }),
```
hugocaillard, line 28: The closing `)` of `map-get?` isn"t on the same indentation leval as the opening one.
Should be
```clarity
  properties: (map-get? name-properties
    {
      name: name,
      namespace: namespace,
    }
  )
```
or even better imo
```clarity
```clarity
  properties: (map-get? name-properties {
    name: name,
    namespace: namespace,
  })
```

hugocaillard (changes requested):

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -834,8 +835,42 @@ impl<'a> Aggregator<'a> {
         let indentation = self.settings.indentation.to_string();
         let base_indent = format!("{}{}", previous_indentation, indentation);
 
+        let is_map_get = list.len() > 2
```
hugocaillard, line 838: There variable is unused

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -834,8 +835,42 @@ impl<'a> Aggregator<'a> {
...
+                let fn_name =
+                    self.format_source_exprs(slice::from_ref(&list[0]), previous_indentation);
+                let arg = self.format_source_exprs(slice::from_ref(&list[1]), previous_indentation);
+
+                return format!("({} {})", fn_name.trim(), arg.trim());
+            }
+        }
         // TODO: this should ignore comment length
-        for expr in list.iter() {
+        for (i, expr) in list.iter().enumerate() {
```
hugocaillard, line 873: Looks like `i` is unused here, do we need `enumerate()`?

hugocaillard (changes requested):

hugocaillard (approved):
lgtm 🚀

PR #1718, opened 2025-03-21
ci: update cargo audit and clippy in ci

hugocaillard:
### Description

Update clarinet CI.
Run cargo clippy

I removed the `audit` worklflow to only have it in the main ci workflow.

Note:

the `jupyter-kernel`  components still exists but has been removed from the workspace. No need to exclude it anymore


tippenein (approved):
LGTM

PR #1719, opened 2025-03-21
fix: fix trait references

tippenein:


components/clarinet-cli/src/lsp/native_bridge.rs
```diff
@@ -58,7 +58,7 @@ pub async fn start_language_server(
                         LspRequest::Initialize(_) => {
                             process_mutating_request(request, &mut editor_state)
                         }
-                        _ => process_request(request, &editor_state),
+                        _ => process_request(request, &editor_state).await,
```
hugocaillard, line 61: ooouh interesting. does it fully fixes the native bridge?

hugocaillard (dismissed):
lgtm 🙌

components/clarinet-cli/src/lsp/native_bridge.rs
tippenein: lol. this wasn't supposed to be commited 😓

hugocaillard (approved):

PR #1720, opened 2025-03-21
chore: fmt single line for functions with 1 arg

tippenein:
- [x] single line for function with 1 arg 
- [x] single line for let with 1 arg


hugocaillard (approved):
lgtm!

PR #1721, opened 2025-03-21
fix: format comments within map properly

tippenein:
- [x] adds parse phase to golden tests to prevent this issue in the future
- [x] fix contents of map values with newlined comments


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1450,6 +1487,52 @@ mod tests_formatter {
...
+    #[test]
+    fn weird_nesting() {
+        let src = r#"(merge name-props {
+  something: u1,
+  ;; comment
+  renewal-height:
+    ;; If still within lifetime, extend from current renewal height; otherwise, use new renewal height
+    (if (< burn-block-height
+        (unwrap-panic (get-renewal-height (unwrap-panic (get-id-from-bns name namespace))))
+      )
```
hugocaillard, line 1499: Shouldn't this two line have 1 less indentation level?

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1450,6 +1487,52 @@ mod tests_formatter {
...
+  (map-set name-properties {
+    name: name,
+    namespace: namespace,
+  }
+    (merge name-props {
+      renewal-height:
+        ;; If still within lifetime, extend from current renewal height; otherwise, use new renewal height
+        (if (< burn-block-height
+            (unwrap-panic (get-renewal-height (unwrap-panic (get-id-from-bns name namespace))))
+          )
```
hugocaillard, line 1523: Same here

hugocaillard (changes requested):
Left a comment, although I'm not sure this is specificly related to this PR

components/clarinet-format/src/formatter/mod.rs
tippenein: it's because it's a continuation of the `if` conditional. It does look weird from the break line, but what would be the alternative?

hugocaillard: An alternative would be
```clarity
(if (<
    burn-block-height
    (unwrap-panic (get-renewal-height (unwrap-panic (get-id-from-bns name namespace))))
)
    (+
        (unwrap-panic (get-renewal-height (unwrap-panic (get-id-from-bns name namespace))))
        lifetime
    )
    new-renewal-height
)
```

Two things here:
1. Regarding the indentation, I think this is the "strict way", closing parenthesis on the same indentation level as the opening one. We agreed to have a "cosmetic extra indentation level" in the paramaters of `define-functions`. But I don't know if we want to have it in other places.
2. In the `<` function call, if the arguments need to be on multiple lines, I think both should be on new lines.

Also: we exceed the 80 characters limits here right?

The original source code of the file definitely makes it hard for the formatter to do its job properly (like this comment after `renewal-height::` is definitely not helping, and the nested unwrap-panic...)
But it makes it a good case study I guesss

hugocaillard (approved):
lgtm.
Let's address [this comment](https://github.com/hirosystems/clarinet/pull/1721/files#r2014327485) in future Rs

PR #1727, opened 2025-03-27
ci: update gh

hugocaillard:
### Description

Need to double check the permissions that are set
The winget flow needs rework as well


bgridley (approved):
LGTM

PR #1732, opened 2025-04-01
feat: cache simnet remote metadata

hugocaillard:
### Description

In the context of mainnet execution simulation (MXS), the simnet fetches a lot of data. This data is immutable and can easily be cached.
This PR only stores the metadata, making the MXS already way faster.
In future work, we could also cache all of the marf values or add an option to do so.

The FileSystem handling strategy, especially for wasm, is different from other approaches existing in the repo. But it's simpler and should be the proferred strategy from now on
- until now, we were passing a WASMFileAccessor from JS to Rust to handle fs operation. It was trciky to get this down to the datastore, and we needed the FS to be handle synchronously. 
- we are instead directly:
  - directly accessing node fs in the context of ndoe.js
  - or setting up a globally avaialble vfs() function in the context of web browser.
  
See this code: https://github.com/hirosystems/clarinet/blob/a5866c80479cbd0807556e5a800b4e0ad0ffd7c1/components/clarity-repl/src/repl/remote_data/fs/mod.rs#L6-L14


Cargo.toml
```diff
@@ -28,7 +28,7 @@ pox-locking = { git = "https://github.com/stacks-network/stacks-core.git", branc
 
 colored = { version = "3" }
 crossterm = { version = "0.28" }
-ratatui = { version = "0.29.0" }
+ratatui = { version = "0.29.0", default-features = false, features = ["crossterm"] }
```
hugocaillard, line 31: remove a cargo warning

components/clarity-repl/src/repl/remote_data/fs/node.rs
```diff
@@ -0,0 +1,39 @@
...
+    #[wasm_bindgen(js_name = writeFileSync)]
+    fn write_file_sync(path: &str, data: &[u8]);
+}
+
+pub fn get_file_from_cache(cache_location: &Path, name: &Path) -> Option<String> {
+    let cache_path = cache_location.join(name);
+    if !exists_sync(cache_path.to_str().unwrap()) {
+        None
+    } else {
+        Some(String::from_utf8(read_file_sync(cache_path.to_str().unwrap())).unwrap())
```
tippenein, line 22: The double unwrap might be fine. Is this just assumed to fail with a generic read_file_sync error?

tippenein (approved):
Looks good! This seems like it’ll be useful for other optimizations for sure

components/clarity-repl/src/repl/remote_data/fs/node.rs
hugocaillard: I reworked it with a few less unwrap, that was too mucn.

> Is this just assumed to fail with a generic read_file_sync error?

Not sure to understand the question.
But yes, i think it fine to simply panic and throw the fs error

tippenein: Cool 
sorry, this was meant to be non-blocking, but the change looks good 👍

tippenein (approved):

PR #1736, opened 2025-04-02
ci: publish clarinet-sdk packages to npm

hugocaillard:
### Context

The ci.yaml already crates a release on github for the clarinet-cli.
When the cli release is triggered, we can publish the clarinet-sdk package to npm.

### Questions

Ideally, we only want to to run after ci-sdk.yaml has run and is valid. But I'm not sure how to implement. We could even merge both


hugocaillard:
I remove the caching step from the npm release build because:
- it doesn't run on PR, but on only on release, the speed doesn't matter much
- the job doesn't run often, so the previous job caches are very likely to be expired between two releases

.github/workflows/ci-sdk-release.yaml
```diff
@@ -0,0 +1,48 @@
...
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Build sdk package
+        run: npm run build:sdk
+
+      - name: Publish sdk package
+        run: npm publish -w components/clarinet-sdk/node -w components/clarinet-sdk/browser --provenance --access public
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
deantchi, line 48: Nit: lets add a new line after this.

deantchi (changes requested):
Great work @hugocaillard! Just one tiny change.

.github/workflows/ci-sdk-release.yaml
hugocaillard: There's one already.
GitHub shows a warning when there isn't

hugocaillard:
I rebased the branch on main and forced push, but didn't malke any other change

deantchi (approved):
LGTM

PR #1738, opened 2025-04-03
ci: improve caching

hugocaillard:
### Description

- Refined cached directories
- Improve caching key strategies
  - Use `${{ matrix.os }}-${{ matrix.target }}` or `${{ runner.os }}` as the base of the cache key
  - Include `${{ github.job }}` in all cache keys
  - Use a hash of `rustc -vV`
- Run unit tests on linux, macos, windows
- Also remove useless steps and refactor the `needs` config to improve parallelization
- Don't include stacks-devnet-js in the tests
- Prevent "incremental build" and "debug info" in tests, speeding up the compilation while reducing the cache size

### Results

Previous CI duration was around 45min, with cache ([summary](https://github.com/hirosystems/clarinet/actions/runs/14224650357))

About 20min with no cache ([summary](https://github.com/hirosystems/clarinet/actions/runs/14263808460/attempts/1))
About 17min with cache ([summary](https://github.com/hirosystems/clarinet/actions/runs/14263808460/attempts/2))

On this benchmarks, by no cache I mean no cache on the cargo_test steps, they are the one that take a significant amount of time and are blocking the "build docker image" step that is quite long as well. So it's a good way to measure full CI duration with no cache (because the other jobs will run in paralell of those, and take less time)

### One more thing

The last step of the CI is to build the Docker image, it requires the previous steps to be finished, and takes up to 10min because the clarinet binary is being built again, in the Docker image directly.
We can instead just use the artifact built in previous steps, and build the image in 30 seconds instead of 10 minutes.

**So we end up with a CI of 9 minutes**, see the [summary](https://github.com/hirosystems/clarinet/actions/runs/14265232089?pr=1738) of the last steps


tippenein (approved):

deantchi (approved):
Well done @hugocaillard !
You've introduced a lot of great improvements and it looks good as far as I can see.

PR #1749, opened 2025-04-08
format pox-4 

tippenein:
- Formats pox-4 contract
- adds it to the golden tests


tippenein:
Opened for discussion in the diff. Probably some things that need tweaking, but this will likely end up in the golden tests..

tippenein:
@hugocaillard we can wait for #1790. I don't want to lose the significant newlines

hugocaillard (approved):
lgtm
Looks like git lost track of the actual commits in the PR 🤔 
But that'll be squashed anyway 👍

PR #1751, opened 2025-04-08
Feat/mxs historical bitcoin data

hugocaillard:

### DIFF
--- a/components/clarity-repl/src/repl/datastore.rs
+++ b/components/clarity-repl/src/repl/datastore.rs
@@ -1108,12 +1108,16 @@ impl BurnStateDB for Datastore {
     /// Returns Some if `self.get_burn_start_height() <= height < self.get_burn_block_height(sortition_id)`, and None otherwise.
     fn get_burn_header_hash(
         &self,
-        _height: u32,
-        sortition_id: &SortitionId,
+        height: u32,
+        _sortition_id: &SortitionId,
     ) -> Option<BurnchainHeaderHash> {
         println!("get_burn_header_hash");
-        println!("sortition_id: {:?}", sortition_id);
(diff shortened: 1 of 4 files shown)


components/clarity-repl/src/repl/datastore.rs
```diff
@@ -628,12 +634,15 @@ impl Datastore {
...
             constants,
+            remote_network_info: None,
+            client,
         }
     }
 
     fn new_with_remote_data(
         clarity_datastore: &ClarityDatastore,
         constants: StacksConstants,
+        client: HttpClient,
```
tippenein, line 645: Why pass the `client` and then use `clarity_datastore.client` anyway. Is the diff just deceiving me?

tippenein (approved):
One small question but otherwise LGTM

components/clarity-repl/src/repl/datastore.rs
obycode: `client `is used in the `Datastore` constructor, but it's not used to fetch the sortition on 669.

obycode: I think that might be fine.

obycode (dismissed):

components/clarity-repl/src/repl/datastore.rs
hugocaillard: Yes I think both are fine, it doesn't matter much which one we use they are basically the same thing.
In the past I used `clarity_datastore.client` because it was available there. But actually, now that I pass a client as the argument, i might as well just use it 👍 
I made the change @tippenein

tippenein (approved):
Still LGTM

PR #1760, opened 2025-04-11
fix: mxs tenure info

hugocaillard:
### Description

In the context of mainnet-execution simulation, fixes:
- `(get-tenure-info? vrf-seed <height>)` - `vrf-seed` is fetched in `/v3/sortitions/burn_height/<burn-height>`
- `(get-tenure-info? time <height>)` - `burn_block_time` was already fetched in `/extended/v2/blocks/<height>`


obycode (approved):
LGTM!

PR #1766, opened 2025-04-15
ci: allow running ci sdk release manually

hugocaillard:
### Description

Because the job wasn't trigger automatically for 2.16.0. We'll need to triggerit manually


deantchi (approved):
looks good

PR #1770, opened 2025-04-16
ci: update release to use GH App token

deantchi:
Uses a GitHub App generated token for Homebrew releases

### DIFF
--- a/.github/workflows/pkg-version-bump.yaml
+++ b/.github/workflows/pkg-version-bump.yaml
@@ -25,9 +25,23 @@ jobs:
       contents: write
       issues: write
     steps:
+      - name: Generate release bot app token
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pkg-version-bump.yaml
```diff
@@ -25,9 +25,23 @@ jobs:
...
     steps:
+      - name: Generate release bot app token
+        id: generate_token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.HIROSYSTEMS_PKG_RELEASE_BOT_APP_ID }}
+          private-key: ${{ secrets.HIROSYSTEMS_PKG_RELEASE_BOT_APP_PEM }}
+
+      - name: Get bot user ID
+        id: bot-user-id
```
hugocaillard, line 36: @deantchi I noticed we usually don't specify steps `id`s.
Do we need to?
If so, it should have consistent casing (`-` vs `_`)

deantchi: Right, `id` is not required for this step.  I will remove it.

PR #1771, opened 2025-04-17
ci: winget release token

hugocaillard:
### Description

use  release bot token to open winget release pr

### DIFF
--- a/.github/workflows/pkg-version-bump.yaml
+++ b/.github/workflows/pkg-version-bump.yaml
@@ -14,43 +14,30 @@ on:
       - released
 
 jobs:
-  generate-token:
-    name: Generate Token
-    runs-on: ubuntu-latest
-    outputs:
-      token: ${{ steps.generate-bot-token.outputs.token }}
-      app-slug: ${{ steps.generate-bot-token.outputs.app-slug }}
-      user-id: ${{ steps.get-bot-user-id.outputs.user-id }}
+  homebrew:
+    name: Homebrew
+    runs-on: macos-latest
     steps:
       - name: Generate release bot app token
-        id: generate-bot-token
+        id: generate-token
         uses: actions/create-github-app-token@v1
         with:
           app-id: ${{ secrets.HIROSYSTEMS_PKG_RELEASE_BOT_APP_ID }}
           private-key: ${{ secrets.HIROSYSTEMS_PKG_RELEASE_BOT_APP_PEM }}
-          skip-token-revoke: true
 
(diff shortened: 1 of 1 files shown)


.github/workflows/pkg-version-bump.yaml
```diff
@@ -18,29 +18,36 @@ jobs:
...
-      app-slug: ${{ steps.generate_token.outputs.app-slug }}
+      token: ${{ steps.generate-bot-token.outputs.token }}
+      app-slug: ${{ steps.generate-bot-token.outputs.app-slug }}
+      user-id: ${{ steps.get-bot-user-id.outputs.user-id }}
     steps:
       - name: Generate release bot app token
-        id: generate_token
+        id: generate-bot-token
         uses: actions/create-github-app-token@v1
         with:
```
deantchi, line 28: I think if you're planning on splitting the Token Generation tasks into its own job, you'll need to add `skip-token-revoke` so the post step won't revoke it automatically.  [see here](https://github.com/actions/create-github-app-token?tab=readme-ov-file#skip-token-revoke)

.github/workflows/pkg-version-bump.yaml
```diff
@@ -18,29 +18,36 @@ jobs:
...
+        id: get-bot-user-id
+        env:
+          GH_TOKEN: ${{ steps.generate-bot-token.outputs.token }}
+        run: echo "user-id=$(gh api "/users/${{ steps.generate-bot-token.outputs.app-slug }}[bot]" --jq .id)" >> "$GITHUB_OUTPUT"
 
   homebrew:
     name: Homebrew
-    needs: generate-token
+    needs:
+      - generate-token
```
deantchi, line 41: The `Winget` job has this as a single-string format, this is a list-string format, should we make these consistent?

deantchi (changes requested):

.github/workflows/pkg-version-bump.yaml
hugocaillard: > Optional: If truthy, the token will not be revoked when the current job is complete.


So when will it be revoked? When the whole workflow is complet?

hugocaillard: Alright thanks.
Well, we learned that this was a bad idea all along any way

bgridley (approved):
I think I now know what all of this mostly does now. Thanks for working through all this pain. Looks good.

deantchi (approved):
I think this looks good! :shipit:

PR #1773, opened 2025-04-21
fix: trait and eol comment fix

tippenein:
- [x] fix eol comment spacing
- [x] fix stackflow trait formatting
- [x] fix multi-line list formatting

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -358,35 +358,57 @@ impl<'a> Aggregator<'a> {
         result
     }
 
+    // (define-trait trait-name (
+    //   (func1-name
+    //     (arg1-type arg2-type ...)
+    //     (return-type)
+    //   )
+    //   (func2-name (arg1-type arg2-type ...) (return-type))
+    // )
     fn define_trait(&self, exprs: &[PreSymbolicExpression], previous_indentation: &str) -> String {
         let mut acc = "(define-trait ".to_string();
         let indentation = &self.settings.indentation.to_string();
         let space = format!("{}{}", indentation, previous_indentation);
         acc.push_str(&self.format_source_exprs(slice::from_ref(&exprs[1]), previous_indentation));
+        acc.push_str(" (");
         acc.push('\n');
         acc.push_str(&space);
-        acc.push('(');
         let mut iter = exprs[2].match_list().unwrap().iter().peekable();
         while let Some(expr) = iter.next() {
             if let Some(list) = expr.match_list() {
                 let trailing = get_trailing_comment(expr, &mut iter);
-                acc.push('\n');
+                acc.push('(');
+                // name
+                acc.push_str(&self.display_pse(&list[0], previous_indentation));
                 let double = format!("{}{}", space, indentation);
-                acc.push_str(&double);
-                acc.push_str(&self.format_list(list, &double));
+                // args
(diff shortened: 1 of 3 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -358,35 +358,71 @@ impl<'a> Aggregator<'a> {
...
+                                list[i + 1].span().start_column - list[i].span().end_column - 1;
+                            let spaces = " ".repeat(count as usize);
+                            acc.push_str(&spaces);
+                            acc.push_str(&self.display_pse(&list[i + 1], previous_indentation));
+                            i += 2; // Skip both the list and its comment
+                        } else {
+                            i += 1; // Just skip the list
+                        }
+                    } else {
+                        i += 1; // Skip any non-list elements
```
obycode, line 405: Does this just silently delete an expression?

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -358,35 +358,71 @@ impl<'a> Aggregator<'a> {
...
+                let mut i = 1;
+                while i < list.len() {
+                    if let Some(element_list) = list[i].match_list() {
+                        // Found a list (either args or return type)
+                        acc.push('\n');
+                        acc.push_str(&double);
+                        acc.push_str(&self.format_list(element_list, &double));
+
+                        // Check if next item is a comment
+                        if i + 1 < list.len() && is_comment(&list[i + 1]) {
```
obycode, line 394: Would it be worth refactoring this logic, to call a function in both scenarios?

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1577,14 +1637,30 @@ mod tests_formatter {
...
         assert_eq!(src, result);
     }
+
+    #[test]
+    fn define_multiline_list() {
+        let src = r#"  (
+    (optional <sip-010>) ;; token
+    uint                 ;; amount
+    principal            ;; with
+    uint                 ;; nonce
```
obycode, line 1647: I know this is my example code, but I will reiterate that I would also be okay with the formatter rejecting these trailing comments and forcing a pre-comment, e.g.:

```clarity
  (
    ;;token
    (optional <sip-010>)
    ;; amount
    uint
    ;; with
    principal
    ;; nonce
    uint
   )
```

tippenein: I think it's fine to have the aligned eol comments. It wasn't the main bug here anyway. The fix for this was properly handling multi-line lists. The comments are just a cherry on top. Besides, the formatter goes through in order, so if we made them into pre-comments, we'd have to re-order the PSE, or do a read from the end.. which would get messy

components/clarinet-format/src/formatter/mod.rs
tippenein: It should be handled above (L#391), but I need to update this wording 👍

components/clarinet-format/src/formatter/mod.rs
tippenein: yeah, I think this whole index junk is a bad idea.. Let me try a `iter`

tippenein: Refactored to use peekable like everything else 👍

obycode (approved):
Yeah, that looks a lot better. 👍

PR #1775, opened 2025-05-01
chore: inline values in match

### LINKED ISSUE #1747: formatter - match arguments
Described [here](https://github.com/hirosystems/clarinet/discussions/1689#discussioncomment-12691662) is a proposal to stack match arguments on the same line if they fit, so instead of

```
    (match (sbtc-transfer amount tx-sender (as-contract tx-sender))
      success
      (ok id)
      error
      (err (* error u1000))
    )
```

We would have

```
    (match (sbtc-transfer amount tx-sender (as-contract tx-sender))
      success (ok id)
      error (err (* error u1000)))))
    )
```

tippenein:
closes #1747 

If comments are present, we special case it to the old fully-aligned way.. Otherwise we inline as much as possible.

For example, this existing code is difficult to fit into the inlined way

```
        (match registered-at-value
            is-registered
            ;; If it was registered, check if registered-at is lower than current blockheight
            ;; This check works to make sure that if a name is fast-claimed they have to wait 1 block to transfer it
            (asserts! (< is-registered burn-block-height) ERR-OPERATION-UNAUTHORIZED)
            ;; If it is not registered then continue
            true
        )
```

Having comments describing the values is useful obviously, but in an ideal world it would be something like 

```
        (match registered-at-value
            ;; If it was registered, check if registered-at is lower than current blockheight
            ;; This check works to make sure that if a name is fast-claimed they have to wait 1 block to transfer it
            is-registered (asserts! (< is-registered burn-block-height) ERR-OPERATION-UNAUTHORIZED)
            ;; If it is not registered then continue
            true
        )
```

Changing the order in which we place the comments is seemingly dangerous even though it might make sense in this example.


components/clarinet-format/tests/golden.rs
```diff
@@ -69,8 +70,14 @@ fn test_irl_contracts() {
...
             pretty_assertions::assert_eq!(result, intended, "Mismatch in file: {:?}", file_name);
             // parse resulting contract
-            let pse = clarity::vm::ast::parser::v2::parse(&result);
-            assert!(pse.is_ok());
+            let (_statements, diagnostics, success) =
+                clarity::vm::ast::parser::v2::parse_collect_diagnostics(&result);
+
+            if !diagnostics.is_empty() {
+                println!("Result of re-parsing file: {}", file_name.to_str().unwrap());
+                println!("Message: {:?}", diagnostics[0].message);
```
tippenein, line 78: small improvement to show the error message when the golden tests fail to parse the formatted version

hugocaillard (approved):
lgtm!

Issue #1769, opened 2025-04-16
Missing blank line before `define-<function>`


hugocaillard:
**Describe the bug**

With the following code, the formatter won't leave a blank line before `define-private`

```clar
(define-data-var count uint u10)

(define-private (get-count)
  (var-get count)
)
```

tippenein:
This is easy enough to add, but when you have a comment before the define-private, it gets messy. Trying to find a way that avoids completely unmaintainable code that can handle that case. Unfortunate way that I've set up top level function spacing I think..

```
(define-data-var count uint u10)

;; comment
(define-private (get-count)
  (var-get count)
)
```

tippenein:
rustfmt will not add newlines above top-level definitions and considering we have top-level newlines preserved I _think_ we should be in line with rustfmt on behavior. I'll close this for now, but I'd be willing to hear arguments in favor of adding the newlines even if original source doesn't have them

PR #1776, opened 2025-05-01
fix: top-level spacing

### LINKED ISSUE #1748: formatter - top-level exprs not handled correctly
The formatter makes some assumptions about what types of expressions can be top-level. These assumptions are wrong. For example this code is valid  but handled incorrectly:

```
(define-map ns
  uint
  bool
)
(let ((x (+ u1 u1)))
  (map-insert ns x true)
)(define-public (get-value)
  (ok (map-get? ns u2))
)
```

1) Fix the underlying assumptions to make the above clarity formatted as:

```
(define-map ns uint bool)
(let ((x (+ u1 u1)))
  (map-insert ns x true)
)

(define-public (get-value)
  (ok (map-get? ns u2))
)
```

2) The `define-map` should be on 1 line

tippenein:
fix #1748


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1405,6 +1412,19 @@ mod tests_formatter {
...
     }
 
+    #[test]
+    fn top_level_exprs() {
+        let src = r#"(let ((x (+ u1 u1)))
+  (map-insert ns x true)
+)
+(define-public (get-value)
+  (ok (map-get? ns u2))
+)
```
hugocaillard, line 1422: Shouldn't it be
```clarity
(let ((x (+ u1 u1)))
  (map-insert ns x true)
)

(define-public (get-value)
  (ok (map-get? ns u2))
)

```

With an extra new-line before define-public?

tippenein: That is part of this issue: https://github.com/hirosystems/clarinet/issues/1769

The problem there was the top-comments get in the way of easily new-lining define- functions

hugocaillard (approved):
lgtm

PR #1797, opened 2025-05-07
chore: simplify prepare-release and remove vscode special casing

tippenein:
- vscode version is now the same as clarinet overall, so no need to prompt
- update clarinet-sdk deps


hugocaillard (approved):
lgtm! thanks @tippenein

PR #1806, opened 2025-05-12
fix: list formatting breakage

### LINKED ISSUE #1799: Formatter breaks code after comments
**Describe the bug**
After formatting https://github.com/citycoins/protocol/blob/a6e79965b4e5ef5b8c6c04a11b4858120a48789b/contracts/proposals/ccip012-bootstrap.clar#L18 
the resulting clarity code is not valid anymore.

Using clarinet 3.0.1

Similar for https://github.com/citycoins/protocol/blob/63e2128c770b6ae82786582d01719514339e8d11/contracts/proposals/ccip016-missed-payouts.clar#L28

**To Reproduce**
1. Checkout out https://github.com/citycoins/protocol
2. Run `clarinet check`
3. Run `clarinet format --in-place`
4. Run `clarinet check`
5. See error

**Expected behavior**
Code remains valid.

tippenein:
- closes #1799 

- [x] fix inline comment newline issue
- [x] fix comment as item of a list
- [x] actually hand `format_list` for ListCons

This fixes the breakage in citycoins protocol contracts

Screenshot 2025-05-12 at 11 52 28 AM

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -201,6 +201,8 @@ impl<'a> Aggregator<'a> {
                                 self.format_booleans(list, previous_indentation)
                             }
 
+                            NativeFunctions::ListCons => self.format_list(list, previous_indentation),
+
                             // everything else that's not special cased
                             NativeFunctions::Add
                             | NativeFunctions::Subtract
@@ -240,7 +242,6 @@ impl<'a> Aggregator<'a> {
                             | NativeFunctions::StringToUInt
                             | NativeFunctions::IntToAscii
                             | NativeFunctions::IntToUtf8
-                            | NativeFunctions::ListCons
                             | NativeFunctions::FetchVar
                             | NativeFunctions::FetchEntry // map-get?
                             | NativeFunctions::SetEntry // map-set?
@@ -342,7 +343,6 @@ impl<'a> Aggregator<'a> {
                         };
                         let result = &formatted.to_string();
                         if let Some(comment) = trailing_comment {
-                            println!("COMMENT: {:?}", comment);
                             let mut result_with_comment = result.to_string();
                             result_with_comment.push(' ');
                             result_with_comment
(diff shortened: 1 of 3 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -941,7 +1020,7 @@ impl<'a> Aggregator<'a> {
                 _ => value.to_string(),
             },
             PreSymbolicExpressionType::List(ref items) => {
-                self.format_list(items, previous_indentation)
+                self.display_list(items, previous_indentation)
```
tippenein, line 1023: this is nasty duplication but if we fall into the `display_pse` branch we want to format it like before. If the atom is a `ListCons` we use the new `format_list`

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
```diff
@@ -264,7 +264,8 @@
         index: uint,
       },
       remaining: uint,
-      txins: (list 8 {
+      txins: (list 8
+        {
```
hugocaillard, line 268: not sure I follow this change.
If this `}` is on a new line, I think the indentation level shold be increased for the next lines.
And the matching `}` should be on the same indentation level

```clarity
      txins: (list 8
        {
          outpoint: {
            hash: (buff 32),
            index: uint,
          },
          scriptSig: (buff 256), ;; just big enough to hold a 2-of-3 multisig script
          sequence: uint,
        }
      ),
```

hugocaillard (approved):
I think what's the PR is trying to fix is looking good.

I left a questions regarding a side effect of theses changes in clarity-bitcoin.clar

components/clarinet-format/tests/golden-intended/clarity-bitcoin.clar
tippenein: Good catch. I believe this is because there's a difference between checking spans (list breaking on max lines) and the original place the opening bracket would have been 
```
(list 8 {
  ...
})
```

So, the list gets broken up onto multiple lines and the closing parens are still under the impression the `(` and `{` were opened on the same line.. Tricky, but i've made an issue [here](https://github.com/hirosystems/clarinet/issues/1810)

PR #1808, opened 2025-05-13
fix: mxs burn chain hash data

hugocaillard:
### Description

In the context of `mxs`, `get_burn_header_hash` was comparing burn block height against a stacks block height.
Resulting in weird result if the burn block height was higher than the stacks block height

### DIFF
--- a/components/clarity-repl/src/repl/datastore.rs
+++ b/components/clarity-repl/src/repl/datastore.rs
@@ -148,6 +148,7 @@ pub struct Datastore {
     current_epoch_start_height: u32,
     constants: StacksConstants,
     remote_network_info: Option<RemoteNetworkInfo>,
+    initial_burn_height: u32,
     client: HttpClient,
 }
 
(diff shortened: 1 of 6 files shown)


components/clarity-repl/src/repl/datastore.rs
```diff
@@ -1163,22 +1166,20 @@ impl BurnStateDB for Datastore {
...
+        if burn_height > self.burn_chain_height {
             return None;
         }
 
-        if let Some(remote_info) = &self.remote_network_info {
-            if height <= remote_info.initial_height {
-                let sortition = self.client.fetch_sortition(height);
-                return Some(sortition.burn_block_hash);
-            }
+        if self.remote_network_info.is_some() && burn_height <= self.initial_burn_height {
```
hugocaillard, line 1176: This is the fix

tippenein (approved):
LGTM

PR #1812, opened 2025-05-13
feat: devnet block data cache

tippenein:
main changes
- [x] handle creating devnet block cache
- [x] copy from global on devnet startup
- [x] bitcoin node picks up cached data
- [ ] stacks node picks up cached data
- [ ] allow --no-cache
- [ ] manifest checking
- [ ] find a clean way of bundling the global cache data with clarinet

other changes
- [x]  bind_containers_volumes defaults true (since we need that to populate ~/.cache data)


.gitignore
```diff
@@ -16,6 +16,7 @@ npm-debug.log*
 components/stacks-devnet-js/dist
 components/stacks-devnet-js/build
 *.tar.gz
+!components/stacks-network/data/devnet_default_snapshot.tar.gz
```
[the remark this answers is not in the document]
tippenein: I went the route of global snapshot population. It's vastly simpler than docker snapshots and allows us flexibility later on with custom snapshots. Briefly, we'll need the global cache set up eventually for custom snapshots, so it makes more sense (to me at least) to keep all the snapshot data in one place

components/stacks-network/Cargo.toml
```diff
@@ -39,6 +39,8 @@ clarinet-files = { path = "../clarinet-files" }
 clarinet-deployments = { path = "../clarinet-deployments" }
 hiro-system-kit = { path = "../hiro-system-kit", features = ["log"] }
 stacks-codec = { path = "../stacks-codec" }
+flate2 = "1.1.1"
+tar = "0.4.44"
```
hugoclrd: We have this non-enforced convention in the repo where crates.io dependencies are listed first, followed by local dependencies.
Can you move these two new dependencies to lines 31 and 32?

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -1112,153 +1302,79 @@ fn get_stacking_tx_method_and_args(
     (method.to_string(), arguments)
 }
 
-#[cfg(test)]
```
tippenein, line 1115: Not sure how I ended up deleting this.. Adding back now

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -1113,6 +1302,83 @@ fn get_stacking_tx_method_and_args(
...
+
+    if !output.status.success() {
+        return Err(format!(
+            "Export command failed: {}",
+            String::from_utf8_lossy(&output.stderr)
+        ));
+    }
+
+    // Wait for export to complete
+    std::thread::sleep(Duration::from_secs(5));
```
tippenein, line 1351: This is not great..

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,255 @@
...
+                extractor: make_extractor(|config| config.stacks_api_image_url.clone()),
+            },
+            // Stacking orders
+            SignificantField {
+                name: "pox_stacking_orders".to_string(),
+                extractor: make_extractor(|config| {
+                    let mut orders = config.pox_stacking_orders.clone();
+                    orders.sort_by(|a, b| a.wallet.cmp(&b.wallet));
+                    orders
+                        .iter()
```
jbencin: Minor nit:

```suggestion
                        .into_iter()
```

Clippy will complain, may result in slightly better performance

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,255 @@
...
+    ) -> Result<bool, String> {
+        for field in &self.significant_fields {
+            let default_value = (field.extractor)(default_config);
+            let user_value = (field.extractor)(user_config);
+
+            if default_value != user_value {
+                return Err(format!(
+                    "user_value: {:?}\ndefault_value: {:?}",
+                    user_value, default_value
+                ));
```
jbencin: Can you add the `indoc` crate to `clarinet-files` (actually you can make it a workspace dependency) and do this instead:

```suggestion
                return Err(formatdoc!("
                    user_value: {user_value:?}
                    default_value: {default_value:?}"
                ));
```

The `formatdoc!` macro strips the extra indentation, and makes it so you can have readable multiline strings

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    /// Get the names of fields that are different
+    pub fn get_different_field_names(
+        &self,
+        default_config: &DevnetConfig,
+        user_config: &DevnetConfig,
+    ) -> Vec<String> {
+        self.get_different_fields(default_config, user_config)
+            .into_iter()
+            .collect()
```
jbencin: Not needed:

```suggestion
        self.get_different_fields(default_config, user_config)
```

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        let mut report = format!(
+            "Found {} significant difference(s):\n\n",
+            different_fields.len()
+        );
+
+        for field in different_fields {
+            report.push_str(&format!("• {}\n", field));
+
+            report.push('\n');
```
jbencin: Prefer `write!()` to `push_str(&format(...))`, as it doesn't allocate an intermediate string: 

```suggestion
            write!(report, "• {field}\n\n").unwrap();
```

Make sure to import the `std::fmt` version, **not** the `std::io` one:

```rust
use std::fmt::Write as _;
```

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -456,6 +456,9 @@ struct DevnetStart {
     /// Display streams of logs instead of terminal UI dashboard
     #[clap(long = "no-dashboard")]
     pub no_dashboard: bool,
+    /// Don't use snapshot data on startup
+    #[clap(long = "no-snapshot")]
```
jbencin: Alternatively, you could name the option `from-genesis`. Maybe more self-explanatory?

tippenein: I like it! I'll change that up 👍  We'll keep the snapshot terminology in the codebase though

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,254 @@
...
+        for field in &self.significant_fields {
+            let default_value = (field.extractor)(default_config);
+            let user_value = (field.extractor)(user_config);
+
+            if default_value != user_value {
+                return Err(indoc::formatdoc!(
+                    "
+                user_value: {user_value:?}
+                default_value: {default_value:?}"
+                ));
```
jbencin: Fix whitespace:

```suggestion
                return Err(indoc::formatdoc!("
                    user_value: {user_value:?}
                    default_value: {default_value:?}"
                ));
```

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -178,11 +188,21 @@ pub async fn start_chains_coordinator(
...
     mining_command_tx: Sender<BitcoinMiningCommand>,
     mining_command_rx: Receiver<BitcoinMiningCommand>,
+    using_snapshot: bool,
     ctx: Context,
 ) -> Result<(), String> {
     let mut should_deploy_protocol = true; // Will change when `stacks-network` components becomes compatible with Testnet / Mainnet setups
     let boot_completed = Arc::new(AtomicBool::new(false));
-    let mut current_burn_height = 0;
+    let mut current_burn_height = if using_snapshot { 143 } else { 0 };
+    let starting_block_height = if using_snapshot { 38 } else { 0 };
```
jbencin: Can these be named constants instead of hard-coded numbers?

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -234,14 +254,82 @@ pub async fn start_chains_coordinator(
...
+            }
+        }
+
+        let _ = observer_event_tx.send(ObserverEvent::StacksChainEvent((
+            StacksChainEvent::ChainUpdatedWithBlocks(StacksChainUpdatedWithBlocksData {
+                new_blocks: events
+                    .clone()
+                    .iter()
+                    .map(|event| StacksBlockUpdate::new(event.clone()))
+                    .collect::<Vec<_>>(),
```
jbencin: This can be cleaned up and cloning `events` is not necessary:

```suggestion
                new_blocks: events
                    .iter()
                    .cloned()
                    .map(StacksBlockUpdate::new)
                    .collect(),
```

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -234,14 +254,82 @@ pub async fn start_chains_coordinator(
...
+    let event_pool: Vec<StacksBlockData> = if using_snapshot {
+        let mut events = vec![];
+        let events_cache_path = get_global_snapshot_dir()
+            .join("events_export")
+            .join("events_cache.tsv");
+
+        let mut chain_ctx = StacksChainContext::new(&chainhook_types::StacksNetwork::Devnet);
+        if let Ok(file_content) = fs::read_to_string(&events_cache_path) {
+            for line in file_content.lines() {
+                if line.split('\t').nth(2).unwrap_or("") == "/new_block" {
```
jbencin: Bind the `split()` result to a variable so you only have to split once:

```suggestion
                let line_parts = line.split('\t');
                if line_parts.nth(2).unwrap_or("") == "/new_block" {
```

And then change:

```rust
line.split('\t').nth(3).unwrap_or(""),
```

to:

```rust
line_parts.next().unwrap_or(""),
```

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -646,6 +745,97 @@ fn should_publish_stacking_orders(
...
+    let project_snapshot_dir = get_project_snapshot_dir(devnet_config);
+
+    // Project snapshot marker
+    let project_marker = project_snapshot_dir.join("epoch_3_ready");
+    if !project_marker.exists() {
+        match std::fs::File::create(&project_marker) {
+            Ok(_) => {
+                let _ = devnet_event_tx.send(DevnetEvent::success(
+                                            "Project snapshot data prepared up to epoch 3.0. Future project starts will be faster.".to_string(),
+                                        ));
```
jbencin: This looks weird... did you run `cargo fmt`?

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -646,6 +745,97 @@ fn should_publish_stacking_orders(
...
+
+    let global_marker = global_snapshot_dir.join("epoch_3_ready");
+    if !global_marker.exists() {
+        // Copy project snapshot to global snapshot as a template
+        if project_snapshot_dir != global_snapshot_dir {
+            // Copy bitcoin data
+            let project_bitcoin_snapshot = project_snapshot_dir.join("bitcoin");
+            let global_bitcoin_snapshot = global_snapshot_dir.join("bitcoin");
+            if project_bitcoin_snapshot.exists() {
+                let _ = copy_directory(&project_bitcoin_snapshot, &global_bitcoin_snapshot, None);
```
jbencin: Probably shouldn't ignore errors here, this could easily fail due to a permissions error, you should at least issue a warning if that happens

components/stacks-network/src/chains_coordinator.rs
jbencin: Is this necessary because the export command doesn't block until it's finished?

Is there something you could loop and poll for rather than sleeping?

components/stacks-network/src/lib.rs
```diff
@@ -76,6 +79,76 @@ async fn do_run_devnet(
...
+                Err(e) => {
+                    let _ = devnet_events_tx.send(DevnetEvent::warning(format!(
+                        "Failed to extract embedded snapshot: {}. Continuing without snapshot.",
+                        e
+                    )));
+                }
+            }
+        }
+        match setup_snapshot_directories(&devnet_config, &devnet_events_tx) {
+            Ok(using_snapshot) => {
```
jbencin: IMO it's easier to read matching on `Ok(true)`/`Ok(false)` like you did above. Clippy also warns about this I think. Not really important though

components/stacks-network/src/orchestrator.rs
```diff
@@ -1870,10 +1874,45 @@ events_keys = ["*"]
         ctx.try_log(|logger| slog::info!(logger, "Created container stacks-api: {}", container));
         self.stacks_api_container_id = Some(container);
 
+        // // If we have events to import, we need to copy the file to the container
```
jbencin: There's no need to commit commented out code. You should either put it inside an `if` block or `#[cfg(...)]` or delete it

components/stacks-network/src/orchestrator.rs
```diff
@@ -2805,14 +3091,41 @@ events_keys = ["*"]
...
+                        // if createwallet fails it likely means we need to load the existing wallet
+                        match rpc_load_call {
+                            Ok(r) => {
+                                if r.status().is_success() {
+                                    break;
+                                } else {
+                                    let err = r.text().await;
+                                    let msg = format!("{:?}", err);
+                                    // if it returns "Wallet is already loaded" we break out
+                                    if err.unwrap().contains("is already loaded") {
```
jbencin: Can we remove the `unwrap()` here and return an error instead?

components/clarinet-files/src/devnet_diff.rs
tippenein: That's how rustfmt decided to do it. No idea why

components/stacks-network/src/chains_coordinator.rs
tippenein: yea, it formats on save

components/stacks-network/src/chains_coordinator.rs
tippenein: This isn't the sleep I was thinking of. There's one for the stacks-api export that pauses, exports, then waits for the bitcoin node to start mining again. I probably need to audit all the sleeps that were added though. Some are necessary, but sleep isn't the best way to accomplish it obviously

components/clarinet-files/src/devnet_diff.rs
jbencin: Weird. Could add `#[rustfmt::skip]` here

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -48,13 +56,18 @@ use stackslib::types::chainstate::StacksPublicKey;
...
+use std::path::PathBuf;
 use std::str;
 use std::str::FromStr;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::mpsc::{channel, Receiver, Sender};
 use std::sync::Arc;
 use std::time::Duration;
 
+static SNAPSHOT_STACKS_START_HEIGHT: u64 = 38;
+static SNAPSHOT_BURN_START_HEIGHT: u64 = 143;
```
jbencin: Prefer `const` unless you specifically need to allocate memory for a variable:

```suggestion
const SNAPSHOT_STACKS_START_HEIGHT: u64 = 38;
const SNAPSHOT_BURN_START_HEIGHT: u64 = 143;
```

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -646,6 +756,103 @@ fn should_publish_stacking_orders(
...
+                            e
+                        )));
+                    });
+            }
+
+            // Copy stacks data
+            let project_stacks_snapshot = project_snapshot_dir.join("stacks");
+            let global_stacks_snapshot = global_snapshot_dir.join("stacks");
+            if project_stacks_snapshot.exists() {
+                let _ = copy_directory(
```
jbencin, line 797: Same as above, should warn on failure here

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -234,14 +264,83 @@ pub async fn start_chains_coordinator(
...
+                }
+            }
+        }
+
+        //     let _ = observer_event_tx.send(ObserverEvent::StacksChainEvent((
+        //         StacksChainEvent::ChainUpdatedWithBlocks(StacksChainUpdatedWithBlocksData {
+        //             new_blocks: events
+        //                 .iter()
+        //                 .cloned()
+        //                 .map(|event| StacksBlockUpdate::new(event.clone()))
```
jbencin: `clone()` isn't needed here because of `cloned()` above. Also try to avoid committing commented code

```suggestion
        //                 .map(StacksBlockUpdate::new)
```

components/stacks-network/src/orchestrator.rs
```diff
@@ -2692,105 +2901,166 @@ events_keys = ["*"]
...
+                    Ok(_r) => break,
+                    Err(e) => {
+                        error_count += 1;
+                        if error_count > max_errors {
+                            return Err(e);
+                        } else if error_count > 1 {
+                            let _ = devnet_event_tx.send(DevnetEvent::error(e));
+                        }
+                    }
+                }
```
jbencin: Optional change, but IMO `let`/`else` is a little cleaner here and results in less indentation (same thing at lines 2926, 2993, and 3029):

```suggestion
                let Err(e) = rpc_call else {
                    break;
                };
                error_count += 1;
                if error_count > max_errors {
                    return Err(e);
                } else if error_count > 1 {
                    let _ = devnet_event_tx.send(DevnetEvent::error(e));
                }
```

components/clarinet-files/src/devnet_diff.rs
tippenein: added fmt skip 👍

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,260 @@
...
+    }
+
+    fn clean_stacking_orders(
+        &self,
+        user_value: String,
+        default_value: String,
+    ) -> Vec<(String, String, String)> {
+        let mut result = Vec::new();
+        let user_orders: Vec<String> = user_value.split(",").map(|s| s.to_string()).collect();
+        let default_orders: Vec<String> = default_value.split(",").map(|s| s.to_string()).collect();
```
jbencin: Minor, but slightly cleaner:

```suggestion
        let user_orders: Vec<String> = user_value.split(",").map(String::from).collect();
        let default_orders: Vec<String> = default_value.split(",").map(String::from).collect();
```

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,260 @@
...
+            [
+                "start_at_cycle",
+                "duration",
+                "wallet",
+                "slots",
+                "btc_address",
+            ],
+        ) {
+            order_vals.push((field.to_string(), v));
+        }
```
jbencin: Can this be cleaned up a bit:

```suggestion
            let vals: Vec<_> = order.split("-").map(String::from).collect();
            let btc_address = vals[4].to_owned();
            let fields = [
                "start_at_cycle",
                "duration",
                "wallet",
                "slots",
                "btc_address",
            ];
            let order_vals: Vec<_> = fields
                .map(String::from)
                .into_iter()
                .zip(vals)
                .collect();
```

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,260 @@
...
+                if default_val != &user_val {
+                    result.push((field, user_val, default_val.clone()));
+                }
+            }
+        }
+        result
+    }
+}
+
+fn construct_stacking_orders(orders: &Vec<String>) -> HashMap<String, Vec<(String, String)>> {
```
jbencin: As a general rule, it's better to pass slices if you aren't going to consume/modify a `Vec`:

```suggestion
fn construct_stacking_orders(orders: &[String]) -> HashMap<String, Vec<(String, String)>> {
```

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -646,6 +741,110 @@ fn should_publish_stacking_orders(
...
+    let devnet_config = &devnet_event_observer_config.devnet_config;
+    let global_snapshot_dir = get_global_snapshot_dir();
+    let project_snapshot_dir = get_project_snapshot_dir(devnet_config);
+
+    // Project snapshot marker
+    let project_marker = project_snapshot_dir.join("epoch_3_ready");
+    if !project_marker.exists() {
+        match std::fs::File::create(&project_marker) {
+            Ok(_) => {
+                #[rustfmt::skip]
```
jbencin: If I remove this line, `cargo fmt` doesn't change the formatting

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -646,6 +741,110 @@ fn should_publish_stacking_orders(
...
+                        "Failed to copy stacks snapshot: {}",
+                        e
+                    )));
+                });
+            }
+
+            match std::fs::File::create(&global_marker) {
+                Ok(_) => {
+                    let _ = devnet_event_tx.send(DevnetEvent::success("Global template cache data prepared. Future project initializations will be faster.".to_string(),
+                                        ));
```
jbencin: More `cargo fmt` misbehavior. If you just fix it manually `cargo fmt` won't change it

components/stacks-network/src/orchestrator.rs
```diff
@@ -994,11 +1020,46 @@ rpcport={bitcoin_node_rpc_port}
...
+                &container,
+                &bitcoin_snapshot,
+                "/root/.bitcoin/",
+                devnet_event_tx,
+                "Bitcoin",
+            )
+            .await?;
+        }
+
+        // if !no_snapshot {
```
jbencin: Uncomment or remove

jbencin (dismissed):

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,286 @@
...
+                extractor: make_extractor(|config| config.bitcoin_node_image_url.clone()),
+            },
+            SignificantField {
+                name: "stacks_node_image_url".to_string(),
+                extractor: make_extractor(|config| config.stacks_node_image_url.clone()),
+            },
+            SignificantField {
+                name: "stacks_api_image_url".to_string(),
+                extractor: make_extractor(|config| config.stacks_api_image_url.clone()),
+            },
```
hugoclrd: @tippenein 
I think this was relevant when the idea was to have the snapshot in the docker images (using the snapshot would possubly not work with custom images). But now, I could potentially use a custom bitcoind or stacks-node image and still use the snapshot right?

jbencin:
> Was there a a `no-snapshot` option at some point?

It was renamed to `--from-genesis`

jbencin (dismissed):

components/clarinet-files/src/devnet_diff.rs
tippenein: Possibly. It depends the bounds of how different it can be, because imagining some stacks_api image where it's using an entirely different setup and doesn't have the import/export script for events, we'd have an issue. The stacks and bitcoin node should be fine though.

tippenein: I've removed all 3 for now because it's unlikely and more likely to be frustrating for anyone trying to use a different image

hugoclrd: Right, that sounds find.
And we can still use --from-genesis when using an image that isn't compatible .

This is some power user stuff anyway

components/clarinet-files/Cargo.toml
```diff
@@ -8,6 +8,7 @@ license = "GPL-3.0"
 [dependencies]
 dirs = "6.0"
 libsecp256k1 = "0.7.0"
+indoc = "2.0.6"
```
hugoclrd: I don't think it's used

components/clarinet-files/src/devnet_diff.rs
```diff
@@ -0,0 +1,268 @@
...
+                extractor: make_extractor(|config| config.epoch_2_5),
+            },
+            SignificantField {
+                name: "epoch_3_0".to_string(),
+                extractor: make_extractor(|config| config.epoch_3_0),
+            },
+            SignificantField {
+                name: "epoch_3_1".to_string(),
+                extractor: make_extractor(|config| config.epoch_3_1),
+            },
```
hugoclrd: Since we boot in 3.0, developers should be able to set ac custom 3.1 height right?

hugocaillard (changes requested):
On last thing

hugocaillard (approved):
LGTM!  🚀 
Can't wait to release

PR #1818, opened 2025-05-16
Fix/clippy 1 87 0

hugocaillard:
### Description

Allow `clippy::result_large_err` for now. Will have to revisit once it handled upstream in stacks-core


csgui (approved):
This is an interesting thing to think about. Even if errors are rare in an application, every allocation of the Result must handle the full size, which can increase the memory pressure, especially when errors are propagated through the `?` operator.

tippenein (approved):

PR #1822, opened 2025-05-17
fix: top level newline issue

### LINKED ISSUE #1811: Formatter example - AIBTC
Sharing here as we just upgraded to Clarinet 3.0.1 and ran the formatter. Most of what we do was close to the formats but thought it'd be interesting to capture the changes on their own for review.

Nothing to contest yet but will add here if anything is found.
https://github.com/aibtcdev/aibtcdev-daos/pull/38

tippenein:
- closes #1811

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -317,8 +317,12 @@ impl<'a> Aggregator<'a> {
                                             " {}\n",
                                             &self.display_pse(comment, previous_indentation)
                                         )
-                                    } else if iter.peek().is_some() {
-                                        " ".to_string()
+                                    } else if let Some(next) = iter.peek() {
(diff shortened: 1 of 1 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -357,10 +361,7 @@ impl<'a> Aggregator<'a> {
                         self.to_inner_content(list, previous_indentation)
                     };
                     // if it's a top level expression, newline it
-                    if previous_indentation.is_empty()
-                        && result.ends_with(")")
-                        && !result.ends_with(")\n")
-                    {
+                    if previous_indentation.is_empty() && (result.ends_with(")")) {
```
tippenein, line 364: The changes to `format_list` caused a regression where the (print ...) was ending with ") " so this condition wasn't being picked up. The change above catches it

PR #1824, opened 2025-05-19
feat: Add `--stdin` flag to formatter

### LINKED ISSUE #1816: allow --stdin for formatter
Add a flag to clarinet fmt to take file contents from stdin

```
echo '(ok true)' | clarinet fmt --stdin
```

This would conflict with `-f`

jbencin:
### Description

Fixes #1816

This is needed to use the formatter with Helix editor (at least until LSP working). May also be useful with other text editors

Also, some refactoring in here so that all files aren't read into RAM at the same time

#### Command

```sh
echo '(define-public (hello-world (input int)) (begin (print (+ 2 input)) (ok input)))' | clarinet format --stdin
```

#### Output

```clarity
(define-public (hello-world (input int))
  (begin
    (print (+ 2 input))
    (ok input)
  )
)
```

---

### Checklist

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -124,6 +124,59 @@ struct Formatter {
     #[clap(long = "in-place", conflicts_with = "dry_run")]
     /// Replace the contents of a file with the formatted code
     pub in_place: bool,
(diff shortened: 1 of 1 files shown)


tippenein (approved):
:shipit:

hugocaillard (approved):
lgtm!

PR #1826, opened 2025-05-20
feat: accept ast only in the formatter

hugocaillard:
### Description

As part of the TS to Clarity transpiler project, I need to be able to format a Clarity AST (with no source) 

This PR makes the source optional in the formatter, and adds a `format_ast` function along with `format_file`

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -69,11 +69,16 @@ impl ClarityFormatter {
     /// formatting for files to ensure a newline at the end
     pub fn format_file(&self, source: &str) -> String {
         let pse = clarity::vm::ast::parser::v2::parse(source).unwrap();
-        let agg = Aggregator::new(&self.settings, &pse, source);
+        let agg = Aggregator::new(&self.settings, &pse, Some(source));
         let result = agg.generate();
 
         // make sure the file ends with a newline
         result.trim_end_matches('\n').to_string() + "\n"
+    }
+    /// formatting an AST without a source file
+    pub fn format_ast(&self, pse: &[PreSymbolicExpression]) -> String {
+        let agg = Aggregator::new(&self.settings, pse, None);
+        agg.generate()
     }
     /// Alias `format_file` to `format`
     pub fn format(&self, source: &str) -> String {
@@ -88,7 +93,7 @@ impl ClarityFormatter {
         // `previous_indentation` for format_source_exprs
         let indentation_level = source.chars().take_while(|c| c.is_whitespace()).count();
         let leading_spaces = &source[..indentation_level];
-        let agg = Aggregator::new(&self.settings, &pse, source);
+        let agg = Aggregator::new(&self.settings, &pse, Some(source));
 
         let result = agg.generate();
         if leading_spaces.is_empty() {
@@ -104,13 +109,17 @@ impl ClarityFormatter {
 pub struct Aggregator<'a> {
     settings: &'a Settings,
     pse: &'a [PreSymbolicExpression],
-    source: &'a str,
+    source: Option<&'a str>,
 
     cache: RefCell<HashMap<(usize, String), String>>,
 }
 
 impl<'a> Aggregator<'a> {
-    pub fn new(settings: &'a Settings, pse: &'a [PreSymbolicExpression], source: &'a str) -> Self {
+    pub fn new(
+        settings: &'a Settings,
+        pse: &'a [PreSymbolicExpression],
+        source: Option<&'a str>,
+    ) -> Self {
         Aggregator {
             settings,
             pse,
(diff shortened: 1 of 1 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -123,10 +132,9 @@ impl<'a> Aggregator<'a> {
         // this handles if we're formatting a section of code rather than the whole file
         let indentation_level = self
             .source
-            .chars()
-            .take_while(|c| c.is_whitespace())
-            .count();
-        self.format_source_exprs(self.pse, &self.source[..indentation_level])
+            .map_or(0, |s| s.chars().take_while(|c| c.is_whitespace()).count());
```
tippenein, line 135: This might be wrong. I think we'd have to check the PSE `span().start_column` to get an accurate indentation level

tippenein: `self.pse[0].span().start_column - 1` I think. It's possible this is just easier than counting whitespaces from source anyway. I believe there are tests for this, so it'd be good for the formatter. A question might be how LSP passes the PSEs in though. Worth trying in the browser to make sure

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1981,4 +1989,13 @@ mod tests_formatter {
...
     }
+
+    #[test]
+    fn format_ast_without_source() {
+        let src = "(define-private (noop) (ok true))";
+        let ast = clarity::vm::ast::parser::v2::parse(src).unwrap();
+        let formatter = ClarityFormatter::new(Settings::default());
+        let result = formatter.format_ast(&ast);
+        assert_eq!(result, "(define-private (noop)\n  (ok true)\n)\n");
+    }
```
tippenein, line 2000: Could you add this as well? I don't think `ignored_exprs` will work without the source string
```
    #[test]
    fn format_ast_without_source_ignore() {
        let src = ";; @format-ignore\n(define-private (noop) (ok true))";
        let ast = clarity::vm::ast::parser::v2::parse(src).unwrap();
        let formatter = ClarityFormatter::new(Settings::default());
        let result = formatter.format_ast(&ast);
        assert_eq!(result, src);
    }
```

hugocaillard: Indeed, that is precisely the kind of edge case I wouldn't have anticipated.
But that would be the expected behavior, right?

If we only have an AST and no source code as a "source of truth", it wouldn't make sense to "ignore formatting". The idea is to be able to take a raw AST, possibly with no span at all, and format it in text.

I think it makes sense to me

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -121,12 +130,22 @@ impl<'a> Aggregator<'a> {
...
+                    .pse
+                    .first()
+                    .map_or(0, |expr| expr.span().start_column.saturating_sub(1)))
+                    as usize
+            }
+        };
+        let previous_indentation = match self.source {
+            Some(source) => &source[..indentation_level],
+            None => &" ".repeat(indentation_level),
+        };
```
hugocaillard, line 146: This is what I came up with following your previous review @tippenein 
This is a bit involved; I'm not sure we would encounter a use case where we need code section formatting without a text source. But for completeness, we have it

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1981,4 +2000,24 @@ mod tests_formatter {
...
+    }
+
+    #[test]
+    fn format_ast_without_source_handle_indentation() {
+        let src = "  (begin (+ 1 2) (ok true))\n";
+        let ast = clarity::vm::ast::parser::v2::parse(src).unwrap();
+        let formatter = ClarityFormatter::new(Settings::default());
+        let expected = formatter.format_file(src);
+        let result = formatter.format_ast(&ast);
+        assert_eq!(result + "\n", expected);
```
hugocaillard, line 2021: I'm not exactly sure why `expected` gives an extra `\n` (and not `result`).
Let me know if you have an idea, I'll investigate

hugocaillard (commented):
Ready for re-review @tippenein 
Sorry for the force push, I messed up my branch

components/clarinet-format/src/formatter/mod.rs
tippenein: It's possible I misunderstood what "without a text source" means. Looks ok to me

components/clarinet-format/src/formatter/mod.rs
tippenein: It's because of the `format_file` rather than `format_section`. `format_file` expects to end with an extra newline

tippenein (approved):
I fixed that test to use format_section. So, when I switched that I noticed the (begin was no longer indented. That's because format_file assumes starting column of 1, and also a bug with format_ast where the `format_src_exprs` function doesn't prefix previous_indentation spaces, even though it's accurately input.

components/clarinet-format/src/formatter/mod.rs
hugocaillard: That's what I am currently exploring with the transpiler where I go from TS AST to Clarity AST/PSE (and then AST to clarity source code with the formatter). So there's no Clarity source code

PR #1827, opened 2025-05-21
fix: Use correct `default()` function for initializing LSP options

jbencin:
### Description

By default, all the LSP initialization options are initialized to `false`, which means that we have to add extra config to the editor to get anything working

This is because the struct `InitializationOptions` is using the `default()` generated by `#[derive(Default)]` rather than the one we wrote. This PR removes the derived `default()` 

#### Breaking change?

No

### Example

We will no longer need this extra JSON to editor configs:

```json
  "lsp": {
    "clarity-lsp": {
      "enable_lsp_tasks": true,
      "initialization_options": {
        "completion": true,
        "completion_smart_parenthesis_wrap": true,
        "completion_include_native_placeholders": true,
        "document_symbols": false,
        "formatting": true,
        "go_to_definition": true,
        "hover": true,
        "signature_help": true
      }
    }
  },
``` 

### Checklist

### DIFF
--- a/components/clarity-lsp/src/common/requests/capabilities.rs
+++ b/components/clarity-lsp/src/common/requests/capabilities.rs
@@ -5,7 +5,7 @@ use lsp_types::{
 };
 use serde::{Deserialize, Serialize};
 
-#[derive(Debug, Clone, Deserialize, Serialize, Default)]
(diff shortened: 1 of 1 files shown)


hugocaillard (approved):
ah... good catch, thanks @jbencin

PR #1832, opened 2025-05-30
chore: Apply `clippy::manual_let_else`

jbencin:
### Description

Apply Clippy lint `manual_let_else`, which finds `match` statements that can be simplified to `let`/`else` statements. This change results in almost 200 fewer LOC in the repository

#### Breaking change?

No

### Example

For example, this code:
```rust
let docker = match &self.docker_client {
    Some(ref docker) => docker,
    _ => return Err("unable to get Docker client".into()),
};
```

Simplifies to:

```rust
let Some(docker) = &self.docker_client else {
    return Err("unable to get Docker client".into());
};
```

### DIFF
--- a/components/clarinet-cli/src/deployments/mod.rs
+++ b/components/clarinet-cli/src/deployments/mod.rs
@@ -52,9 +52,8 @@ fn get_deployments_files(
     let mut project_dir = project_root_location.clone();
     let prefix_len = project_dir.to_string().len() + 1;
     project_dir.append_path("deployments")?;
-    let paths = match fs::read_dir(project_dir.to_string()) {
-        Ok(paths) => paths,
-        Err(_) => return Ok(vec![]),
+    let Ok(paths) = fs::read_dir(project_dir.to_string()) else {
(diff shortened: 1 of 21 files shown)


hugoclrd:
The failing test is just a flaky one
`called `Result::unwrap()` on an `Err` value: "http error - status: 429 Too Many Requests - message: {\n  \"message\":\"API rate limit exceeded\",\n  \"request_id\":\"4872c51d3fb6b3669ff4fe2deef50d1e\"\n}"`

The api calls should use the provided API key, and not hit a 429.
So it's either:
- not using the key
- really hitting a rate limit when all tests are running in parrallel with the same key

We'll have to look it, but for this PR, you can just restart the tests

jbencin:
I re-ran the tests and it passed, but let's make an issue for any flaky tests and fix ASAP. We had a major problem with those in the stacks-core repo and it really slowed down development with all the time we spent re-running and looking into failing tests

hugocaillard (approved):
lgtm! thanks Jeff

Issue #878, opened 2023-02-09
Clarinet `requirements add` handling arguments weakly


mefrem:
Currently, Clarinet can add a project contract requirement by passing a principal contract in the cli like so:

`clarinet requirements add SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait`

What's not yet supported is passing trait references, like:
`clarinet requirements add SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait.nft-trait`

Furthermore, as a bonus: It would be great to clarity or correct this hint/help text returned `clarinet requirements add --help`

<img width="769" alt="Screenshot 2023-02-09 at 3 26 39 PM" src="https://user-images.githubusercontent.com/51917427/217942861-f940420b-0e03-45e1-a482-0d937bcd915d.png">

Currently, passing `clarinet requirements add "SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait"` (note the double quotes, like as indicated in the help dialog) does not yield a correctly imported requirement. I had to pass the principal contract reference with **no double quotes** (or even **the single quote prefix** which is the Clarity type convention). The request to this issue would be to have a better help dialog without the quotes, or have an input formatter that can take a principal contract reference (or trait reference) with any style or quantity of quotes and find the correct contract on the Stacks chain to place inside the `requirements-publish` contract for the deployment plan.

hugocaillard:
My approach for this one would be to simply update the `--help` doc and remove the `""`

PR #1834, opened 2025-06-01
fix(cli): Improve argument handling for `requirements add`

fabiocarlos97:
### Description

- Closes: #878 where the clarinet requirements add command handled arguments weakly, particularly around quoted identifiers and trait references.

This PR implements the following improvements:
- **Flexible Identifier Parsing:** The contract_id argument for clarinet requirements add now correctly handles:
Identifiers with or without surrounding single (') or double (") quotes.
Trailing or leading whitespace around the identifier.
Trait references, including nested trait references (e.g., SP...contract.trait1.trait2). The full identifier string is now preserved and stored in Clarinet.toml.
- **Updated Help Text**: The help text for the contract_id argument in clarinet requirements add has been updated to reflect the supported formats and remove the erroneous suggestion to always use double quotes.

### Checklist

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -1012,28 +1012,27 @@ pub fn main() {
                 let manifest = load_manifest_or_exit(cmd.manifest_path, true);
 
                 // Trim quotes from contract_id
-                let mut contract_id = cmd.contract_id.trim();
-                contract_id = contract_id.trim_matches('"');
-                contract_id = contract_id.trim_matches('\'');
-
-                let parts: Vec<&str> = contract_id.split('.').collect();
-                let final_contract_id = if parts.len() == 3 {
-                    format!("{}.{}", parts[0], parts[1])
-                } else {
-                    contract_id.to_string()
+                let contract_id = {
+                    let cleaned = cmd.contract_id.trim().trim_matches('"').trim_matches('\'');
+                    let parts: Vec<&str> = cleaned.split('.').collect();
+                    if parts.len() == 3 {
+                        format!("{}.{}", parts[0], parts[1])
+                    } else {
+                        cleaned.to_string()
+                    }
                 };
 
                 let change = TOMLEdition {
                     comment: format!(
                         "{} with requirement {}",
                         yellow!("Updated Clarinet.toml"),
-                        green!(format!("{}", final_contract_id))
+                        green!(format!("{}", contract_id))
                     ),
                     manifest_location: manifest.location,
                     contracts_to_rm: vec![],
                     contracts_to_add: HashMap::new(),
                     requirements_to_add: vec![RequirementConfig {
-                        contract_id: final_contract_id.to_string(),
+                        contract_id: contract_id.to_string(),
                     }],
                 };
                 if !execute_changes(vec![Changes::EditTOML(change)]) {
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -259,7 +259,7 @@ struct RemoveContract {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 struct AddRequirement {
-    /// Contract id (ex. SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait or SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait.nft-trait), optionally enclosed in single or double quotes.
+    /// Contract id (ex. SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait)
     pub contract_id: String,
     /// Path to Clarinet.toml
     #[clap(long = "manifest-path", short = 'm')]
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -1016,17 +1016,24 @@ pub fn main() {
                 contract_id = contract_id.trim_matches('"');
                 contract_id = contract_id.trim_matches('\'');
 
+                let parts: Vec<&str> = contract_id.split('.').collect();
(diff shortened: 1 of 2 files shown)


components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -259,7 +259,7 @@ struct RemoveContract {
 
 #[derive(Parser, PartialEq, Clone, Debug)]
 struct AddRequirement {
-    /// Contract id (ex. "SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait")
+    /// Contract id (ex. SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait or SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait.nft-trait), optionally enclosed in single or double quotes.
```
hugoclrd:
```suggestion
    /// Contract id (ex. SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait)
```

Even though it handles multiple formats, I think the documentation should remain concise

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1011,17 +1011,22 @@ pub fn main() {
             Requirements::AddRequirement(cmd) => {
                 let manifest = load_manifest_or_exit(cmd.manifest_path, true);
 
+                // Trim quotes from contract_id
+                let mut contract_id = cmd.contract_id.trim();
```
hugoclrd: Does it support the notation with trait contract-id + name `.nft-trait.nft-trait` ?

hugocaillard (changes requested):
Thanks a lot @fabiocarlos97 
Added a comment and a question.

I realise that this part of the code isn't tested at all, but we recently increased our standards in terms of unit testing and code coverage.
As such, I opened this PR #1836 that add some tests to the CLI, included `requirements add`
Once this PR is merged (feel free to give it a review), would you be ok to rebase your PR on top of it and add the test scenarios that your PR supports?

components/clarinet-cli/src/frontend/cli.rs
fabiocarlos97: Hi @hugocaillard,
Thanks for the guidance! I've rebased my PR on top of the latest main (which includes #1836).
I've also added the following test scenarios to components/clarinet-cli/tests/cli.rs as you requested:
`test_requirement_add_trait_reference`: This test verifies adding requirements using the PRINCIPAL.CONTRACT-NAME.TRAIT-NAME notation.
`test_requirement_add_with_quotes`: This test ensures that requirements are correctly processed even if the contract/trait identifier is surrounded by double quotes.
All tests, including these new ones, are passing in my development environment (GitHub Codespace).
So, yes, it does support this notation, and it also handles cases where the input might be quoted, as demonstrated by the passing tests.
Please let me know if there's anything else needed!

hugocaillard (changes requested):
Thanks @fabiocarlos97 
The `trim` part is looking good.

But the second part of the PR still requires changes.
If add a trait contract with the `<address>.<contract-name>.<trait-name>` notation like so:
```
clarinet requirements add SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait
```

It does end up in the requirements
```toml
[[project.requirements]]
contract_id = 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait'
```

But Clarinet doesn't know what to do with it, because it expects on the ``<address>.<contract-name>` notation in the TOML.

So when trying to use the contract in a smart contract, clarinet doesn't find the requirement.
The trait-name should be removed from the toml

fabiocarlos97:
> Thanks @fabiocarlos97 The `trim` part is looking good.
> 
> But the second part of the PR still requires changes. If add a trait contract with the `<address>.<contract-name>.<trait-name>` notation like so:
> 
> ```
> clarinet requirements add SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait
> ```
> 
> It does end up in the requirements
> 
> ```toml
> [[project.requirements]]
> contract_id = 'SP3FBR2AGK5H9QBDH3EEN6DF8EK8JY7RX8QJ5SVTE.sip-010-trait-ft-standard.sip-010-trait'
> ```
> 
> But Clarinet doesn't know what to do with it, because it expects on the ``
> .` notation in the TOML.
> 
> So when trying to use the contract in a smart contract, clarinet doesn't find the requirement. The trait-name should be removed from the toml

Hey, that was an oversight on my part. I've updated it, now it should be fine. Please have a look

fabiocarlos97:
@hugocaillard  Hey, do you have any updates?

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1011,17 +1011,29 @@ pub fn main() {
...
+                contract_id = contract_id.trim_matches('"');
+                contract_id = contract_id.trim_matches('\'');
+
+                let parts: Vec<&str> = contract_id.split('.').collect();
+                let final_contract_id = if parts.len() == 3 {
+                    format!("{}.{}", parts[0], parts[1])
+                } else {
+                    contract_id.to_string()
+                };
```
hugoclrd: I'd like to simplify a bit further.
- chain the `trim`
- put everything in its own context

And I think we'll be good


```suggestion

                let contract_id = {
                    let cleaned = cmd.contract_id.trim().trim_matches('"').trim_matches('\'');
                    let parts: Vec<&str> = cleaned.split('.').collect();
                    if parts.len() == 3 {
                        format!("{}.{}", parts[0], parts[1])
                    } else {
                        cleaned.to_string()
                    }
                };
```

fabiocarlos97: I've changed as you've suggested 👍

fabiocarlos97:
@fabiocarlos97 hey please check now

PR #1836, opened 2025-06-02
tests: basic cli tests

hugocaillard:
Implement some basic unit tests for:

- clarinet new
- clarinet contract new
- clarinet requirements add

### DIFF
--- a/components/clarinet-cli/tests/cli.rs
+++ b/components/clarinet-cli/tests/cli.rs
@@ -88,7 +88,7 @@ fn test_contract_new() {
     ];
     let stdout = String::from_utf8_lossy(&output.stdout);
     let stdout_lines: Vec<_> = stdout.lines().map(str::trim).collect();
-    let actual = stdout_lines[..expected_lines.len()].to_vec();
+    let actual = &stdout_lines[..expected_lines.len()];
     assert_eq!(actual, expected_lines);
 
     let contract_path = project_path
--- a/components/clarinet-cli/tests/cli.rs
+++ b/components/clarinet-cli/tests/cli.rs
@@ -1,27 +1,24 @@
-use std::{path::Path, process::Command};
+use std::{fs, path::Path, process::Command};
 
 use clarinet_files::{FileLocation, ProjectManifest, ProjectManifestFile};
 
 #[track_caller]
 fn parse_manifest(project_dir: &Path) -> ProjectManifest {
     let manifest_path = project_dir.join("Clarinet.toml");
-    let manifest_str = std::fs::read_to_string(&manifest_path).unwrap();
-    let manifest_file: ProjectManifestFile =
-        toml::from_str(&manifest_str).expect("failed to parse Clarinet.toml");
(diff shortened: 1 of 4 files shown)


components/clarinet-cli/tests/cli.rs
```diff
@@ -0,0 +1,124 @@
...
+        "Created file .vscode/tasks.json",
+        "Created file .gitignore",
+        "Created file .gitattributes",
+        "Created file package.json",
+        "Created file tsconfig.json",
+        "Created file vitest.config.js",
+    ];
+    let stdout = String::from_utf8_lossy(&cmd.stdout);
+    let stdout_lines: Vec<_> = stdout.lines().map(str::trim).collect();
+    let actual = stdout_lines[..expected_lines.len()].to_vec();
```
jbencin: No need to allocate a `Vec`:

```suggestion
    let actual = &stdout_lines[..expected_lines.len()];
```

hugoclrd:
thx @jbencin

components/clarinet-cli/tests/cli.rs
```diff
@@ -0,0 +1,124 @@
...
+    assert_eq!(actual, expected_lines);
+
+    let project_path = temp_dir.path().join(project_name);
+    assert!(project_path.is_dir(), "Project directory missing");
+    let clarinet_toml = project_path.join("Clarinet.toml");
+    assert!(clarinet_toml.is_file(), "Clarinet.toml missing");
+
+    let manifest_str = fs::read_to_string(&clarinet_toml).expect("Failed to read Clarinet.toml");
+    let expected = format!("[project]\nname = \"{}\"", project_name);
```
jbencin: Can we use [`indoc`](https://github.com/dtolnay/indoc) for multiline strings? It's already in our `Cargo.lock` file. It's so much easier to read these with proper whitespace instead of `\n`s

components/clarinet-cli/tests/cli.rs
```diff
@@ -0,0 +1,124 @@
...
+        "Created file .gitignore",
+        "Created file .gitattributes",
+        "Created file package.json",
+        "Created file tsconfig.json",
+        "Created file vitest.config.js",
+    ];
+    let stdout = String::from_utf8_lossy(&cmd.stdout);
+    let stdout_lines: Vec<_> = stdout.lines().map(str::trim).collect();
+    let actual = &stdout_lines[..expected_lines.len()];
+    assert_eq!(actual, expected_lines);
```
jbencin, line 59: Not sure if it makes sense to match the text output, how about checking that each of these files/directories exists and is not empty?

hugoclrd: I think it would make sense to test both. This is testing that we do give feedback to the users about what's happening.
But I agree that it would also make sense to check that the files exist, I'll look into it

hugoclrd: Added some checks to make sure the files exist and are not empty 👍

PR #1848, opened 2025-06-04
chore: Apply some Clippy lints

jbencin:
### Description

Apply some Clippy lints:

| Clippy lint | Details |
| ------------|---------|
| `bool_to_int_with_if` | Find and replace `if condition { 1 } else { 0 }` |
| `manual_assert` | Replace `if condition { panic!() }` with `assert!()` |
| `manual_string_new` | Replace `"".to_string()` with `String::new()` |
| `needless_continue` | Remove `continue` statements which have no effect |
| `needless_pass_by_value` | Pass by reference when appropriate, for example pass `&str` instead of `String` |

I only partially applied `needless_pass_by_value` since I didn't want this PR to result in a massive diff. I focused on large objects or functions that were called a lot. I replaced a lot of `String`s with `&str`, which got rid of several unnecessary heap allocations

#### Breaking change?

No

### DIFF
--- a/components/clarinet-cli/src/devnet/package.rs
+++ b/components/clarinet-cli/src/devnet/package.rs
@@ -14,43 +14,35 @@ pub struct ConfigurationPackage {
     project_manifest: ProjectManifest,
 }
 
-fn pack_to_file(file_name: &str, package: ConfigurationPackage) -> Result<(), String> {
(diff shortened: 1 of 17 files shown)


hugoclrd:
Closing for now. For too many conflict and inactivity.
We may want to re-assess this PR in a near future

Issue #647, opened 2022-11-07
Trim trailing whitespace on console command parsing


jeremyjs:
Weird issue, I noticed if there's a trailing space on some commands, it throws an error. Probably just trim whitespace on the internal parsing.

```
>> ::set_tx_sender 'ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5
Usage: ::set_tx_sender <address>
```

hashpalk:
Can I work on this one?

hashpalk:
Can I work on this one, @hugocaillard ?

hugoclrd:
Hello @hashpalk, yes, good pick! Thanks

PR #1849, opened 2025-06-04
fix: trim whitespace in set_tx_sender command parsing

hashpalk:
## Problem
The `::set_tx_sender` command was failing when provided with addresses containing trailing whitespace. This caused unnecessary errors for users when copying/pasting addresses or when accidentally including trailing spaces. #647

## Solution
Added whitespace trimming to the `parse_and_set_tx_sender` function to handle both leading and trailing whitespace in the address argument. This makes the command more user-friendly while maintaining the existing behavior for other command parsing functions.

## Testing
The fix can be tested with commands like:
```clarity
::set_tx_sender 'ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5 '  # with trailing space
::set_tx_sender ' ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5'  # with leading space
```

Both should now work correctly, whereas previously they would fail with a usage error.

## Impact
This is a minimal, focused change that only affects the `::set_tx_sender` command parsing. No other command behaviors are modified.

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -870,7 +870,7 @@ impl Session {
     }
 
     fn parse_and_set_tx_sender(&mut self, command: &str) -> String {
-        let args: Vec<_> = command.split(' ').collect();
+        let args: Vec<_> = command.split(' ').map(|s| s.trim()).collect();
 
         if args.len() != 2 {
             return format!("{}", "Usage: ::set_tx_sender <address>".red());


tippenein:
@hashpalk thanks for submitting your first PR! You're on the right path here, but the `trim` is not doing what you'd expect. The args look like `["::set_tx_sender", "ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5", ""]`. I think the easiest way would be to filter out the empty strings. what you probably want instead is `.filter(|&s| !s.is_empty()` to filter those out, leaving only the sender address.

Feel free to pattern the tests at the bottom of this file and add a small one for `set_tx_sender`. Let me know if you need anything else 👍

tippenein:
fixed in #1861

PR #1850, opened 2025-06-04
fix: Resolve duplicate contract name conflicts in requirements

### LINKED ISSUE #948: Duplicate contracts when remapping requirements
When a project has requirements, clarinet handles deploying them and then remapping the original deployer address to the deployer address. This becomes a problem when two requirement contracts have the same name. The second one to deploy will be rejected by the mempool, because there is already a contract deployed with that name. This can be reproduced by adding both the mainnet and the testnet `nft-trait`contracts as requirements in a project:

```
clarinet requirements add SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait
clarinet requirements add ST1NXBK3K5YYMD6FD41MVNP3JS1GABZ8TRVX023PT.nft-trait
```

To avoid this, clarinet could deploy each requirement from a newly generated address, instead of using the default deployer address for all of them.

willyross:
Problem:
When a project has multiple requirement contracts with the same name (e.g. `nft-trait` from both mainnet and testnet), the second deployment fails because it tries to use the same contract name under the default deployer address.

Solution:
- Append the original issuer's address to the contract name to ensure uniqueness
- Keep using the default deployer address for deployment
- Maintain all existing remapping functionality

Example:
```bash
# Original commands that would fail
clarinet requirements add SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait
clarinet requirements add ST1NXBK3K5YYMD6FD41MVNP3JS1GABZ8TRVX023PT.nft-trait

# Now deploys as:
 default_deployer.nft-trait-SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9
 default_deployer.nft-trait-ST1NXBK3K5YYMD6FD41MVNP3JS1GABZ8TRVX023PT
```

This change ensures that:
1. Each requirement contract has a unique name
2. The original contract identity is preserved in the name
3. All contracts are still deployed under the default deployer
4. The remapping system continues to work as expected

Fixes #948

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -523,8 +523,18 @@ pub async fn generate_default_deployment(
                             _ => {}
                         }
 
+                        // Create unique contract name by appending issuer address
+                        let unique_contract_name = format!("{}-{}", 
+                            contract_id.name,
+                            contract_id.issuer.to_address().replace(".", "-")
+                        );
+                        let unique_contract_id = QualifiedContractIdentifier::new(
+                            default_deployer_address.clone(),
+                            ContractName::try_from(unique_contract_name).unwrap(),
+                        );
+
                         let data = RequirementPublishSpecification {
-                            contract_id: contract_id.clone(),
+                            contract_id: unique_contract_id,
                             remap_sender: default_deployer_address.clone(),
                             source: source.clone(),
                             location: contract_location,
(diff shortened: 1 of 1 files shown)


hugoclrd:
@willyross 
Does it mean that, in the code, we would call the contracts like so:

```clar
(use-trait nft-trait '<deployer>.nft-trait-SP2PABAF9FTAJYNFZH93XENAJ8FVY99RRM50D2JG9.nft-trait)
(use-trait nft-trait '<deployer>.nft-trait-ST1NXBK3K5YYMD6FD41MVNP3JS1GABZ8TRVX023PT.nft-trait)
```

If so, this might not be the ideal solution

hugoclrd:
Hello @willyross 
Have you seen my previous comment?

hugoclrd:
I'm closing this stale PR. It looks like it's close but might not work as it is right.
@willyross Please feel free to re-open and review/address my previous comment.

Thanks

Issue #1846, opened 2025-06-03
New Explorer SSR broken in Devnet


hugocaillard:
**Describe the bug**

The new explorer doesn't work with Clarinet devnet.
We are using the latest image.

A hot fix would be to pin an image of the older version of the devnet.
We'll take a day or two to try figuring out a fix first.

Start the devnet in any Clarinet project, or a new one:

```console
$ clarinet new test-proj
$ cd test-proj
$ clarinet devnet start
```

You should see the terminal UI and the containers spinning up
Go to `localhost:3999`, on the devnet you should see:

<img width="1147" alt="Image" src="https://github.com/user-attachments/assets/315f4cbe-a4ef-46e7-965e-1ad03c2438c0" />

It should be fixable with en var, we want the backend to call the docker url, something like [that](https://github.com/hirosystems/clarinet/blob/d5eaf95a3d6fbd483164880f6d0a0d14b0bb5286/components/stacks-network/src/orchestrator.rs#L2188-L2191):
```rust
format!(
    "NEXT_PUBLIC_REGTEST_API_SERVER=http://stacks-api.{}:{}",
    self.network_name, devnet_config.stacks_api_port
),
```

But the front should call `localhost:3999`?

hugoclrd:
Related issue in Explorer repo: https://github.com/hirosystems/explorer/issues/2312

PR #1853, opened 2025-06-05
fix: use old explorer in devnet

hugocaillard:
### Description

Related to
- Clarinet issue: #1846 
- Explorer issue: https://github.com/hirosystems/explorer/issues/2312

As a quick fix, Clarinet devnet will use the old Explorer until it can use the new one.


andresgalante (approved):
@He1DAr let's send a PR to remove when we fix https://github.com/hirosystems/explorer/issues/2312

PR #1856, opened 2025-06-06
fix: leading newlines formatting problem

tippenein:
The presence of leading newlines was messing up file formatting. This fixes it and adds a golden test to avoid regressions


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -68,12 +68,20 @@ impl ClarityFormatter {
...
+        let leading_newlines = &source[0..leading_newlines_count];
+
+        let trimmed_source = source.trim_start_matches(['\n', '\r']);
+        let pse = clarity::vm::ast::parser::v2::parse(trimmed_source).unwrap();
+        let agg = Aggregator::new(&self.settings, &pse, Some(trimmed_source));
         let result = agg.generate();
 
         // make sure the file ends with a newline
-        result.trim_end_matches('\n').to_string() + "\n"
+        leading_newlines.to_string() + &result.trim_end_matches('\n').to_string() + "\n"
```
jbencin: I don't see any need to preserve and reattach the leading newlines. Even if we did, multiple newlines should be compressed into a single one, like how it's done in the rest of the file. IMO you can just trim all whitespace from the beginning and end, then add a `\n` when you write it out.

Also the `to_string()` calls here are creating unnecessary intermediate allocations

What's our policy on newlines? Do we always use `\n`, or try to preserve carriage returns (`\r`) if they are present? If the latter, we should be consistent and add one to the end of the file if detected

components/clarinet-format/tests/golden-intended/ntt-manager-v1.clar
```diff
@@ -0,0 +1,261 @@
+
```
jbencin: Do we need this entire file here, or can we do a minimal example with a leading newline and a map, and name it `leading-newline.clar` so we know what it's testing?

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -68,12 +68,13 @@ impl ClarityFormatter {
...
-        let pse = clarity::vm::ast::parser::v2::parse(source).unwrap();
-        let agg = Aggregator::new(&self.settings, &pse, Some(source));
+        let trimmed_source = source.trim_start_matches(['\n', '\r']);
+        let pse = clarity::vm::ast::parser::v2::parse(trimmed_source).unwrap();
+        let agg = Aggregator::new(&self.settings, &pse, Some(trimmed_source));
         let result = agg.generate();
 
         // make sure the file ends with a newline
-        result.trim_end_matches('\n').to_string() + "\n"
+        format!("{}\n", result.trim_end_matches('\n'))
```
jbencin: I think you should trim the `\r`s off the end also

components/clarinet-format/tests/golden-intended/ntt-manager-v1.clar
tippenein: I'd like to keep it (at least the parts that are broken with the leading newline) but I can definitely rename the file

components/clarinet-format/src/formatter/mod.rs
tippenein: Agreed, we can just trim leading newlines 👍 
Also, Made an issue for autodetecting newline style: https://github.com/hirosystems/clarinet/issues/1862
Good call

tippenein:
#1862 will handle the newline detecting more completely

PR #1857, opened 2025-06-06
feat: Add optional LSP logging to file

jbencin:
### Description

Add optional LSP logging to file. The log file and the filter can be set by the following optional fields in the `initialization_options` struct:

```json
"initialization_options": {
  "logFile": "/path/to/clarity-lsp.log",
  "logLevel": "<DEBUG|INFO|WARN|ERROR>"
}
```

#### Breaking change?

No

### TODO

- [x] Fix formatting of `SystemTime`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -573,9 +573,9 @@ dependencies = [
 
 [[package]]
 name = "chrono"
-version = "0.4.40"
+version = "0.4.41"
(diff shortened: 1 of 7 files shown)


jbencin:
I've learned how to view the RPC messages to/from the LSP in Zed without adding code, so this is unnecessary now

PR #1858, opened 2025-06-09
chore: Apply `clippy::iter_with_drain`

jbencin:
### Description

Apply `clippy::iter_with_drain`, which finds calls to `.drain(..)` which can be replaced by `into_iter()`. `into_iter()` is more efficient because it doesn't have to clear the collection being iterated over, and allows reuse of allocated memory

#### Breaking change?

No

### Example

This code:

```rust
for (account, token, value) in accounts_to_credit.drain(..) {
    self.credit_token(account, token, value);
}
```

Can be simplified to (note that `into_iter()` is implicit):

```rust
for (account, token, value) in accounts_to_credit {
    self.credit_token(account, token, value);
}
```

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -419,28 +419,22 @@ impl Session {
     pub fn start(&mut self) -> Result<(String, Vec<(ContractAnalysis, String, String)>), String> {
         let mut output_err = Vec::<String>::new();
         let output = Vec::<String>::new();
-        let contracts = vec![];
+        let contracts = vec![]; // This is never modified, remove?
 
         if !self.settings.include_boot_contracts.is_empty() {
             self.load_boot_contracts();
         }
 
-        if !self.settings.initial_accounts.is_empty() {
-            let mut initial_accounts = self.settings.initial_accounts.clone();
-            for account in initial_accounts.drain(..) {
-                let Ok(recipient) = PrincipalData::parse(&account.address) else {
-                    output_err.push("Unable to parse address to credit".red().to_string());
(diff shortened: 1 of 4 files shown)


components/clarity-repl/src/repl/session.rs
```diff
@@ -419,28 +419,22 @@ impl Session {
     pub fn start(&mut self) -> Result<(String, Vec<(ContractAnalysis, String, String)>), String> {
         let mut output_err = Vec::<String>::new();
         let output = Vec::<String>::new();
-        let contracts = vec![];
+        let contracts = vec![]; // This is never modified, remove?
```
tippenein, line 422: 👍 remove it

jbencin: `output` right above it is never written to either, so I'll remove that too.

It looks like maybe we intended to do something with these variables though. And I'm not sure if `start()` is used anywhere outside of this repo though, so IDK if this is a breaking API change. I'm going to flag @hugocaillard for an approval here (just check the `start()` function please!)

jbencin: Changing the function signature to remove the empty returned variables broke some tests, so I reverted the changes by force-pushing an older commit. If I change this it will be a separate PR, I don't want to do too much here

jbencin: > broke some tests

Never mind, it was just flaky tests, but still gonna do a separate PR for this

tippenein: yeah, I've been getting flaky failures too ☹️ 
sounds good tho 👍

Issue #1829, opened 2025-05-28
Native LSP: Go to definition/declaration/etc. doesn't work


jbencin:
**Describe the bug**

None of the "Go to" functionality is currently working in the native LSP (only tested in Zed so far)

**To Reproduce**
1. Open Clarity file in editor configured to use Native LSP
2. Find a variable or function
3. Right click and choose "Go to definition/declaration/implementation"
4. Nothing will happen

**Expected behavior**
It should move the cursor to the definition/declaration/implementation

**Environment (please complete the following information):**
 - OS (version): Manjaro Linux
 - Clarinet version: 6a588a6a
 - Text editor: Zed 0.188.3

PR #1859, opened 2025-06-10
fix: "Go to definition" functionality for native LSP

jbencin:
### Description

Fixes: #1829

This PR fixes "Go to definition" functionality for native LSP, which is broken because the following code:

```rust
Url::parse(&definition_contract_location.to_string()).ok()
```

Always returns `None` for `FileLocation::FileSystem` paths (because the `to_string()` output is missing the `file://` prefix). This is fixed by implementing `TryInto<Url>` for `FileLocation` to do the correct type conversion

#### Breaking change?

I don't think this will affect the VSCode extension (I'm guessing it always uses `FileLocation::Url`)

### Notes

In a future PR I think we can get rid of `FileLocation` enum completely and just use `Url` directly

### DIFF
--- a/components/clarinet-files/src/lib.rs
+++ b/components/clarinet-files/src/lib.rs
@@ -62,7 +62,18 @@ impl fmt::Display for FileLocation {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             FileLocation::FileSystem { path } => write!(f, "{}", path.display()),
-            FileLocation::Url { url } => write!(f, "{}", url),
+            FileLocation::Url { url } => write!(f, "{url}"),
+        }
+    }
+}
+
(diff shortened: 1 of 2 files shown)


hugoclrd:
> I don't think this will affect the VSCode extension (I'm guessing it always uses FileLocation::Url)


Right, it shouldn't break.
You can give it a try. In `component/clarity-vscode` run `npm i && npm run dev`. It'll open a test project in vscode web.

 > In a future PR I think we can get rid of FileLocation enum completely and just use Url directly
 
 Yes! I agree that the FileLocation is probably not necessary and adds useless complexity

jbencin:
> Right, it shouldn't break

Tried this and goto functionality is still working in VSCode. I don't think it ever uses filesystem paths. Actually, the parts of the `Url` crate dealing with the filesystem don't even build for this target because it's a `no_std` environment

PR #1860, opened 2025-06-10
feat: Add release build for `x86_64-unknown-linux-musl`

jbencin:
### Description

Add target `x86_64-unknown-linux-musl` for Clarinet release build. Having a static binary will simplify Linux packaging

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -282,7 +282,9 @@ jobs:
         uses: docker://rust:alpine
         env:
           RUSTFLAGS: -C target-feature=-crt-static
-        run: cargo build --release --features=telemetry --locked --target ${{ matrix.target }}
+        with:
+          entrypoint: /bin/sh
+          args: -c "cargo build --release --features=telemetry --locked --target ${{ matrix.target }}"
 
       # Steps for Windows Code Signing with DigiCert
       - name: Windows - Setup Certificate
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -272,6 +277,15 @@ jobs:
         if: matrix.target != 'x86_64-unknown-linux-musl'
         run: cargo build --release --features=telemetry --locked --target ${{ matrix.target }}
 
+      - name: Build - Cargo (linux-musl)
+        if: matrix.target == 'x86_64-unknown-linux-musl'
+        uses: docker://rust:alpine
```
hugoclrd: Is this needed?

jbencin: I could also use `apt` and `rustup` to install the cross-build tools directly on Ubuntu, but I was trying to be consistent since this is how it's done for the Devnet JS MUSL build

jbencin: The `rust:alpine` Docker build failed, but the Ubuntu cross build works, so I'm not going to look into the Docker failure. The Ubuntu cross build is simpler anyways, it adds fewer lines to the CI file

jbencin:
I tested the static binary using built by CI and it works. At least `clarinet --help` and `clarinet console` do, but if the binary loads and runs at all it's probably good

PR #1864, opened 2025-06-16
[WIP] ts2clar

### LINKED ISSUE #1783: Transpiler PoC (v0)
The POC is targeting experienced Clarity developers who can handle debugging the generated Clarity output.

- Do some case studies to see how existing projects could have benefited from this
  - [sBTC](https://github.com/hirosystems/ts-transpiler-case-study/pull/1)
- Detail a pretty advanced spec that is still flexible and will change in the future ([ts2clar](https://github.com/hirosystems/ts2clar))

Transpiler features:
- TS parser
- Extract and transpile top-level data struct and function definitions (`(define-*)`)
- Transpile simple expressions and functions body

At the end of this POC, the transpiler should be able to transpile a basic "counter" smart contract and be ready to adopt more complex features.

hugoclrd:
fix: #1783 

## Disclaimers

At this stage, I'm primarily looking for architectural feedback. The current process is to:
  - Convert the TS code into an intermediate representation (IR), mostly to get the top-level definitions
  - Transpile this IR in Clarity PSEs
    - There's a sub-step here that handles function bodies
  - Format the PSEs to Clarity code with the Clarity formatters
--- 

- There are still thousands of syntax decisions to take and nitpicking to do; this is not the focus of this particular PR and will be addressed 1 by 1 on separate GitHub issues or discussions
- This is only one half of the project; the other half is the Clarity standard TS lib definition (available in [this repo](https://github.com/hirosystems/ts-transpiler-case-study/blob/main/types/clarity.d.ts) - private)
  - In TS tooling, the approach is often to let the tooling (such as eslint) handle the type safety, because there are no other TS type implementations than the official JS one (and the recent port in Go). That's the approach of SWC and OXC. They will happily transpile invalid TS code into JS
  - There is currently a similar approach. ts2clar will have some support for basic type-inference at some point, and the error handling for invalid code will need to be greatly improved for v1.


hugo-stacks:
Closing, might re-consider later

PR #1866, opened 2025-06-18
perf: Remove unnecesary `clone()`s and `to_string()`s

jbencin:
### Description

Remove a few unnecessary instances where the entire contract source code or `SymbolicExpression`s were being unnecessarily cloned by `ActiveContractData` functions. Specifically:

- Remove `.to_string()` in `ActiveContractData::new()` (pass in a `String` instead)
- Reorder field initializtion in `ActiveContractData::new()` so that we don't clone `ast.expressions`
- Don't clone the source in `update_clarity_version()`
- Use `.clone_into()` in `update_sources()` to re-use existing memory allocations

Hopefully this will make the LSP feel a bit snappier

### DIFF
--- a/components/clarity-lsp/src/common/requests/signature_help.rs
+++ b/components/clarity-lsp/src/common/requests/signature_help.rs
@@ -90,7 +90,7 @@ mod definitions_visitor_tests {
     use super::get_signatures;
 
     fn get_source_signature(
-        source: &str,
+        source: String,
         position: &Position,
     ) -> Option<Vec<lsp_types::SignatureInformation>> {
         let contract = &ActiveContractData::new(Clarity2, Epoch21, None, source);
@@ -100,7 +100,7 @@ mod definitions_visitor_tests {
     #[test]
     fn get_simple_signature() {
         let signatures = get_source_signature(
-            "(var-set counter )",
+            "(var-set counter )".to_owned(),
             &Position {
                 line: 1,
                 character: 18,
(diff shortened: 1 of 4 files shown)


components/clarity-lsp/src/common/requests/signature_help.rs
```diff
@@ -150,9 +150,9 @@ mod definitions_visitor_tests {
                 continue;
             }
 
-            let src = format!("({} )", &method);
+            let src = format!("({method} )");
```
hugoclrd, line 153: Does clippy does that or is it just a personal preference?
Are there any pros or cons to the two approaches?

hugocaillard (approved):
LGTM!
Thanks @jbencin 

> Hopefully this will make the LSP feel a bit snappier
I'd be curious to bench that. Especially in the vscode extension (since it's slower). Nothing formal but maybe just opening a big contract in vscode and check the timing (with the option `clarity-lsp.debug.logRequestsTimings`).

components/clarity-lsp/src/common/requests/signature_help.rs
jbencin: The ability to inline variables into a format string was added a few years ago in Rust 1.58 ([link](https://blog.rust-lang.org/2022/01/13/Rust-1.58.0/)). AFAIK it doesn't change the generated code, but it's generally what you see in modern Rust code and for me at least it's easier because I can just read left-to-right without looking after the comma

hugoclrd: Ok, I see. I _think_ I tend prefer the "old" way because it's consistent with the other syntaxes and it makes it easier to refactor if for some reason, you need to replace it with an expression like

```rs
format!("{} ", method.to_lowercase());
// or
format!("{} ", get_method());
```

But that's not a strong opinion

Issue #785, opened 2023-01-19
Request: Jump to definition for traits


obycode:
It would be great to make jump to definition work for traits, i.e. cmd+click on a `<trait-name>` or on the trait identifier in `use-trait` and `impl-trait`.

PR #1867, opened 2025-06-20
feat: Go to trait definition for `impl-trait` and `use-trait`

jbencin:
### Description

Fixes: #785

Clicking "Go to Definition" on `impl-trait` and `use-trait` names should go to the corresponding `define-trait` declaration

#### Breaking change?

No

### Status

- [x] Go to trait works for `impl-trait`
- [x] Go to trait works for `use-trait`

### DIFF
--- a/components/clarity-lsp/src/common/requests/definitions.rs
+++ b/components/clarity-lsp/src/common/requests/definitions.rs
@@ -123,16 +123,19 @@ impl<'a> ASTVisitor<'a> for Definitions {
         _name: &'a ClarityName,
         trait_identifier: &TraitIdentifier,
     ) -> bool {
-        eprintln!(
-            "Insert use trait '{_name}' at ({}, {}), defined at {trait_identifier:?}",
-            expr.span.start_line, expr.span.start_column
-        );
+        let Some(keyword) = expr.match_list().and_then(|expr| expr.get(1)) else {
+            return true;
+        };
         self.tokens.insert(
-            (expr.span.start_line, expr.span.start_column),
+            (keyword.span.start_line, keyword.span.start_column),
             DefinitionLocation::External(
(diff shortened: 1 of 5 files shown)


components/clarity-lsp/src/common/requests/definitions.rs
```diff
@@ -578,8 +620,37 @@ pub fn get_public_function_definitions(
...
+            });
+    }
+}
+
+pub fn get_public_function_and_trait_definitions(
+    definitions: &mut HashMap<ClarityName, Range>,
+    expressions: &[SymbolicExpression],
+) {
+    get_public_function_definitions(definitions, expressions);
+    get_trait_definitions(definitions, expressions);
```
hugoclrd, line 653: Would it make sense to factorize `get_public_function_definitions` and `get_trait_definitions` so that we avoid looping over expressions (and run `.match_list()...`) twice?

hugocaillard (commented):
Just a comment/question

components/clarity-lsp/src/common/requests/definitions.rs
jbencin: I thought about doing that, but I didn't think looping over the top-level `SymbolicExpressions` twice was an expensive operation, and I wanted to avoid touching `get_public_function_definitions()` and risk breaking it.

If you want me to I can do it though, not a problem

If we're looking for performance improvements in definition lookups, there's a clone of the entire definitions hashmap in `get_definition_location()` that I'd like to get rid of

hugoclrd: Ok, that makes sense! I like that approach

hugocaillard (approved):
lgtm!

PR #1874, opened 2025-06-26
fix: better error messages for contract call failures

### LINKED ISSUE #1708: Invalid contract calls lead to panic with weird error
**Describe the bug**
Simple invalid contract calls in tests lead to panic and a weird error message.

**To Reproduce**
I've pushed a repo where I can see the error. Simply clone https://github.com/vini-btc/weird-error-message, npm install and run tests.

**Expected behavior**
Some error message should point out the contract and/or method does not exist.

**Screenshots**

Image

 - OS (version): MacOS Sonoma 14.6.1
 - Docker version 24.0.5
 - Clarinet version 2.15.1

**Additional context**
Seems to be the same error message discussed in https://github.com/hirosystems/clarinet/issues/1676, but in this case there are no contract constants anywhere. It was just an invalid call.

tippenein:
- fixes #1708 
Specify if contract calls are to non-existent methods or contracts

Screenshot 2025-06-26 at 10 32 40 AM

### DIFF
--- a/components/clarity-repl/src/repl/interpreter.rs
+++ b/components/clarity-repl/src/repl/interpreter.rs
@@ -33,6 +33,31 @@ use super::remote_data::HttpClient;
 use super::settings::{ApiUrl, RemoteNetworkInfo};
 use super::{ClarityContract, DEFAULT_EPOCH};
 
+#[derive(Debug, Clone)]
+pub enum ContractCallError {
+    NoSuchContract(String),
+    NoSuchFunction(String),
+    Uncategorized(String),
+}
+
+impl std::fmt::Display for ContractCallError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ContractCallError::NoSuchContract(contract_id) => {
+                write!(f, "NoSuchContract({})", contract_id)
+            }
+            ContractCallError::NoSuchFunction(function_name) => {
+                write!(f, "NoSuchFunction({})", function_name)
+            }
+            ContractCallError::Uncategorized(message) => {
+                write!(f, "Uncategorized({})", message)
+            }
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


components/clarity-repl/src/repl/session.rs
```diff
@@ -560,9 +560,21 @@ impl Session {
             Ok(result) => result,
             Err(e) => {
                 self.set_tx_sender(&initial_tx_sender);
+                let mut user_friendly_message =
+                    format!("Error calling contract function '{method}': {e}");
+                if e.contains("Failed to read non-consensus contract metadata") {
```
jbencin: Can we refactor the function that returns this error to return an `enum` error type instead of a `String`? Parsing the text of our own internal errors is a very fragile way to handle this

tippenein: Agreed. I can move it out of this area, but I can't change the string matching since the "Failed to read non-consensus..." message is from stacks-core clarity_db.

components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -33,6 +33,31 @@ use super::remote_data::HttpClient;
...
+    NoSuchContract(String),
+    NoSuchFunction(String),
+    Uncategorized(String),
+}
+
+impl std::fmt::Display for ContractCallError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ContractCallError::NoSuchContract(contract_id) => {
+                write!(f, "NoSuchContract({})", contract_id)
```
jbencin: Clippy 1.88 is going to make you inline these

```suggestion
                write!(f, "NoSuchContract({contract_id})")
```

jbencin (approved):

jbencin (approved):

PR #1878, opened 2025-06-27
chore: Address my comments on PR #1877

jbencin:
### Description

I was reviewing #1877 and had some comments, but the PR was merged before I finished. This PR addresses them. Basically, it does:

- Replace `format!("{var}")` with `var.to_string()`
- Clean up one string using `formatdoc!`
- Fix some eagerly evaluated format strings in `ok_or()`

### DIFF
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -688,7 +688,7 @@ impl SDK {
         session
             .interpreter
             .get_data_var(&contract_id, var_name)
-            .ok_or("value not found".into())
+            .ok_or_else(|| "value not found".into())
     }
 
     #[wasm_bindgen(js_name=getBlockTime)]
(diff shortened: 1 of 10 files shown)


components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -688,7 +688,7 @@ impl SDK {
         session
             .interpreter
             .get_data_var(&contract_id, var_name)
-            .ok_or("value not found".into())
+            .ok_or_else(|| "value not found".into())
```
hugoclrd, line 691: What's the reason to prefer ok_or vs ok_or_else?

hugoclrd:
> the PR was merged before I finished.

😕

components/clarinet-sdk-wasm/src/core.rs
jbencin: `ok_or()` is eagerly evaluated, so even if the condition is not triggered, the format string will be allocated. It's okay to use with literal values like `u32` or `&str`, but you shouldn't do anything expensive like allocate memory inside eagerly evaluated functions like `ok_or()` or `unwrap_or()`, prefer the `*_or_else()` variants instead

jbencin: There's even a warning about it in the official Rust docs: https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or

PR #1882, opened 2025-06-30
add --save-container-logs flag

### LINKED ISSUE #983: Feature request: flag to save docker logs
For `stacks-devnet-js` and `clarinet integrate` it would be nice to have a flag that would save the logs before cleaning up the docker images.

tippenein:
- closes #983 

streams stacks-node.log from the container if flag is enabled


components/clarinet-cli/src/devnet/start.rs
```diff
@@ -22,6 +22,7 @@ pub fn start(
     log_tx: Option<Sender<LogData>>,
     display_dashboard: bool,
     no_snapshot: bool,
+    save_container_logs: bool,
```
jbencin: We're passing three `bool`s to this function now, can we refactor to pass these inside a struct?

components/stacks-network/src/lib.rs
```diff
@@ -141,19 +146,19 @@ async fn do_run_devnet(
 
     // The event observer should be able to send some events to the UI thread,
     // and should be able to be terminated
-    let hooks = match chainhooks.take() {
+    let hooks = match config.chainhooks {
```
jbencin: Can replace this `match` statement with `unwrap_or_default()` I think

components/stacks-network/src/orchestrator.rs
```diff
@@ -3510,6 +3521,87 @@ events_keys = ["*"]
...
+        // Start streaming stacks-node logs
+        if let Some(container_id) = &self.stacks_node_container_id {
+            let log_path = PathBuf::from(&devnet_config.working_dir).join("stacks-node.log");
+            let container_id = container_id.clone();
+            let docker = docker.clone();
+            let log_path_clone = log_path.clone();
+
+            // Spawn a background thread to stream logs
+            std::thread::spawn(move || {
+                let rt = tokio::runtime::Runtime::new().unwrap();
```
jbencin: `start_container_logs_streaming()` is an async function so there's already a Tokio runtime, so I don't think it's necessary to spawn another

tippenein: I did try that, but the logs print out a few lines and then it stops streaming. Wasn't sure what caused it, but with this new runtime it continues streaming

tippenein: this seems to work

```
                let rt = tokio::runtime::Builder::new_current_thread()
                    .enable_all()
                    .build()
                    .unwrap();
```
instead of `Runtime::new()`

jbencin: This function returns a `Builder` struct which is also used to create a new runtime: https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html

You should already be in the tokio runtime, so you should be able to use [`tokio::task`](https://docs.rs/tokio/latest/tokio/task/index.html) to start a thread using `spawn` or `spawn_blocking`. Does this not work?

tippenein: It doesn't work for some reason. The log will be printed for a few seconds and then doesn't receive any more even though the docker logs are still going

components/stacks-network/src/orchestrator.rs
```diff
@@ -3476,6 +3439,85 @@ events_keys = ["*"]
...
+
+        // Start streaming stacks-node logs
+        if let Some(container_id) = &self.stacks_node_container_id {
+            let log_path = PathBuf::from(&devnet_config.working_dir).join("stacks-node.log");
+            let container_id = container_id.clone();
+            let docker = docker.clone();
+            let log_path_clone = log_path.clone();
+
+            // Spawn a background thread to stream logs
+            std::thread::spawn(move || {
```
jbencin, line 3457: Have you tried doing this without using `std::thread`? It looks like you are doing a lot here:

- Spawning a new OS thread
- Creating another tokio runtime for that new thread
- Running a blocking task on it

I think if you just cut out the `std::thread` usage and stuck with the tokio API, you could just directly spawn a new blocking task here without all the extra stuff

tippenein: refactor issue is here: https://github.com/hirosystems/clarinet/issues/1905

PR #1884, opened 2025-06-30
perf: Apply `clippy::redundant_clone`

jbencin:
### Description

Apply `clippy::redundant_clone`, which removes several unnecesary `.clone()` and `.to_string()` calls

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -1214,6 +1214,7 @@ pub fn main() {
             );
 
             if cmd.enable_clarity_wasm {
+                #[allow(clippy::redundant_clone)]
                 let mut manifest_wasm = manifest.clone();
                 manifest_wasm.repl_settings.clarity_wasm_mode = true;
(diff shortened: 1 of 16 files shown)


components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1211,6 +1211,7 @@ pub fn main() {
             );
 
             if cmd.enable_clarity_wasm {
+                #[allow(clippy::redundant_clone)]
```
jbencin, line 1217: The clone here is required because `manifest` can be used again if the `telemetry` feature is enabled

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -1436,6 +1436,7 @@ mod tests {
         let current_tip = *clarity_datastore.current_chain_tip.borrow();
         assert_ne!(current_tip, initial_tip);
 
+        #[allow(clippy::redundant_clone)]
```
jbencin, line 1439: The clone here is required because `ClarityDatastore` has a custom `Clone` implementation which is doing something non-standard

PR #1886, opened 2025-07-01
Fix: preserve inner expr newlines (#1790)

### LINKED ISSUE #1790: preserve newlines within expressions
Formatter still removes extra new lines within functions. For example, in this snippet, the new line before `(ok)` is removed:

```clarity
(define-public (increment)
  (begin
    (try! (stx-transfer? (var-get cost) tx-sender (var-get contract-owner)))

    (ok (var-set count (+ (var-get count) u1)))
  )
)
```

tippenein:
- fixes #1790

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -1384,6 +1384,8 @@ mod tests_formatter {
     #[allow(unused_imports)]
     use std::assert_eq;
 
+    use indoc::indoc;
+
     use super::{ClarityFormatter, Settings};
     use crate::formatter::Indentation;
     #[macro_export]
@@ -1436,11 +1438,14 @@ mod tests_formatter {
 
     #[test]
     fn intact_comment_spacing() {
-        let src = r#";; (define-read-only (has-access)
-;;   (begin
-;;     (ok true)
-;;   )
-;; )"#;
+        let src = indoc!(
+            r#"
+          ;; (define-read-only (has-access)
+          ;;   (begin
+          ;;     (ok true)
+          ;;   )
+          ;; )"#
+        );
         let result = format_with_default(&String::from(src));
         assert_eq!(src, result);
 
@@ -1452,13 +1457,16 @@ mod tests_formatter {
     fn test_multi_function() {
         let src = "(define-public (my-func) (ok true))\n(define-public (my-func2) (ok true))";
         let result = format_with_default(&String::from(src));
-        let expected = r#"(define-public (my-func)
-  (ok true)
-)
-(define-public (my-func2)
-  (ok true)
-)
-"#;
+        let expected = indoc!(
+            r#"
+            (define-public (my-func)
+              (ok true)
+            )
+            (define-public (my-func2)
+              (ok true)
+            )
+            "#
+        );
         assert_eq!(expected, result);
     }
     #[test]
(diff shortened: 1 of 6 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1437,6 +1484,19 @@ mod tests_formatter {
...
     #[test]
+    fn test_preserve_newlines_inner_function() {
+        let src = r#"(define-public (increment)
+  (begin
+    (try! (stx-transfer? (var-get cost) tx-sender (var-get contract-owner)))
+
+    (ok (var-set count (+ (var-get count) u1)))
+  )
+)
+"#;
```
jbencin: We've started using the `indoc` crate in some of the other components to keep multiline strings readable and indented with the rest of the code:

```suggestion
        let src = indoc!(r#"
          define-public (increment)
            (begin
              (try! (stx-transfer? (var-get cost) tx-sender (var-get contract-owner)))
              (ok (var-set count (+ (var-get count) u1)))
            )
          )"#;
```

You can add the workspace version in Cargo.toml:

```toml
indoc = { workspace = true }
```

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1331,6 +1367,17 @@ fn chars_since_last_newline(acc: &str) -> usize {
     }
 }
 
+// Helper to insert at most one blank line if there are blank lines between two expressions
+fn push_blank_lines(acc: &mut String, prev_end_line: u32, curr_start_line: u32) {
+    if prev_end_line > 0 && curr_start_line > 0 && curr_start_line > prev_end_line {
```
jbencin: I could be wrong, but it looks like you're using `0` to represent the absence of start_line/end_line here. In Rust you should do this with an `Option<u32>` instead.

tippenein: There wouldn't be an absence of start_line since it's initially populated from the current `span`

components/clarinet-format/src/formatter/mod.rs
tippenein: This is what rustfmt does with formatting indoc ☹️ 

```
    fn intact_comment_spacing() {
        let src = indoc!(
            r#"
          ;; (define-read-only (has-access)
          ;;   (begin
          ;;     (ok true)
          ;;   )
          ;; )"#
        );
        let result = format_with_default(&String::from(src));
        assert_eq!(src, result);
```

tippenein: Made it a separate issue for now. 
https://github.com/hirosystems/clarinet/issues/1891

components/clarinet-format/src/formatter/mod.rs
jbencin: What about end_line? It's just when you see this sort of pattern in Rust code (comparing an integer type to `0`), it usually means you could be using an `Option` type instead

components/clarinet-format/tests/golden.rs
```diff
@@ -61,6 +61,7 @@ fn test_irl_contracts() {
             let src = fs::read_to_string(&path).expect("Failed to read source file");
 
             let file_name = path.file_name().expect("Failed to get file name");
+            println!("file: {file_name:?}");
```
jbencin, line 64: Is this a debug print that should be removed?

tippenein: this is useful to have when testing because if the golden test fails you want to know which file it failed on

components/clarinet-format/src/formatter/mod.rs
tippenein: Changed it to Option and removed the extra `> 0`  checks. Seems like they weren't needed 👍

jbencin (approved):

PR #1889, opened 2025-07-02
refactor: safe epoch spec serialization

hugocaillard:
### Description

Instead of manually creating the `[[burnchain.epochs]]` configs for the devnet stacks_node config, this PR iterates over the `EpochSpec` struct to generate a struct that is then serialized to toml.

### Motivation

In stacks-devnet-api, we jsut found out that the config for epoch 3.1 isn't properly handled:
https://github.com/hirosystems/stacks-devnet-api/blob/ca9cff797874df97085f443efd34d8ff42fba39e/src/lib.rs#L1395-L1397

With this, it would be safer to handle new epochs addition.


jbencin (approved):
Nice Rust code, and I like that we're moving dependency versioning into the top-level Cargo.toml

hugoclrd:
Thanks @jbencin 

> dependency versioning into the top-level Cargo.toml
Yeah we can do that everytime I modify a depenency that is used in >= 2 components (and likely to be used in more)
And soon it'll be clean 👍

PR #1894, opened 2025-07-03
refactor: epoch 3.1 improvements

hugocaillard:
### Description

Small details that caught my eyes when investigating one an epoch 3.1 issue


components/clarinet-files/src/devnet_diff.rs
```diff
@@ -79,6 +79,9 @@ impl DevnetDiffConfig {
                 name: "epoch_3_0".to_string(),
                 extractor: make_extractor(|config| config.epoch_3_0),
             },
+            // With the default snapshot, the devnet starts at epoch 3.0
+            // So users can set custom higher epoch heights
```
hugoclrd, line 83: I thought this comment would be useful so that, next time we add an epoch, we remeber that epochs > 3.0 are _not_ significant fields

PR #1896, opened 2025-07-03
add sender and recipient validation messages

### LINKED ISSUE #1875: repl panicks if invalid address provided to sdk
**Describe the bug**

Ran into this from some generated code, the culprit turned out to be the full contract address getting put into the sender field of `simnet.callPublicFn`.

**To Reproduce**

```ts
const contractName = "ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.aibtc-dao-users"
const daoAddress = "ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.aibtc-base-dao"
it("get-or-create-user-index() can create a user when called by the DAO", () => {
    // act
    const createReceipt = simnet.callPublicFn(
      contractAddress,
      "get-or-create-user-index",
      [Cl.principal(address1)],
      daoAddress
    );
});
```

Example of the resulting panic:
```
stderr | tests/contracts/dao/extensions/aibtc-dao-users.test.ts > DAO-context functions: aibtc-dao-users > increase-user-reputation() updates reputation and preserves createdAt
panicked at components/clarity-repl/src/repl/session.rs:892:62:
Unable to parse address: Runtime(ParseError("Invalid principal literal: Invalid crockford 32 string"), None)

Stack:

Error: 
    at module.exports.__wbg_new_8a6f238a6ece86ea (/home/whoabuddy/Dev/aibtcdev/aibtcdev-daos/node_modules/@hirosystems/clarinet-sdk-wasm/clarinet_sdk.js:1389:17)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:5361479)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:4484843)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:4735064)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:4569274)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:2396801)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:1677396)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:4447022)
    at null.<anonymous> (wasm://wasm/01e6d1aa:1:4737115)
    at SDK.callPublicFn (/home/whoabuddy/Dev/aibtcdev/aibtcdev-daos/node_modules/@hirosystems/clarinet-sdk-wasm/clarinet_sdk.js:665:26)
```

If it could report it was related to the sender parameter or even better highlight the line it came from that'd be ideal!

**Expected behavior**

A clear error message would have been nice here, took a little while to spot the issue and started by iterating over all new implementations of `callPublicFn`.

 - OS (version): Linux Mint
 - Clarinet version: 3.1.0

tippenein:
- fixes #1875 

Validates input a bit better to print out the failing field instead of panicking

### DIFF
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -852,7 +852,7 @@ impl SDK {
             return Err(format!("Invalid sender address '{}'.", args.sender));
         }
 
-        if PrincipalData::parse_standard_principal(&args.recipient).is_err() {
+        if PrincipalData::parse(&args.recipient).is_err() {
             return Err(format!("Invalid recipient address '{}'.", args.recipient));
         }
 
@@ -1088,7 +1088,7 @@ impl SDK {
 
     #[wasm_bindgen(js_name=mintSTX)]
     pub fn mint_stx(&mut self, recipient: String, amount: u64) -> Result<String, String> {
-        if PrincipalData::parse_standard_principal(&recipient).is_err() {
+        if PrincipalData::parse(&recipient).is_err() {
             return Err(format!("Invalid recipient address '{recipient}'."));
         }
 
--- a/components/clarinet-sdk-wasm/src/test_wasm.rs
+++ b/components/clarinet-sdk-wasm/src/test_wasm.rs
@@ -111,19 +111,17 @@ async fn it_handles_invalid_sender_address() {
 }
 
 #[wasm_bindgen_test]
-async fn it_handles_invalid_recipient_address() {
+async fn it_handles_contract_recipient_address() {
     let mut sdk = init_sdk().await;
 
-    // Test with invalid recipient address
     let result = sdk.transfer_stx(&crate::core::TransferSTXArgs::new(
         1000,
-        "ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.invalid-contract".into(), // Invalid: contract address
+        "ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.basic-contract".into(), // valid: contract address
         "ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM".into(),
     ));
 
-    assert!(result.is_err());
-    let error_msg = result.unwrap_err();
-    assert!(error_msg.contains("Invalid recipient address"));
+    // contract addresses are valid recipients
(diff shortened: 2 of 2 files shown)


hugocaillard (changes requested):
I think clippy won't pass after #1895 is merged

hugoclrd:
Rebased to include #1895

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -844,6 +848,14 @@ impl SDK {
         args: &TransferSTXArgs,
         advance_chain_tip: bool,
     ) -> Result<TransactionRes, String> {
+        if PrincipalData::parse_standard_principal(&args.sender).is_err() {
+            return Err(format!("Invalid sender address '{}'.", args.sender));
+        }
+
+        if PrincipalData::parse_standard_principal(&args.recipient).is_err() {
+            return Err(format!("Invalid recipient address '{}'.", args.recipient));
+        }
```
hugoclrd: It should be possible to send stx to contract.

Send _from_ a contract though could be useful from a testing point of view but probably doesn't make sense since it would need to be initiated from a contract call

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -1060,6 +1088,10 @@ impl SDK {
 
     #[wasm_bindgen(js_name=mintSTX)]
     pub fn mint_stx(&mut self, recipient: String, amount: u64) -> Result<String, String> {
+        if PrincipalData::parse_standard_principal(&recipient).is_err() {
+            return Err(format!("Invalid recipient address '{recipient}'."));
+        }
+
```
hugoclrd: mint_stx is used for testing purpose to give stx to any arbitrary address.
I guess it make sense to mint stx to a contract in the simnet (if someone has a use case for it)

components/clarinet-sdk-wasm/src/core.rs
tippenein: I made send-from contract disabled, but maybe that's too restrictive? I guess there's not much issue in using `parse` instead of `parse_standard_principal` ?

tippenein: ah, there are tests for checking send-from with a contract, so I'll change it to `parse`

hugocaillard (approved):
lgtm!

PR #1898, opened 2025-07-03
perf: Fix `O(n^2)` string generation

jbencin:
### Description

Small PR that fixes an `O(n^2)` string building algorithm in `Session::encode()`. Currently we loop over each byte, rewrite and reallocate the entire string to add a new byte each time, which is very inefficient if we have a large number of bytes to write


hugocaillard (approved):
lgtm

PR #1899, opened 2025-07-04
feat: use official stacks node image with multi platform support

hugocaillard:
### Description

fix: #1840

wip

- boot from genesis works
- boot bfrom snapshot does not


components/stacks-network/src/orchestrator.rs
```diff
@@ -220,10 +212,11 @@ impl DevnetOrchestrator {
     }
 
     pub async fn prepare_local_network(&mut self) -> Result<ServicesMapHosts, String> {
-        let (docker, _, devnet_config) = self.docker_and_configs()?;
+        let docker = self.docker_client.as_ref().ok_or(DOCKER_ERR_MSG)?;
+        let devnet_config = self.get_devnet_config()?;
```
hugoclrd, line 216: @tippenein I saw we both addressed the code duplication here. I kept my version when fixing the conflicts because I thought there was too much complexity in the nested `match` that return
```
        (
            &Docker,
            &clarinet_files::NetworkManifest,
            &clarinet_files::DevnetConfig,
        ),
```
        
I feel like we had unnecessary coupling between Docker and devnet_config.
Let me know what you think

tippenein: Yea, I'm fine with either 👍

tippenein (approved):
👍  and nice cleanup as well

components/stacks-network/src/orchestrator.rs
```diff
@@ -888,27 +891,28 @@ rpcport={bitcoin_node_rpc_port}
...
             ));
         }
 
         let mut cmd_args = vec![
             "/usr/local/bin/bitcoind".into(),
             "-conf=/etc/bitcoin/bitcoin.conf".into(),
             "-nodebuglogfile".into(),
-            "-pid=/run/bitcoind.pid".into(),
-            "-datadir=/root/.bitcoin".into(),
+            format!("-pid={BITCOIND_DATA_DIR}/bitcoind.pid"),
```
tippenein, line 907: /run/bitcoind.pid -> /home/.bitcoin/bitcoind.pid ?

hugoclrd:
Thanks @tippenein

PR #1904, opened 2025-07-07
feat: add --create-global-snapshot flag

tippenein:
avoid the possibility of creating global snapshots when running --from-genesis and hitting block 142

this adds the `--create-global-snapshot` flag and removes the previous beforehand

Note: If using this internally, we'd still need to tar.gz the contents and put it in `stacks-network/data`


components/stacks-network/src/chains_coordinator.rs
```diff
@@ -406,15 +407,17 @@ pub async fn start_chains_coordinator(
...
-                            let _ = create_global_snapshot(
-                                &config,
-                                &devnet_event_tx,
-                                mining_command_tx.clone(),
-                            )
-                            .await;
+                        if create_new_snapshot {
+                            // Check if we've reached the target height for database export (142)
+                            // If we've reached epoch 3.0, create the global snapshot
+                            if bitcoin_block_height == config.devnet_config.epoch_3_0 {
```
hugoclrd: Can the ifs be combined?
I tought clippy would complain about that

components/stacks-network/src/chains_coordinator.rs
```diff
@@ -736,11 +739,44 @@ fn should_publish_stacking_orders(
     true
 }
 
+pub async fn remove_global_snapshot(devnet_event_tx: &Sender<DevnetEvent>) -> Result<(), String> {
```
hugoclrd: Does it need to be `pub` ?

tippenein: nope. updated

hugocaillard (approved):
lgtm! thx @tippenein

Issue #1901, opened 2025-07-06
BIP39 mnemonic parsing error in Clarinet 3.2.0 across all environments


rigocrypto:
### **Describe the bug**
Clarinet 3.2.0 fails to parse valid BIP39 mnemonics in settings files (Simnet.toml, Devnet.toml, Testnet.toml), consistently returning "mnemonic (located in ./settings/Simnet.toml) for deploying address is invalid: bip39 error" even when using standard, valid BIP39 mnemonics.

### **To Reproduce**
1. Install Clarinet 3.2.0 on Windows
2. Create a new Clarinet project or use existing project
3. Create `settings/Simnet.toml` with valid BIP39 mnemonic:
   ```toml
   [network]
   name = "simnet"
   
   [accounts.deployer]
   mnemonic = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"
   balance = 100_000_000_000_000
   ```
4. Run `clarinet check` or `clarinet deployments generate`
5. See error: "mnemonic (located in ./settings/Simnet.toml) for deploying address is invalid: bip39 error"

### **Expected behavior**
Clarinet should accept valid BIP39 mnemonics and proceed with contract validation and deployment generation without errors.

### **Screenshots/Console Output**
```
rigocrypto@RigoCrypto:/mnt/c/Users/servi/RigoCrypto/bitmind-chain/bitmind-chain/smart-contracts$ clarinet check
error: mnemonic (located in ./settings/Simnet.toml) for deploying address is invalid: bip39 error
```

### **Environment (please complete the following information):**
- **OS (version):** Windows 10/11 (Build 10.0.26100)
- **Docker version:** N/A (not using Docker)
- **Clarinet version:** 3.2.0 (latest version)

### **Additional context**
- **Environments tested:** PowerShell, Git Bash, WSL2 Ubuntu
- **All environments produce the same error**
- **Multiple valid mnemonics tested:**
  - Standard test mnemonic: "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about"
  - Generated 12-word: "siren mesh lunch swing ride mesh swing ride mesh lunch swing ride"
  - Generated 24-word: "gravity machine north sort system female filter attitude volume fold club stay feature office ecology stable narrow fog split tooth myth federal glove shrug"
- **File encodings tested:** UTF-8, UTF-8 without BOM
- **Line endings tested:** CRLF, LF
[...]

PR #1906, opened 2025-07-07
fix: 12 word mnemonics

### LINKED ISSUE #1888: 12 word mnemonic
I've seen several confused users recently getting stuck because they tried to use a 12 word mnemonic in Testnet.toml or Mainnet.toml. I believe Clarinet only supports 24-word mnemonics currently. Is that an easy change to make? If not, can we make sure that there is a clear error message?

tippenein:
- closes #1888 and #1901 

uses bip39 mnemonics to allow 12 and 24 word while not breaking the current implementation

### DIFF
--- a/components/clarinet-utils/Cargo.toml
+++ b/components/clarinet-utils/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2021"
 bip39 = "2.2.0"
 bip32 = "0.5.3"
 libsecp256k1 = "0.7.0"
-rand = "0.8"
+rand.workspace = true
 
 [lib]
 name = "clarinet_utils"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -29,6 +29,7 @@ pox-locking = { git = "https://github.com/stacks-network/stacks-core.git", branc
 colored = { version = "3" }
 crossterm = { version = "0.28" }
 pretty_assertions = "1.4"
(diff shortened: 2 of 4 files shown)


components/clarinet-utils/Cargo.toml
```diff
@@ -6,8 +6,10 @@ license = "MIT"
 edition = "2021"
 
 [dependencies]
-bip32 =  { version = "0.5.3", features = ["mnemonic"] }
+bip39 = "2.2.0"
+bip32 = "0.5.3"
 libsecp256k1 = "0.7.0"
+rand = "0.8"
```
hugoclrd: Already use in two other packages, can it be a workspace dependency?

hugocaillard (changes requested):

components/clarinet-utils/Cargo.toml
```diff
@@ -6,8 +6,10 @@ license = "MIT"
 edition = "2021"
 
 [dependencies]
-bip32 =  { version = "0.5.3", features = ["mnemonic"] }
+bip39 = "2.2.0"
+bip32 = "0.5.3"
 libsecp256k1 = "0.7.0"
+rand.workspace = true
```
hugoclrd, line 12: Fancy

hugocaillard (approved):
lgtm! thx @tippenein

PR #1907, opened 2025-07-08
feat: add "latest" epoch manifest option

### LINKED ISSUE #1893: Allow epoch "latest" in Clarinet.toml
In the Clarinet.toml (manifest file) contract list, developers can set the epoch of contracts
```toml
[contracts.counter]
clarity_version = 3 # optional
epoch = 3.1 # optional
```

The epoch filed is optional, and it defaults to epoch `"2.0"`. That's because in early version of Clarinet, this field didn't exist, so the absence of the settings means the first epoch.

However, we should encourage developers to set the epoch to the latest, because when they will deploy ac contract on Mainnet, it will be deployed to the current/latest epoch.

We should find a way to explicitely set the epoch of a contract to be the latest.
Could be by accepting the value to be `latest` and when parsing the file, clarinet would set it to the latest epoch.
That should be the default value.

```toml
[contracts.counter]
clarity_version = 3
epoch = 'latest'
```

- [ ] Update manifest epoch handling so that it support a way to set it to latest
- [ ] Update the `generate` project to use this value in new projects

tippenein:
- closes #1893

### DIFF
--- a/components/clarinet-deployments/src/types.rs
+++ b/components/clarinet-deployments/src/types.rs
@@ -35,6 +35,8 @@ pub enum EpochSpec {
     Epoch3_0,
     #[serde(rename = "3.1")]
     Epoch3_1,
+    #[serde(rename = "latest")]
+    Latest,
 }
 
 impl From<StacksEpochId> for EpochSpec {
@@ -66,6 +68,7 @@ impl From<EpochSpec> for StacksEpochId {
             EpochSpec::Epoch2_5 => StacksEpochId::Epoch25,
             EpochSpec::Epoch3_0 => StacksEpochId::Epoch30,
             EpochSpec::Epoch3_1 => StacksEpochId::Epoch31,
+            EpochSpec::Latest => StacksEpochId::Epoch31,
         }
     }
 }
@@ -100,6 +103,7 @@ impl From<&DevnetConfig> for BurnchainEpochConfig {
                     EpochSpec::Epoch2_5 => config.epoch_2_5,
                     EpochSpec::Epoch3_0 => config.epoch_3_0,
                     EpochSpec::Epoch3_1 => config.epoch_3_1,
+                    EpochSpec::Latest => config.epoch_3_1,
                 };
                 EpochConfig {
                     epoch_name: epoch,
@@ -1556,6 +1560,10 @@ mod tests {
 
                 [[burnchain.epochs]]
                 epoch_name = "3.1"
+                start_height = 9
+
+                [[burnchain.epochs]]
+                epoch_name = "latest"
                 start_height = 9
                 "#
             }
--- a/components/clarinet-files/src/project_manifest.rs
+++ b/components/clarinet-files/src/project_manifest.rs
@@ -5,7 +5,7 @@ use std::str::FromStr;
 use clarity::types::StacksEpochId;
 use clarity::vm::ClarityVersion;
 use clarity_repl::repl;
-use clarity_repl::repl::{ClarityCodeSource, ClarityContract, ContractDeployer};
+use clarity_repl::repl::{ClarityCodeSource, ClarityContract, ContractDeployer, DEFAULT_EPOCH};
 use serde::ser::SerializeMap;
 use serde::{Deserializer, Serialize, Serializer};
 use serde_json::Value as JsonValue;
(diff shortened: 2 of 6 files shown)


components/clarinet-files/src/project_manifest.rs
```diff
@@ -383,12 +383,13 @@ fn get_epoch_and_clarity_version(
     settings_epoch: Option<&str>,
     settings_clarity_version: Option<&str>,
 ) -> Result<(StacksEpochId, ClarityVersion), String> {
-    // if neither epoch or version are specified in clarinet.toml use: epoch 2.05 and clarity 1
+    // if neither epoch or version are specified in clarinet.toml use: epoch DEFAULT_EPOCH and clarity 3
     // if epoch is specified but not version: use the default version for that epoch
 
     let epoch = match settings_epoch {
-        None => StacksEpochId::Epoch2_05,
+        None => DEFAULT_EPOCH,
```
hugoclrd: Needs to be `StacksEpochId::Epoch2_05`, otherwise it would be a breaking change for (really old) projects that didn't have the epoch set by default..
Maybe that would be an acceptable breaking change, but I'm not sure.
Maybe in future versions we'll just make the epoch setting required to have explicit settings and fully avoid this issue.

components/clarinet-files/src/project_manifest.rs
```diff
@@ -399,15 +400,15 @@ fn get_epoch_and_clarity_version(
...
+                "epoch field invalid (value supported: latest, 2.0, 2.05, 2.1, 2.2, 2.3, 2.4, 3.0, 3.1)"
                     .into(),
             ),
         },
     };
 
     let clarity_version = match settings_clarity_version {
         None => match settings_epoch {
-            None => ClarityVersion::Clarity1,
+            None => ClarityVersion::Clarity3,
```
hugoclrd: Same as epoch setting, this has to be Clarity1

components/clarity-repl/src/repl/mod.rs
```diff
@@ -40,62 +41,20 @@ impl Serialize for ClarityContract {
...
+            StacksEpochId::Epoch10 => map.serialize_entry("epoch", &1.0)?,
+            StacksEpochId::Epoch20 => map.serialize_entry("epoch", &2.0)?,
+            StacksEpochId::Epoch2_05 => map.serialize_entry("epoch", &2.05)?,
+            StacksEpochId::Epoch21 => map.serialize_entry("epoch", &2.1)?,
+            StacksEpochId::Epoch22 => map.serialize_entry("epoch", &2.2)?,
+            StacksEpochId::Epoch23 => map.serialize_entry("epoch", &2.3)?,
+            StacksEpochId::Epoch24 => map.serialize_entry("epoch", &2.4)?,
+            StacksEpochId::Epoch25 => map.serialize_entry("epoch", &2.5)?,
+            StacksEpochId::Epoch30 => map.serialize_entry("epoch", &3.0)?,
+            StacksEpochId::Epoch31 => map.serialize_entry("epoch", &"latest")?,
```
hugoclrd: Hmmm, I'm not sure about the implications of this.
it mean that users won't be able to force it to 3.1? Clarinet will alwasy serialize to latest?
What if we need to be explicit sometimes?

hugocaillard (changes requested):

components/clarity-repl/src/repl/mod.rs
tippenein: Ah! I thought I got rid of that. The intention was "explicitly setting 3.1 would stay as 3.1 and latest would be read in as 3.1. I guess I messed up the serialization

components/clarinet-files/src/project_manifest.rs
tippenein: That's reasonable. I should revert this since it's somewhat redundant with the "latest" addition anyway. No reason for a breaking change here

components/clarity-repl/src/repl/mod.rs
tippenein: Added some tests. Should be ok now. Needs more manual testing first though

components/clarinet-deployments/src/types.rs
```diff
@@ -66,6 +68,7 @@ impl From<EpochSpec> for StacksEpochId {
             EpochSpec::Epoch2_5 => StacksEpochId::Epoch25,
             EpochSpec::Epoch3_0 => StacksEpochId::Epoch30,
             EpochSpec::Epoch3_1 => StacksEpochId::Epoch31,
+            EpochSpec::Latest => StacksEpochId::Epoch31,
```
hugoclrd, line 71: Could be `DEFAULT_EPOCH`

components/clarinet-files/src/project_manifest.rs
```diff
@@ -421,6 +421,68 @@ fn get_epoch_and_clarity_version(
...
+                        StacksEpochId::Epoch30 => "3.0".to_string(),
+                        StacksEpochId::Epoch31 => "3.1".to_string(),
+                    }
+                }),
+            );
+            contract_map.insert(
+                "deployer".to_string(),
+                toml::Value::String(match &contract.deployer {
+                    ContractDeployer::LabeledDeployer(s) => s.clone(),
+                    _ => "__default__".to_string(),
```
hugoclrd, line 475: What does this do?

tippenein: Since the DefaultDeployer enum doesn't have a string representation I had to represent it somehow. I think the whole custom Serialize impl for ProjectManifest is bad and undesirable anyway so let me try to remove all this

tippenein:
closed in favor of https://github.com/hirosystems/clarinet/pull/1927

PR #1914, opened 2025-07-14
feat: Return version at LSP initialization

aldur:
Optional per-spec, but nice to have :) https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult


hugocaillard (approved):
lgtm, thx @aldur

PR #1915, opened 2025-07-16
feat: custom boot contracts 

### LINKED ISSUE #1449: Allow to load custom boot contracts
## Context

Clarinet embeds a copy of the boot contracts.
For stacks core developers, it would be useful to have the ability to load custom code instead.

## Solution

Add on options to override boot contracts sources.
In the Clarinet.toml, it should be possible to specify the path to the source file of a boot contract.
This code would be used in place of the one embedded in clarinet. It could also be used to get the code coverage and therefore remove the need for https://github.com/hirosystems/clarinet/pull/1412, and achieve the same result in a more elegant way

```toml
[override_boot_contracts_source]
pox-4 = "./custom-boot-contracts/pox-4.clar"
```

tippenein:
- closes #1449 

adds section in clarinet.toml
```
[project.override_boot_contracts_source]
pox-4 = "./custom-boot-contracts/pox-5.clar"
```

Allows testing custom boot contracts in simnet

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -470,7 +470,7 @@ pub async fn generate_default_deployment(
             }?;
 
             let (epoch, clarity_version) =
-                get_boot_contract_epoch_and_clarity_version(contract_name.as_str())?;
+                get_boot_contract_epoch_and_clarity_version(contract_name.as_str());
 
             // Set the session to the correct epoch for validation
             session.update_epoch(epoch);
--- a/components/clarity-repl/src/repl/boot/mod.rs
+++ b/components/clarity-repl/src/repl/boot/mod.rs
@@ -172,11 +172,7 @@ pub static BOOT_CONTRACTS_DATA: LazyLock<
     );
     for (deployer, boot_code) in deploy.iter() {
         for (name, code) in boot_code.iter() {
-            let (epoch, clarity_version) = get_boot_contract_epoch_and_clarity_version(name)
-                .unwrap_or({
-                    // Fallback to default epoch and clarity version for unknown contracts
-                    (StacksEpochId::Epoch20, ClarityVersion::Clarity1)
-                });
+            let (epoch, clarity_version) = get_boot_contract_epoch_and_clarity_version(name);
             let boot_contract = ClarityContract {
                 code_source: ClarityCodeSource::ContractInMemory(code.to_string()),
                 deployer: ContractDeployer::Address(deployer.to_address()),
@@ -225,14 +221,8 @@ pub fn get_boot_contracts_data_with_overrides(
         };
 
         // Use standard epoch/version mapping for known boot contracts
-        let (epoch, clarity_version) = get_boot_contract_epoch_and_clarity_version(
-            contract_name.as_str(),
-        )
-        .unwrap_or_else(|e| {
-            eprintln!("Warning: Failed to get epoch/clarity version for {contract_name}: {e}");
-            // Fallback to default epoch and clarity version
-            (StacksEpochId::Epoch20, ClarityVersion::Clarity1)
-        });
+        let (epoch, clarity_version) =
+            get_boot_contract_epoch_and_clarity_version(contract_name.as_str());
 
         for deployer in [&*BOOT_TESTNET_PRINCIPAL, &*BOOT_MAINNET_PRINCIPAL] {
             let boot_contract = ClarityContract {
@@ -255,7 +245,7 @@ pub fn get_boot_contracts_data_with_overrides(
 
 pub fn get_boot_contract_epoch_and_clarity_version(
     contract_name: &str,
-) -> Result<(StacksEpochId, ClarityVersion), String> {
+) -> (StacksEpochId, ClarityVersion) {
     let (epoch, clarity_version) = match contract_name {
         "pox-4" | "signers" | "signers-voting" => {
             (StacksEpochId::Epoch25, ClarityVersion::Clarity2)
@@ -267,10 +257,11 @@ pub fn get_boot_contract_epoch_and_clarity_version(
             (StacksEpochId::Epoch20, ClarityVersion::Clarity1)
         }
         _ => {
-            return Err(format!(
-                "Unknown boot contract '{contract_name}' - cannot validate"
-            ));
+            panic!(
+                "Unknown boot contract '{}' - cannot validate",
+                contract_name
+            );
         }
     };
-    Ok((epoch, clarity_version))
+    (epoch, clarity_version)
 }
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -29,6 +29,7 @@ use super::{
     SessionSettings,
 };
 use crate::analysis::coverage::CoverageHook;
+use crate::repl::boot::get_boot_contract_epoch_and_clarity_version;
 use crate::repl::clarity_values::value_to_string;
 use crate::utils::serialize_event;
 
@@ -149,17 +150,7 @@ impl Session {
                 code.to_string()
             };
 
-            let (epoch, clarity_version) = if (*name).eq("pox-4") {
-                (StacksEpochId::Epoch25, ClarityVersion::Clarity2)
-            } else if (*name).eq("pox-3") {
-                (StacksEpochId::Epoch24, ClarityVersion::Clarity2)
-            } else if (*name).eq("pox-2") || (*name).eq("costs-3") {
-                (StacksEpochId::Epoch21, ClarityVersion::Clarity2)
-            } else if (*name).eq("cost-2") {
-                (StacksEpochId::Epoch2_05, ClarityVersion::Clarity1)
-            } else {
-                (StacksEpochId::Epoch20, ClarityVersion::Clarity1)
-            };
+            let (epoch, clarity_version) = get_boot_contract_epoch_and_clarity_version(name);
 
             let contract = ClarityContract {
                 code_source: ClarityCodeSource::ContractInMemory(contract_source),


hugoclrd:
The example in PR description says
```toml
[project.override_boot_contracts_source]
pox-4-alt = "./custom-boot-contracts/pox-4-alt.clar"
```

But `pox-4-alt` wouldn't be valid right? It need to be an actual boot contract like `pox-4`?

tippenein:
> But pox-4-alt wouldn't be valid right? It need to be an actual boot contract like pox-4?

Initially I had it as override only, but now it works with either. I assume someone might want to test pox-5 along with the original pox-4, in which case we'd need to allow both addition and overriding

components/clarinet-cli/examples/custom-boot-contracts/README.md
```diff
@@ -0,0 +1,76 @@
+# Custom Boot Contracts Example
+
+This example demonstrates how to use override boot contracts in Clarinet with custom implementations.
```
hugoclrd: Specify that it only applies to simnet (`clarinet console`, unit tests with the sdk)

components/clarinet-cli/examples/custom-boot-contracts/settings/Simnet.toml
```diff
@@ -0,0 +1,134 @@
...
+disable_stacks_explorer = true
+disable_stacks_api = true
+disable_postgres = true
+# bind_containers_volumes = true
+# stacks_node_env_vars = ["STACKS_LOG_DEBUG=1"]
+enable_subnet_node = false
+# subnet_contract_id = "STXMJXCJDCT4WPF2X1HE42T6ZCCK3TPMBRZ51JEG.subnet"
+# subnet_leader_mnemonic = "female adjust gallery certain visit token during great side clown fitness like hurt clip knife warm bench start reunion globe detail dream depend fortune"
+# disable_subnet_api = true
+enable_next_features = true
```
hugoclrd: i think this settings don't exist anymore. It's a bit weird to some them here.

I think it would be fine to remove all of the commented settigns and only keep the relevants ones

components/clarinet-deployments/src/lib.rs
```diff
@@ -90,10 +91,24 @@ pub fn setup_session_with_deployment(
 }
 
 pub fn initiate_session_from_manifest(manifest: &ProjectManifest) -> Session {
+    let mut repl_settings = manifest.repl_settings.clone();
+    repl_settings.remote_data.enabled = false;
```
hugoclrd: Not sure this is what we want here 🤔

components/clarinet-cli/examples/custom-boot-contracts/settings/Simnet.toml
hugoclrd: I see now that it was copied from the counter example. I think both could be cleaned

components/clarinet-deployments/src/lib.rs
```diff
@@ -179,11 +194,32 @@ pub fn update_session_with_deployment_plan(
...
+        };
 
         for (contract_id, (contract, ast)) in boot_contracts_data {
-            let result = session.interpreter.run(&contract, Some(&ast), false, None);
-            boot_contracts.insert(contract_id, result);
+            // Only load boot contracts that are in the include_boot_contracts list
+            if session
+                .settings
+                .include_boot_contracts
+                .contains(&contract.name)
```
hugoclrd: I think that `include_boot_contracts` is an old settings that was partially removed from the logic, but remains in some toml and struc (maybe for backaward compat).

It turned out that letting end-users control which boot contracts too load was a bad idea because it could easily lead to errors and invalid deployments. It's best to assume that all boot contract should be deployed and required.

We should remove it fully and cleanly.

hugocaillard (commented):
i left some comments, but I see this PR is actually in progress / draft.

Still think the `include_boot_contracts` comment at this stage (the others, maybe not so much)

components/clarinet-cli/examples/custom-boot-contracts/settings/Simnet.toml
tippenein: Will do 👍

components/clarinet-deployments/src/lib.rs
tippenein: Removed 👌

components/clarinet-cli/examples/custom-boot-contracts/settings/Simnet.toml
```diff
@@ -0,0 +1,80 @@
...
+# stx_address: STNHKEPYEPJ8ET55ZZ0M5A34J0R3N5FM2CMMMAZ6
+# btc_address: mjSrB3wS4xab3kYqFktwBzfTdPg367ZJ2d
+
+[simnet]
+bitcoin_controller_block_time = 20_000
+disable_bitcoin_explorer = true
+disable_stacks_explorer = true
+disable_stacks_api = true
+disable_postgres = true
+enable_subnet_node = false
```
hugoclrd: this setting doesn't exist anymore

components/clarinet-deployments/src/lib.rs
```diff
@@ -179,9 +181,24 @@ pub fn update_session_with_deployment_plan(
...
+            let is_simnet = deployment.genesis.is_some();
+            if is_simnet {
+                clarity_repl::repl::boot::get_boot_contracts_data_with_overrides(
+                    &session.settings.override_boot_contracts_source,
+                )
+            } else {
+                // For non-simnet deployments, ignore custom boot contracts
+                BOOT_CONTRACTS_DATA.clone()
+            }
+        };
```
hugoclrd: In `update_session_with_deployment_plan`, `session` is a _simnet session_. I think we are always in simnet in this context

components/clarinet-deployments/src/lib.rs
```diff
@@ -90,10 +91,11 @@ pub fn setup_session_with_deployment(
...
 
 pub fn initiate_session_from_manifest(manifest: &ProjectManifest) -> Session {
+    // For session initialization, we assume simnet context (used for console, tests, etc.)
+    // Custom boot contracts are allowed in this context
     let settings = SessionSettings {
         repl_settings: manifest.repl_settings.clone(),
-        disk_cache_enabled: true,
-        cache_location: Some(manifest.project.cache_location.to_path_buf()),
+        override_boot_contracts_source: manifest.project.override_boot_contracts_source.clone(),
         ..Default::default()
```
hugoclrd, line 98: Why was this removed?

components/clarity-repl/src/repl/settings.rs
```diff
@@ -45,14 +46,15 @@ pub struct Account {
...
-    pub include_boot_contracts: Vec<String>,
     pub include_costs: bool,
     pub initial_accounts: Vec<Account>,
     pub initial_deployer: Option<Account>,
     pub disk_cache_enabled: bool,
     pub repl_settings: Settings,
     pub cache_location: Option<PathBuf>,
     pub epoch_id: Option<StacksEpochId>,
+    pub override_boot_contracts_source: BTreeMap<String, String>,
+    pub simnet_override_boot_contracts_source: BTreeMap<String, String>,
```
hugoclrd: Why do we need 2 settings?

components/clarinet-deployments/src/lib.rs
```diff
@@ -179,9 +181,24 @@ pub fn update_session_with_deployment_plan(
...
+                    &session.settings.override_boot_contracts_source,
+                )
+            } else {
+                // For non-simnet deployments, ignore custom boot contracts
+                BOOT_CONTRACTS_DATA.clone()
+            }
+        };
 
         for (contract_id, (contract, ast)) in boot_contracts_data {
+            // Always load all boot contracts
```
hugoclrd: This comment can be removed imo

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+                return Err(format!(
+                    "Custom boot contract '{}' is invalid:\n{}",
+                    contract_name,
+                    error_messages.join("\n")
+                ));
+            }
+        }
+    }
+
+    // Only allow overriding existing boot contracts, not adding new ones
```
hugoclrd: I think this comment doesn't match line below

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -1112,11 +1112,7 @@ impl SDK {
     // this method empty the session costs and coverage reports
     // and returns this report
     #[wasm_bindgen(js_name=collectReport)]
-    pub fn collect_report(
-        &mut self,
-        include_boot_contracts: bool,
```
hugoclrd: This `include_boot_contracts` isn't thesame as the old `SessionSetting.include_boot_contracts`

The param is used to tell wether we want to include boot conctracts or not in the code coverage report.
We want to keep this logic

components/clarinet-sdk/node/tests/remote-data.test.ts
```diff
@@ -52,59 +52,83 @@ describe("simnet remote interactions", async () => {
...
-  });
+    try {
+      const { result } = simnet.callReadOnlyFn(counterAddress, "get-count", [], sender);
+      expect(result).toStrictEqual(Cl.uint(0));
+    } catch (error) {
+      // Skip test if remote contract is not available
+      console.warn("Remote contract not available, skipping test:", error);
+      return;
+    }
+  }, 10000); // Increase timeout to 10 seconds
```
hugoclrd: Is this to just ignore the flaky tests?
Not sure we want it, it'll just silence the tests. I wouldn't include this change in this PR

components/clarinet-cli/examples/custom-boot-contracts/deployments/default.simnet-plan.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - pox
+    - pox-2
+    - pox-3
+    - pox-4
+    - lockup
+    - costs-2
+    - costs-3
+    - cost-voting
+    - bns
+    - pox-5
```
hugoclrd: I think this file was generated when there was the feature enable additional boot contracts

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+            // Check if this is a valid boot contract name
+            if !clarity_repl::repl::boot::BOOT_CONTRACTS_NAMES.contains(&contract_name.as_str()) {
+                eprintln!("Warning: Skipping custom boot contract '{contract_name}' - only existing boot contracts can be overridden. Valid boot contracts are: {:?}", clarity_repl::repl::boot::BOOT_CONTRACTS_NAMES);
+                continue;
+            }
+
+            // Load the additional boot contract source
+            let mut contract_location = base_location.clone();
+            contract_location
+                .append_path(&format!("custom-boot-contracts/{contract_name}.clar"))?;
```
hugoclrd: Since the boot contract path is specified in the manifest, it shouldn't be hard coded here

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+                }
+                _ => {
+                    return Err(format!(
+                        "Unknown boot contract '{contract_name}' - cannot validate"
+                    ));
+                }
+            };
+
+            // Set the session to the correct epoch for validation
+            session.update_epoch(epoch);
```
hugoclrd, line 476: I don't think this is needed.
cf comment L497 (no need to deploy contract)

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+    if !settings.override_boot_contracts_source.is_empty() && !simnet_remote_data {
+        let mut session = Session::new(settings.clone());
+        for (contract_name, file_path) in &settings.override_boot_contracts_source {
+            // Only validate existing boot contracts that are being overridden
+            if !clarity_repl::repl::boot::BOOT_CONTRACTS_NAMES.contains(&contract_name.as_str()) {
+                continue;
+            }
+
+            // Load and validate the custom boot contract
+            let custom_source = match clarity_repl::repl::boot::load_custom_boot_contract(file_path)
```
hugoclrd: nit: use map_err instead

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+            // Create a temporary contract for validation
+            let temp_contract = ClarityContract {
+                code_source: ClarityCodeSource::ContractInMemory(custom_source),
+                deployer: ContractDeployer::Address(default_deployer_address.to_address()),
+                name: contract_name.clone(),
+                clarity_version,
+                epoch: clarity_repl::repl::Epoch::Specific(epoch),
+            };
+
+            let (_, diagnostics, ast_success) = session.interpreter.build_ast(&temp_contract);
```
hugoclrd, line 485: The diagnostics should be added to `contract_diags`, that is currently declared many lines below, but can be moved up.
Same for the `ast_success` logic

Instead of throwing an error if `!ast_success`

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+                Err(deploy_errors) => {
+                    let error_messages: Vec<String> =
+                        deploy_errors.iter().map(|d| d.message.clone()).collect();
+                    return Err(format!(
+                        "Custom boot contract '{}' is invalid:\n{}",
+                        contract_name,
+                        error_messages.join("\n")
+                    ));
+                }
+            }
```
hugoclrd: `generate_default_deployment` is already a huge function.

It currently doesn't vlaide that deployment of contract, so I don't think we want to add this logic here.

Instead, in `fn update_session_with_deployment_plan()`, we already deploy the boot cotnract `session.interpreter.run(&contract, Some(&ast), false, None);` and get the diagnostics from here.

The boot contract diagnostics are currently ignored (because we consider the boot contract to always be valid).
But with this PR, we could get the custom boot contract diagnostic and treat them as normal contracts diagnostics.
For example [here](https://github.com/hirosystems/clarinet/blob/f90c4b8b61b20f1d3c06fd51074078051e102997/components/clarinet-deployments/src/lib.rs#L56-L57)
And eventually, the diagnostics will show up in the LSP if handled [here](https://github.com/hirosystems/clarinet/blob/f90c4b8b61b20f1d3c06fd51074078051e102997/components/clarity-lsp/src/common/state.rs#L650-L655)

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+            // Validate the custom boot contract
+            if !ast_success {
+                let error_messages: Vec<String> =
+                    diagnostics.iter().map(|d| d.message.clone()).collect();
+                return Err(format!(
+                    "Custom boot contract '{}' is invalid:\n{}",
+                    contract_name,
+                    error_messages.join("\n")
+                ));
+            }
```
hugoclrd: To be removed is `ast_success` and `diagnostics` are handled the same way they are handled for normal contracts
https://github.com/hirosystems/clarinet/blob/f90c4b8b61b20f1d3c06fd51074078051e102997/components/clarinet-deployments/src/lib.rs#L996-L998

components/clarinet-deployments/src/lib.rs
hugoclrd: here, in `load_custom_boot_contract`, it uses `fs::read_to_string`, but a few lines below, the sources are read again using the file_accessor, if it exists.

I think the `file_accessor` logic is right if we want it to work in wasm (especially in the vscode extension).
But also, I'm not sure why the source is read twice

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+            };
+
+            // Create contract ID for the additional boot contract
+            let contract_id = QualifiedContractIdentifier::new(
+                default_deployer_address.clone(),
+                ContractName::try_from(contract_name.clone())
+                    .map_err(|_| format!("Invalid contract name: {contract_name}"))?,
+            );
+
+            // Add to requirements data for AST building
```
hugoclrd: Why are `requirements` mentioned here?

components/clarinet-deployments/src/lib.rs
```diff
@@ -406,6 +441,195 @@ pub async fn generate_default_deployment(
...
+                code_source: ClarityCodeSource::ContractInMemory(source.clone()),
+                deployer: ContractDeployer::Address(default_deployer_address.to_address()),
+                name: contract_name.clone(),
+                clarity_version: DEFAULT_CLARITY_VERSION,
+                epoch: clarity_repl::repl::Epoch::Specific(DEFAULT_EPOCH),
+            };
+
+            let (ast, diagnostics, ast_success) = session.interpreter.build_ast(&contract);
+
+            // Try to deploy the contract to catch any runtime errors that build_ast might miss
```
hugoclrd: I'm lost here, why are we re-validating the boot contracts? It was already done above

components/clarinet-deployments/src/lib.rs
```diff
@@ -872,9 +1096,19 @@ pub async fn generate_default_deployment(
...
+            // Include both standard boot contracts and custom boot contracts from override_boot_contracts_source
+            let mut genesis_contracts = manifest.project.boot_contracts.clone();
+
+            // Add contracts from override_boot_contracts_source to genesis contracts
+            for contract_name in manifest.project.override_boot_contracts_source.keys() {
+                if !genesis_contracts.contains(contract_name) {
+                    genesis_contracts.push(contract_name.clone());
+                }
+            }
+
```
hugoclrd: Won't it always produce the same lisst of boot contracts name?

components/clarinet-files/src/project_manifest.rs
```diff
@@ -33,6 +33,9 @@ pub struct ProjectConfigFile {
     telemetry: Option<bool>,
     requirements: Option<TomlValue>,
     boot_contracts: Option<Vec<String>>,
+    override_boot_contracts_source: Option<BTreeMap<String, String>>,
+    #[serde(rename = "simnet_override_boot_contracts_source")]
+    simnet_override_boot_contracts_source: Option<BTreeMap<String, String>>,
```
hugoclrd: Why two settings here?

components/clarinet-files/src/project_manifest.rs
```diff
@@ -257,6 +264,52 @@ impl ProjectManifest {
...
+            .project
+            .simnet_override_boot_contracts_source
+        {
+            for (contract_name, contract_path) in overrides.iter() {
+                simnet_override_boot_contracts_source
+                    .insert(contract_name.clone(), contract_path.clone());
+            }
+        }
+
+        // Always include all boot contracts - the include_boot_contracts setting is no longer used
```
hugoclrd: I would remove the mention of `include_boot_contracts`, i'm afradi that might add confusion in the future

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -1130,14 +1126,13 @@ impl SDK {
             contract_paths.insert(contract_id.name.to_string(), contract_location.to_string());
         }
 
-        if include_boot_contracts {
```
hugoclrd: Following the comment above, we need this

components/clarity-repl/src/repl/session.rs
```diff
@@ -139,35 +139,95 @@ impl Session {
...
+            };
+
+            let contract = ClarityContract {
+                code_source: ClarityCodeSource::ContractInMemory(contract_source),
+                name: name.to_string(),
+                deployer: deployer.clone(),
+                clarity_version,
+                epoch: Epoch::Specific(epoch),
+            };
+            let _ = self.deploy_contract(&contract, false, None);
```
hugoclrd: I think this can be removed?
Custom boot contracts are already deployed above

components/clarity-repl/src/repl/session.rs
```diff
@@ -430,7 +490,7 @@ impl Session {
         let output = Vec::<String>::new();
         let contracts = vec![]; // This is never modified, remove?
 
-        if !self.settings.include_boot_contracts.is_empty() {
+        if !self.settings.override_boot_contracts_source.is_empty() {
```
hugoclrd: I think we want to fully remove this condition here

components/clarinet-sdk/node/tests/remote-data.test.ts
tippenein: Yea, This ended up not being the issue anyway. Removing 👍

components/clarinet-deployments/src/lib.rs
tippenein: The validation fails if we don't have this. Possibly a different issue, but removing this we get:

```
'pox-4' has validation errors:
  - contract epoch (2.5) does not match current epoch (2.05)
  ```

File: components/clarinet-sdk-wasm/src/core.rs [hunk 28]
```diff
@@ -11,9 +11,9 @@ use clarinet_deployments::{
     generate_default_deployment, initiate_session_from_manifest,
     update_session_with_deployment_plan,
 };
-use clarinet_files::{
-    FileAccessor, FileLocation, ProjectManifest, StacksNetwork, WASMFileSystemAccessor,
-};
+#[cfg(target_arch = "wasm32")]
```
hugoclrd: `clarinet-sdk-wasm` is always and only compiled for wasm32, so that's not needed

components/clarinet-cli/examples/custom-boot-contracts/settings/Simnet.toml
```diff
@@ -0,0 +1,79 @@
+[network]
+name = "simnet"
```
hugoclrd: I don't think `Simnet.toml` is ever used.
Should be `Devnet.toml` (the `name` doesn't really matter here I believe)

components/clarinet-deployments/src/lib.rs
```diff
@@ -408,6 +455,249 @@ pub async fn generate_default_deployment(
...
+                            default_deployer_address.clone(),
+                            ContractName::try_from(contract_name.clone())
+                                .unwrap_or_else(|_| ContractName::from("unknown")),
+                        ),
+                        deploy_errors,
+                    );
+                    asts_success = false;
+                    continue;
+                }
+            }
```
hugoclrd: I think the custom boot contracts shouldn't be evaluated/deployed here in `generate_default_deployment_plan`. it happens later in the process in `update_session_with_deployment_plan`.

`generate_default_deployment_plan`does the static analysis and shouldn't require a session.
It seems redundant.

components/clarinet-deployments/src/lib.rs
```diff
@@ -408,6 +455,249 @@ pub async fn generate_default_deployment(
...
+            let mut session = Session::new(settings.clone());
+            let contract = ClarityContract {
+                code_source: ClarityCodeSource::ContractInMemory(source.clone()),
+                deployer: ContractDeployer::Address(default_deployer_address.to_address()),
+                name: contract_name.clone(),
+                clarity_version: DEFAULT_CLARITY_VERSION,
+                epoch: clarity_repl::repl::Epoch::Specific(DEFAULT_EPOCH),
+            };
+
+            let (ast, diagnostics, ast_success) = session.interpreter.build_ast(&contract);
```
hugoclrd: it seems to me that all of this is very similar with the lines above. It's kind of hard to understand I don't fully follow the logic of all this.
Maybe it could be split in smaller functions, or factorized to avoid repetition?

tippenein: I was able to remove the epoch setting.  Since we don't do the unnecessary deploys and such I think it became redundant and stopped causing the validation error it was initially meant to fix 👍

components/clarinet-deployments/src/lib.rs
tippenein: removed 👌

components/clarinet-files/src/project_manifest.rs
```diff
@@ -143,6 +144,8 @@ pub struct ProjectConfig {
     pub cache_location: FileLocation,
     #[serde(skip_deserializing)]
     pub boot_contracts: Vec<String>,
+    #[serde(skip_deserializing)]
```
hugoclrd: This can be removed. We do want to deserialize `override_boot_contracts_source`

components/clarinet-files/src/project_manifest.rs
```diff
@@ -257,6 +305,38 @@ impl ProjectManifest {
...
+            "pox".to_string(),
+            "pox-2".to_string(),
+            "pox-3".to_string(),
+            "pox-4".to_string(),
+            "lockup".to_string(),
+            "costs-2".to_string(),
+            "costs-3".to_string(),
+            "cost-voting".to_string(),
+            "bns".to_string(),
+        ];
```
hugoclrd:
```suggestion
        let boot_contracts = BOOT_CONTRACTS_NAMES
            .iter()
            .map(|n| n.to_string())
            .collect::<Vec<_>>();
```

With `use clarity_repl::repl::boot::BOOT_CONTRACTS_NAMES;`

components/clarinet-deployments/src/lib.rs
```diff
@@ -296,12 +306,30 @@ fn handle_emulated_contract_call(
     result
 }
 
+// Main function that always includes boot contracts by default
```
hugoclrd: Remove this comment.
I think the comment only make sense in the context of this PR. But if someone discover the repo and sees this comment, that won't be helpful at all.

components/clarinet-deployments/src/lib.rs
```diff
@@ -387,8 +415,20 @@ pub async fn generate_default_deployment(
 
     let mut repl_settings = manifest.repl_settings.clone();
     repl_settings.remote_data.enabled = false;
+
+    // Custom boot contracts are only supported on simnet
```
hugoclrd: Remove, I think the comment only describes the `if` below without giving additional information

components/clarity-repl/src/repl/boot/mod.rs
```diff
@@ -200,3 +193,88 @@ pub static BOOT_CONTRACTS_DATA: LazyLock<
     }
     result
 });
+
+/// Load custom boot contract source from file
```
hugoclrd: Remvoe

components/clarity-repl/src/repl/boot/mod.rs
```diff
@@ -200,3 +193,88 @@ pub static BOOT_CONTRACTS_DATA: LazyLock<
...
+    let mut result = BOOT_CONTRACTS_DATA.clone();
+
+    let interpreter = ClarityInterpreter::new(
+        StandardPrincipalData::transient(),
+        Settings::default(),
+        None,
+    );
+
+    for (contract_name, file_path) in overrides {
+        // Only allow overriding existing boot contracts
```
hugoclrd: remove

components/clarity-repl/src/repl/boot/mod.rs
```diff
@@ -200,3 +193,88 @@ pub static BOOT_CONTRACTS_DATA: LazyLock<
...
+        if !BOOT_CONTRACTS_NAMES.contains(&contract_name.as_str()) {
+            eprintln!("Warning: Skipping custom boot contract '{contract_name}' - only existing boot contracts can be overridden. Valid boot contracts are: {BOOT_CONTRACTS_NAMES:?}");
+            continue;
+        }
+
+        // Load custom source
```
hugoclrd: remove

components/clarity-repl/tests/override_boot_contracts.rs
```diff
@@ -0,0 +1,56 @@
...
+use clarinet_deployments::initiate_session_from_manifest;
+use clarinet_files::{FileLocation, ProjectManifest};
+use tempfile::TempDir;
+
+#[test]
+fn test_override_boot_contracts() {
+    let temp_dir = TempDir::new().expect("Failed to create temp dir");
+    let project_path = temp_dir.path();
+
+    // Write Clarinet.toml with override_boot_contracts_source for pox-4
```
hugoclrd: Remove this comment, and I think all comments below

components/clarinet-deployments/src/lib.rs
```diff
@@ -90,10 +90,13 @@ pub fn setup_session_with_deployment(
 }
 
 pub fn initiate_session_from_manifest(manifest: &ProjectManifest) -> Session {
+    // For session initialization, we assume simnet context (used for console, tests, etc.)
```
hugoclrd: session ~= simnet

So yes, i think the coment can be removed

components/clarinet-deployments/src/lib.rs
```diff
@@ -296,12 +306,30 @@ fn handle_emulated_contract_call(
...
+        manifest,
+        network,
+        no_batch,
+        file_accessor,
+        forced_min_epoch,
+    )
+    .await
+}
+
+pub async fn generate_default_deployment_with_boot_contracts(
```
hugoclrd: Do we need this extra function?
I think it can be removed and just keep `generate_default_deployment`

components/clarinet-deployments/src/lib.rs
```diff
@@ -408,6 +454,69 @@ pub async fn generate_default_deployment(
...
+                code_source: ClarityCodeSource::ContractInMemory(custom_source),
+                deployer: ContractDeployer::Address(default_deployer_address.to_address()),
+                name: contract_name.clone(),
+                clarity_version,
+                epoch: clarity_repl::repl::Epoch::Specific(epoch),
+            };
+
+            let (_, diagnostics, ast_success) = session.interpreter.build_ast(&temp_contract);
+
+            // Collect AST diagnostics instead of returning immediately
```
hugoclrd: Remove. Again, only make sense in the context of this PR and is reference to a comment I made

hugoclrd (changes requested):
Almost looking good.

I left a few comments on the many  `// comments` in this PR. I think many are useless, if not misleading in some cases.

I don't comment on all of them, but please review all of the `// comments` in this PR and make sure they are relevant (remove them instead)

components/clarity-repl/src/repl/boot/mod.rs
```diff
@@ -171,19 +172,11 @@ pub static BOOT_CONTRACTS_DATA: LazyLock<
...
-            let (epoch, clarity_version) = match *name {
-                "pox-4" | "signers" | "signers-voting" => {
-                    (StacksEpochId::Epoch25, ClarityVersion::Clarity2)
-                }
-                "pox-3" => (StacksEpochId::Epoch24, ClarityVersion::Clarity2),
-                "pox-2" | "costs-3" => (StacksEpochId::Epoch21, ClarityVersion::Clarity2),
-                "costs-2" => (StacksEpochId::Epoch2_05, ClarityVersion::Clarity1),
-                "genesis" | "lockup" | "bns" | "cost-voting" | "costs" | "pox" => {
+            let (epoch, clarity_version) = get_boot_contract_epoch_and_clarity_version(name)
+                .unwrap_or({
```
hugoclrd: keep the panic
We shouldn't have unknown boot contracts

components/clarity-repl/src/repl/session.rs
```diff
@@ -131,35 +131,46 @@ impl Session {
...
+            };
 
-                let contract = ClarityContract {
-                    code_source: ClarityCodeSource::ContractInMemory(code.to_string()),
-                    name: name.to_string(),
-                    deployer: deployer.clone(),
-                    clarity_version,
-                    epoch: Epoch::Specific(epoch),
-                };
+            let (epoch, clarity_version) = if (*name).eq("pox-4") {
```
hugoclrd: can we use the new `get_boot_contract_epoch_and_clarity_version` here?

hugoclrd (approved):

PR #1929, opened 2025-07-24
remove stacks-codec duplication using stackslib

### LINKED ISSUE #1531: Import stacks-codec from stackslib
Clarinet exports its own copy of the [stacks-codec](https://github.com/hirosystems/clarinet/blob/79ce8a417090ddabc6f3f72716c9a0c82de68de5/components/stacks-codec/src/codec.rs).

It's mostly code copy-pasted for [stacks-core/stackslib](https://github.com/stacks-network/stacks-core/tree/47db1d0a8bf70eda1c93cb3e0731bdf5595f7baa/stackslib). I think it was copy pasted because the codec in stacks lib can not be imported out of the box.

But now it makes it complicated and error prone to maintain and update the codec in clarinet (eg: this PR https://github.com/hirosystems/clarinet/pull/1524). We should consider imported the code from stacks-core

tippenein:
- closes #1531


hugo-stacks:
I think I'd like to re-explore this in the future, Ideally make it work in way where we don't need to have this codec crates in Clarinet, but just use stuff from stacks-core.

What do you think @brady-stacks?

PR #1933, opened 2025-07-25
chore: update lsp dependencies

hugocaillard:
### Description

Part of a continous effort to improve our dependencies maintenance and reduce the size of Clarinet.
This espeically removes tower@0.4, while tower@0.5 was already part of the dependency tree

Update:
- lsp-types
- tower-lsp -> ugprade to the community maintained [tower-lsp-server](https://github.com/tower-lsp-community/tower-lsp-server)

Breaking change:
- lsp-types 0.97 now relies on fluent_uri::Uri insteal of Url
  - this change, again, surfaces the need to fully refactor our FileLocation struct


jbencin:
It looks like you are reverting my changes from #1859. These were necessary to get the Native LSP (which uses the `FileLocation::FileSystem` type for files) working. Unless something has changed in tower-lsp's URL parsing, and it can now handle URL's without the `file://` prefix, this will break the Native LSP

components/clarinet-files/src/lib.rs
```diff
@@ -367,16 +349,15 @@ impl FileLocation {
...
-                    .map_err(|_| format!("unable to conver path {} to url", path.display()))?;
-                Ok(url.to_string())
+            #[cfg(target_arch = "wasm32")]
+            FileLocation::FileSystem { .. } => {
+                Err("Url::from_file_path is not supported in wasm32".to_string())
             }
+            #[cfg(not(target_arch = "wasm32"))]
+            FileLocation::FileSystem { path } => Url::from_file_path(path)
+                .map_err(|_| format!("unable to convert path {} to url", path.display()))
+                .map(|url| url.to_string()),
```
jbencin:
```suggestion
                .map(String::from),
```

components/clarity-lsp/src/common/backend.rs
```diff
@@ -652,4 +653,49 @@ mod range_formatting_tests {
...
+        let request = LspRequest::Definition(params);
+        let response =
+            process_request(request, &editor_state_input).expect("Failed to process request");
+
+        if let LspRequestResponse::Definition(Some(location)) = &response {
+            assert_eq!(location.uri.scheme().unwrap().as_str(), "file");
+            assert_eq!(location.uri.path().as_str(), "/test.clar");
+        } else {
+            panic!("Expected a Definition response, got: {response:?}");
+        }
```
jbencin: Minor, but `let`/`else` is a bit cleaner:

```suggestion
        let LspRequestResponse::Definition(Some(location)) = &response else {
            panic!("Expected a Definition response, got: {response:?}");
        };

        assert_eq!(location.uri.scheme().unwrap().as_str(), "file");
        assert_eq!(location.uri.path().as_str(), "/test.clar");
```

components/clarity-lsp/src/vscode_bridge.rs
```diff
@@ -177,7 +177,7 @@ impl LspVscodeBridge {
             }
 
             for (location, diags) in aggregated_diagnostics.into_iter() {
-                if let Ok(uri) = Url::parse(&location.to_string()) {
+                if let Ok(uri) = location.to_string().parse() {
```
jbencin: Even though I think the VSCode bridge can only use `FileLocation::Url`, it would be more correct here to use `to_url_string()`, so it would work if somehow this code was used with `FileLocation::FileSystem`

PR #1940, opened 2025-07-29
feat: perf hook data

obycode:
Use `::perf <cost field> <expression>` to generate a _perf.data_ file which can be used to generate a flamegraph.

That perf.data can be, for example, fed into https://github.com/brendangregg/FlameGraph/blob/master/flamegraph.pl to generate a flame graph.

Details:
- any element of costs can be measured `::perf read_length <expression>`
- SDK integration for tests allows `simnet.enablePerformance("runtime")`

### Basic SDK usage
```
await simnet.enablePerformance("read_length");

const { result, costs, performance } = simnet.callReadOnlyFn("contract", "method1", args, sender);
```


obycode:
Sample usage:
- Setup Clarinet project
- Enable remote data at block 2428362
  ```toml
  [repl.remote_data]
  enabled = true
  api_url = 'https://api.hiro.so'
  initial_height = 2428362
  ```
- Start the clarinet console
- Set the tx-sender:
  ```
  ::set_tx_sender SP14JPKZT1AGJWXHTVBPMVZQ11T7PGT0205DACDTB
  ```
- Run:
  ```
  ::perf runtime (contract-call? 'SP3BJR4P3W2Y9G22HA595Z59VHBC9EQYRFWSKG743.borrower-v1 borrow (some 0x504e41550100000003b801000000040d00a528f4e5f2313a64c5460053dd314e4d3d0c713c7f51f5c82df581d9979f03063e24b899267f1ff1294745e7ef69e7cc444c34bc881698f861b5c04efdcd354f01020064e9e8d47253868358d30a7c301e81212bd356f604daa87d742943c44c9c641dae23b34bbf97599118ac5518eb01a4a1825372c0f994d9a0f813860a6470f20003571ba83d39d2a6b1babf6b1503178462af5a272539198ef187d5526e300acb1828d294349966e449014780f3244bab5b15579326cd83f44c23827394a5d970f60104637df7bd3d10befda3faf193ea31b7f53827e18ec609e77400440056f0d46fb036dd7a15d4836ce2467bcd77fd215fe9ab50885a1e39ad94be7e45c2d48baf10000685f23589ff039a84890c25821895b2fd4e8df9b1d4eb981a548d1f8ca34cd3b16158e3ac10d6517ab3f8b82f0fe1e3003e61f9d70015cf4b150365b4c3d56fe800085a93cb2fadf5317b94cc8e0f4b7b86f778d3a4f7f3e67455c2732bf01d12766f1531342de301209e8663a30443d5de58e82eab274b9ea637799d7d1a0d96a1b5010acfd59e9fd85f142598c059e48db3c75f6f61ec582a3409d499fc0fa5c9ccdbab18955b9b2dd4bf72531e8035fe9319c72515ce3d08f67cf2d93fb80b775e74e1000b5ceae217c7e500d81fba396e1b6e04b79a08271616d3105d480a141c6a9ab1c127a42b0273e0f06382b1092fd35b10e25e9d35bc509b9bcb387b6c2796707ef0010c8e7895b3c0a64eec0f017e7198375d057031a0ad5e03d0897c103bbf397f64d066610b864dcfece5e7ec8890a7e5121b4b8e3d302176c9ba270cf1243c9bceba000dab1ada825e8bdddb0eb1b3a18d5e0eb34a90ff28f8d5cda9be6048cb2259e0ef4cea28172a9e46255720d93a941c0f28ee9b7df13c842d80ad487c7357e1d1f7000e86d86cf8dbaeeed2cb6700a3d61b6cce92f97da8229671e4d56809219633813134cd9230d8939bd201641ed9278973ca283684acc8731905a0a83573d56dbd0d0010df247fdd1c470bfc0a89f17b57b7bb04667139f712ee6e8ba599bc5b8a51f63a5f8e7cf37ac2de80389299b6651e2f611fa5eef1103e91338edb9dca797e162c0111ef8ad8b5187eb4b234766599fb42f2be5dbf40e7f7a64b1362b7758a7e103fe9456acf2f7112508c3f2e897aa56576f838d644252626531f143a679674f041bc0068827d5d00000000001ae101faedac5851e32b9b23b5f9411a8c2bac4aae3ed4dd7b811dd1a72ea4aa710000000008c45d4f014155575600000000000dd1ef2c00002710da4aa9e997ab3ab76d06970986b98acd5ae5b09802005500e62df6c8b4a85fe1a67db44dc12de5db330f7ac66b72dc658afedf0f4a415b4300000ad680104179000000012796f547fffffff80000000068827d5d0000000068827d5c00000ad5cbdbee2000000000fd74dbd40cdf14384e6f747ed718b18cbaa9fd02f0659ca0d0c3be15078b496765e5b07293c27058d0bb6b36513550c92b124f26ea3667937b4f1f685dac997fd43dcbff915277407cee1b868ea936be83490f501f5717248152a484b0c3a2613dde2de4fc7fa7d7fafae58b7b4a91fc6e3e053bf15e83ad8d715990eb8dcc011406028514e8e4b0253175a9a1fa6fe6fe048bcf270829e3370143210f5392f6f8853a8427442235cebce92b4725891c0969ff6038b113b89176c4eab6e3876d48aa3a5cbb078d003f95f7de6817e266477a61f2247552552544571dd898963816d878cee1b41633897c21d5b2daf17450107d4c71005500eaa020c61cc479712813461ce153894a96a6c00b21ed0cfc2798d1f9a9e9c94a0000000005f5c401000000000001c7b0fffffff80000000068827d5d0000000068827d5c0000000005f5bad90000000000019e530cd711886422b644b2a50d8d6e006e732b170f6d2a1019a7e963c398326d33c99fe36921a31daba1f7f3ecc18169107de6b123f954510f898750a6363371f4851e9bff4d70d4483ddd71418b783fb39f1ab6d148e4fec3cb6bdef4ebc5eb3f492ca135cca70d977afd85b6c1992f445b5b1b7413b9e1c5d9aff16c46060dad05714c7b526f54cdf2faec6a8e3f048bcf270829e3370143210f5392f6f8853a8427442235cebce92b4725891c0969ff6038b113b89176c4eab6e3876d48aa3a5cbb078d003f95f7de6817e266477a61f2247552552544571dd898963816d878cee1b41633897c21d5b2daf17450107d4c71) u11916388)
  ```
- This generates a file, _perf.data_
- Use https://github.com/brendangregg/FlameGraph to generate a flame graph:
  ```sh
  flamegraph.pl perf.data > flame.svg
  ```

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -266,17 +267,37 @@ pub struct SDKOptions {
     pub track_costs: bool,
     #[wasm_bindgen(js_name = trackCoverage)]
     pub track_coverage: bool,
+    #[wasm_bindgen(js_name = trackPerformance)]
+    pub track_performance: bool,
+    performance_cost_field: String,
```
hugoclrd: Given that we don't know exactly what the SDK interface to generate flamegraphs should look like for now, maybe we shouldn't include these new options for now?

hugoclrd (changes requested):
I just left 1 subsantial comment for now, that I think can have a substantial impact for the PR if we decide to remove the track_performance options

components/clarinet-sdk-wasm/src/core.rs
tippenein: I removed the global perf cost field, but added a README on the intended use. Do you mean we should remove `track_performance` as well?

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -778,12 +788,22 @@ impl SDK {
...
+        // Release the session borrow before accessing self.costs_reports
+        let _ = session;
+
         if track_costs {
             if let Some(ref cost) = execution.cost {
                 let contract_id = if contract.starts_with('S') {
                     contract.to_string()
                 } else {
-                    format!("{}.{contract}", self.deployer)
+                    format!("{}.{contract}", deployer)
```
hugoclrd: I never get when does clippy complain about `{deployer}` and `{}`.. But it's fine as long as the CI is happy

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -795,7 +815,14 @@ impl SDK {
             }
         }
 
-        Ok(execution_result_to_transaction_res(&execution))
+        let mut response = execution_result_to_transaction_res(&execution);
+
+        // Add performance data if available
```
hugoclrd: remove

components/clarinet-sdk/node/src/sdkProxy.ts
```diff
@@ -146,6 +149,17 @@ export function getSessionProxy() {
...
+      if (prop === "enablePerformance") {
+        const enablePerformance = async (costField: string) => {
+          try {
+            session.enablePerformance(costField);
+          } catch (error) {
+            throw new Error(error as string);
+          }
+        };
+        return enablePerformance;
+      }
```
hugoclrd: I think this isn't needed.
We only need to override sdk methods when we want to have specific handling for the args or the returned value.
But here, we don't need to do anything different.

We can let `return Reflect.get(session, prop, receiver);` handle it

This is likely useless as well:
```ts
   : K extends "enablePerformance"
       ? (costField: string) => Promise<void>
```

components/clarinet-sdk/node/tests/performance.test.ts
```diff
@@ -0,0 +1,52 @@
...
+afterEach(() => {
+  deleteExistingDeploymentPlan();
+});
+
+describe("performance tracking", () => {
+  it("can track performance during simple operations", () => {
+    const result = simnet.execute("(+ 1 2)");
+    expect(Cl.prettyPrint(result.result)).toBe("3");
+
+    const perfData = simnet.collectPerformanceData();
```
hugoclrd: I think it could be useful to include to have it tested as well
```
    simnet.enablePerformance(...);
```

components/clarity-repl/src/repl/hooks/perf.rs
```diff
@@ -0,0 +1,313 @@
...
+pub enum CostField {
+    Runtime,
+    ReadLength,
+    ReadCount,
+    WriteLength,
+    WriteCount,
+}
+
+impl CostField {
+    pub fn parse_from_str(s: &str) -> Option<Self> {
```
hugoclrd: Implement this using the trait `From` or `TryFrom` for a more idiomatic approach

components/clarinet-sdk/node/src/sdkProxy.ts
```diff
@@ -42,14 +42,17 @@ export type Simnet = {
                 ? GetDataVar
                 : K extends "getMapEntry"
                   ? GetMapEntry
-                  : SDK[K];
+                  : K extends "enablePerformance"
+                    ? (costField: string) => Promise<void>
```
hugoclrd: Can be removed (see comment below)

hugoclrd (changes requested):
Requested a few changes but the overall logic and design lgtm

components/clarinet-sdk-wasm/src/core.rs
tippenein: Yea same. no idea

PR #1941, opened 2025-07-29
chore: improve vscode cargo fmt config

hugocaillard:
### Description

Make vscode rust analyzer setting follow fmt-stacks convention


obycode:
I had the same thought after needing to fix the formatting on #1940.

PR #1947, opened 2025-08-01
--wip-- [skip ci]

hugoclrd:


components/clarity-repl/src/repl/remote_data/context.rs
```diff
@@ -0,0 +1,100 @@
...
+                        contract_context,
+                        &mut call_stack,
+                        Some(publisher.clone()),
+                        Some(publisher.clone()),
+                        None,
+                    );
+
+                    let result = eval(exp, &mut env, &context)?;
+                    last_executed = Some(result);
+                    Ok(())
```
obycode: I think you can just do nothing for non-define expressions.

components/clarity-repl/src/repl/remote_data/context.rs
```diff
@@ -0,0 +1,100 @@
...
+            let mut call_stack = CallStack::new();
+            let mut env = Environment::new(
+                g,
+                contract_context,
+                &mut call_stack,
+                Some(publisher.clone()),
+                Some(publisher.clone()),
+                None,
+            );
+            functions::define::evaluate_define(exp, &mut env)
```
obycode: Do you actually need to evaluate these, or can you just build the context just based on the expressions? None of the names or types would be dependent on evaluation the expressions, only the values, which I think we shouldn't care about, since we'd still want to get those from the actual contract.

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -374,15 +387,90 @@ impl ClarityDatastore {
...
+        // option 1: custom set_contract_context (eval_all re-implemented)
+        context::set_contract_context(
+            &contract_ast.expressions,
+            &mut contract_context,
+            &mut global_context,
+        )?;
+
+        // option 2: call eval_all directly
+        // doesn't "just work"™
+        // values seems to conflict in the datastore
```
obycode: Is that just because this gets called multiple times for the same contract or is it a problem on the first call?

components/clarity-repl/src/repl/datastore.rs
```diff
@@ -374,15 +387,90 @@ impl ClarityDatastore {
         self.client.fetch_clarity_data(&url)
     }
 
+    #[allow(clippy::result_large_err)]
+    fn fetch_and_parse_source(
+        &mut self,
+        contract_id: &QualifiedContractIdentifier,
+    ) -> Result<ContractContext> {
+        let network_id = self
```
obycode: Could you cache these contexts rather than regenerating them?

obycode (commented):
Looking good! I added some quick thoughts.

components/clarity-repl/src/repl/datastore.rs
hugoclrd: So I get this error on what seems to be the first time I try to get (or insert) this particular metadata.
```
level: Error, message: "Runtime Error: Runtime error while interpreting ST000000000000000000002AMW42H.contract-0: Interpreter(Expect(\"Failed to initialize contract: Interpreter(Expect(\\\"Metadata entry 'vm-metadata::6::count' already exists for contract: STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter\\\"))\"))"
```


If I add logs at the beginning of `get_metadata` and `insert_metadata`, I get this:

```
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, analysis)
insert_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, analysis)
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, analysis)
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, analysis)
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, vm-metadata::9::contract-size)
insert_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, vm-metadata::9::contract-size)
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, vm-metadata::9::contract-data-size)
insert_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, vm-metadata::9::contract-data-size)
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, vm-metadata::9::contract)
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, analysis)
get_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, vm-metadata::6::count)
insert_metadata(STJCAB2T9TR2EJM7YS4DM2CGBBVTF7BV237Y8KNV.counter, vm-metadata::6::count)
```

But what I find odd, is that the error comes from here (from stacks-core `ClarityDatabase::insert_metadata`
https://github.com/hugoclrd/stacks-core/blob/799953e64e4733bc639a75f680cd216037446c9e/clarity/src/vm/database/clarity_db.rs#L694-L710). I'm not fully sure what DB it is at this point

components/clarity-repl/src/repl/remote_data/context.rs
hugoclrd: I'm not sure about wether this is needed or not.
Do we care about the actual value just here
```
DefineResult::Variable(name, value) => {
    contract_context.variables.insert(name, value);
}
```

Or not even?

components/clarity-repl/src/repl/datastore.rs
hugoclrd: So all of the metadata are cached in two ways:
- in memory, by design because they are inserted in the datastore
- on disk in `fetch_clarity_metadata`

This first time `get_metadata` is called with `vm-metadata::9::contract`, we compute (fetch the src and built context), but the next time, it will either be already available in the data store, or fetched on Disk.

At least i think that how it works

components/clarity-repl/src/repl/remote_data/context.rs
obycode: If we can make sure that these constants get pulled from the mxs and not from here, then we could ignore this value.

components/clarity-repl/src/repl/datastore.rs
obycode: Oh, so it would hit:
```
        if let Some(cached) = get_file_from_cache(&self.fs_cache_location, &cache_file_path) {
             return Ok(Some(cached));
         }
```
on the second time?

components/clarity-repl/src/repl/remote_data/context.rs
hugoclrd: Ok I see the point.
Well then I'm pretty confident that the data doesn't get the data from there inded.
I'll try to simplify to just the types then. 

What should I insert in `contract_context.variables`? Empty values?

components/clarity-repl/src/repl/datastore.rs
hugoclrd: The second time in the same session, it would hit this and `fetch_clarity_metadata` wouldn't even be called.
```rust
        let metadata = self.metadata.get(&(contract.to_string(), key.to_string()));
        if metadata.is_some() {
            return Ok(metadata.cloned());
        }
```

But indeed, if you start fresh, it would hit the on disk cache

hugoclrd: Does that make sense? It's also possible that I missed something

obycode: Yes, I think so.

components/clarity-repl/src/repl/remote_data/context.rs
obycode: Yeah, whatever is simplest to create.

obycode:
It looks like simnet is using the `none` stored for the constants, so there is a change needed somewhere else to fix that (or we do need to evaluate the initial values).

hugoclrd:
@obycode Could you briefly describe what are the two solutions you mentioned that may allow us to retrieve the constant value?

obycode:
> @obycode Could you briefly describe what are the two solutions you mentioned that may allow us to retrieve the constant value?

Sure, but beware these are just quick high-level ideas without really thinking through the implementation, so your intuition having written this PR is probably better than mine.
1. In `set_contract_context`, instead of filling in `none`s for the constant values, we could retrieve the real value from MXS.
2. In the datastore, in `fetch_clarity_metadata`, do something to ensure that values are retrieved from MXS and not from the local store (at least once and then cached?)

hugoclrd:
> so your intuition having written this PR is probably better than mine

My intuition isn't super reliable when it comes to the datastore

---
> 1. In set_contract_context, instead of filling in nones for the constant values, we could retrieve the real value from MXS.

That would mean doing 1 extra API call for each constant, which I'm afraid would be slow (and heavy on the api)

---
> 2. In the datastore, in fetch_clarity_metadata, do something to ensure that values are retrieved from MXS and not from the local store (at least once and then cached?)

So whenever we access `vm-metadata::9::contract`, we would need to both:
1. fetch it from the network (as it was the case before)
2. compute it locally to get the span
And then merge both results.

--- 

Why not going back to this [first implementation](https://github.com/hirosystems/clarinet/pull/1947/commits/b4036b6afef961c891206b5361a1fd8c8581d1ab#diff-2e387cb9b0480fef55e795d9c778580247b6075db86874700e924dfb4fd84ecaR37) where we evualute the code and get the right values?

Maybe it's a bit annoying to do this _just_ for the constants, but it seems ro be robust

obycode:
> Why not going back to this [first implementation](https://github.com/hirosystems/clarinet/pull/1947/commits/b4036b6afef961c891206b5361a1fd8c8581d1ab#diff-2e387cb9b0480fef55e795d9c778580247b6075db86874700e924dfb4fd84ecaR37) where we evualute the code and get the right values?

Yeah, maybe you're right. Sorry for the diversion!

PR #1950, opened 2025-08-04
chore: make stacks_signer_keys significant for devnet diff

### LINKED ISSUE #1946: Add `stacks_signers_keys` as a SignificantField in Devnet
`stacks_signers_keys` allow users to customize signer keys.
Setting two signers keys will boot the devnet with two signers:
```
stacks_signers_keys = [
  "7287ba251d44a4d3fd9276c88ce34c5c52a038955511cccaf77e61068649c17801",
  "530d9f61984c888536871c6573073bdfc0058896dc1adfe9a6a10dfacadc209101"
]
```

But the start from default snapshot won't work in the case. This field should be set as a SignificantField

tippenein:
- closes #1946

### DIFF
--- a/components/clarinet-files/src/lib.rs
+++ b/components/clarinet-files/src/lib.rs
@@ -5,6 +5,7 @@ extern crate serde_derive;
 
 pub mod clarinetrc;
 
+#[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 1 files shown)


hugoclrd (approved):
lgtm

PR #1951, opened 2025-08-05
refactor: rework session api reference

hugoclrd:
### Description

Refactor the get clarity `functions` and `keywords` in the repl session.
It sligthly simplifies the session, and will have a small performance boost. Probably negligible, but that's an easy refactor

### DIFF
--- a/components/clarity-repl/src/repl/docs.rs
+++ b/components/clarity-repl/src/repl/docs.rs
@@ -1,0 +1,70 @@
+use std::collections::HashMap;
+use std::sync::LazyLock;
+
+use clarity::vm::docs::{
+    make_api_reference, make_define_reference, make_keyword_reference, FunctionAPI,
+};
+use clarity::vm::functions::define::DefineFunctions;
+use clarity::vm::functions::NativeFunctions;
+use clarity::vm::variables::NativeVariables;
+
+fn normalize_description(description: &str) -> String {
+    description.replace('\n', " ")
+}
+
+fn format_api_doc(function_api: &FunctionAPI) -> (String, String) {
+    let description = normalize_description(&function_api.description);
+    let doc = format!(
+        "Usage\n{}\n\nDescription\n{}\n\nExamples\n{}",
+        function_api.signature, description, function_api.example
+    );
+    (function_api.name.clone(), doc)
+}
+
+fn build_api_reference() -> HashMap<String, String> {
+    NativeFunctions::ALL
+        .iter()
+        .map(|func| format_api_doc(&make_api_reference(func)))
+        .chain(
+            DefineFunctions::ALL
+                .iter()
+                .map(|func| format_api_doc(&make_define_reference(func))),
+        )
+        .collect()
+}
+
+fn clarity_keywords() -> HashMap<String, String> {
+    NativeVariables::ALL
+        .iter()
+        .filter_map(|func| {
+            make_keyword_reference(func).map(|key| {
+                let description = normalize_description(key.description);
+                let doc = format!("Description\n{}\n\nExamples\n{}", description, key.example);
+                (key.name.to_string(), doc)
+            })
+        })
+        .collect()
+}
+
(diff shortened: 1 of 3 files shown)


components/clarity-repl/src/repl/docs.rs
```diff
@@ -0,0 +1,70 @@
...
+        .collect()
+}
+
+fn clarity_keywords() -> HashMap<String, String> {
+    NativeVariables::ALL
+        .iter()
+        .filter_map(|func| {
+            make_keyword_reference(func).map(|key| {
+                let description = normalize_description(key.description);
+                let doc = format!("Description\n{}\n\nExamples\n{}", description, key.example);
```
jbencin: Clippy is gonna make you inline this. Also I think the `indoc` macros are preferable for multiline strings:

```suggestion
                let doc = formatdoc!("
                    Description
                    {description}
                    
                    Examples
                    {}", key.example);
```

components/clarity-repl/src/repl/docs.rs
```diff
@@ -0,0 +1,70 @@
...
+                (key.name.to_string(), doc)
+            })
```
jbencin: Pull the `map()` out of `filter_map()` to simplify logic and reduce indentation:

```suggestion
        .filter_map(make_keyword_reference)
        .map(|key| {
            let description = normalize_description(key.description);
            let doc = format!("Description\n{}\n\nExamples\n{}", description, key.example);
            (key.name.to_string(), doc)
        })
```

components/clarity-repl/src/repl/docs.rs
```diff
@@ -0,0 +1,70 @@
...
+        })
+        .collect()
+}
+
+pub static CLARITY_STD_REF: LazyLock<HashMap<String, String>> = LazyLock::new(build_api_reference);
```
jbencin, line 48: Good using `LazyLock` over `lazy_static` here, as that's the preferred way to do it now. We should probably declare an MSRV of Rust 1.80 somewhere

components/clarity-repl/src/repl/docs.rs
hugoclrd: @jbencin Clippy does run the check in the CI.
It's not complaining as long as one of the argument can not be inlined in the brackets (`key.example`)

components/clarity-repl/src/repl/docs.rs
hugoclrd: Like, if a project wants to use clarinet packages as a dependency?
That could make sense, also, we've been using std lazylock for a while now

jbencin: We can use `cargo msrv` to fix this in another PR, don't worry about it for now

components/clarity-repl/src/repl/docs.rs
```diff
@@ -0,0 +1,70 @@
...
+fn clarity_keywords() -> HashMap<String, String> {
+    NativeVariables::ALL
+        .iter()
+        .filter_map(make_keyword_reference)
+        .map(|key| {
+            let description = normalize_description(key.description);
+            let doc = format!("Description\n{}\n\nExamples\n{}", description, key.example);
+            (key.name.to_string(), doc)
+        })
+        })
```
jbencin: Extra closing bracket and paren here

hugoclrd: That was from the code you suggested 🫢

components/clarity-repl/src/repl/docs.rs
hugoclrd: infoc isn't a repl direct dependency of the repl right now. Plus this code has just been moved from a file to an other.
I think I'm fine with a few `\n` for this PR

components/clarity-repl/src/repl/docs.rs
jbencin: Yeah I messed up and didn't select the last line

hugoclrd:
Thx for the thorough review @jbencin

PR #1956, opened 2025-08-08
chore: update the deprecated `set-output` run commands

tippenein:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/


hugoclrd (approved):
lgtm

PR #1957, opened 2025-08-19
feat: further perf additions

tippenein:
- [x] Allow any element of the cost to be used - `::perf read_length <expression>`
- [x] Fix missing line/column numbers
- [x] Clarinet SDK addition via collectPerformanceData (same as costs)


components/clarity-repl/src/repl/hooks/perf.rs
```diff
@@ -32,24 +76,64 @@ pub struct PerfHook {
...
+                    std::fs::File::create(filename).expect("Failed to create perf output file"),
+                )
+            }
+        };
         PerfHook {
             writer,
             expr_stack: Vec::new(),
+            cost_field,
+            #[cfg(target_arch = "wasm32")]
+            buffer_data: Vec::new(),
```
obycode: Shouldn't this be the vec that is written to by the `writer`?

components/clarity-repl/src/repl/hooks/perf.rs
```diff
@@ -139,8 +235,19 @@ impl EvalHook for PerfHook {
...
+            self.cost_field.get_value(&cost)
+        )
+        .expect("Failed to write to perf output");
+
+        // In WASM mode, also store the data in our buffer
+        #[cfg(target_arch = "wasm32")]
+        {
+            let perf_line = format!("{call_stack} {}\n", self.cost_field.get_value(&cost));
+            self.buffer_data.extend_from_slice(perf_line.as_bytes());
+        }
```
obycode: See my comment above -- if `self.writer` is already writing to this buffer, then we don't need this special case.

components/clarity-repl/src/repl/hooks/perf.rs
```diff
@@ -163,3 +270,174 @@ impl EvalHook for PerfHook {
...
+    /// Build a simple mapping from src by parsing into SE
+    fn build_source_mapping(&mut self, _contract_source: &str) {}
+    /// Fetch contract source from API and cache it in the datastore
+    fn fetch_and_cache_contract_source(
+        &mut self,
+        env: &mut Environment,
+        contract: &QualifiedContractIdentifier,
+    ) -> Result<(), String> {
+        // Determine if this is mainnet or testnet based on the contract issuer
+        let is_mainnet = contract.issuer.to_address().contains("SP");
```
obycode: There really should be a method for this (I'll open a PR for it), but you can do:
```suggestion
        let version = contract.issuer.version();
        let is_mainnet = version == C32_ADDRESS_VERSION_MAINNET_MULTISIG
          || version == C32_ADDRESS_VERSION_MAINNET_SINGLESIG;
```

components/clarity-repl/src/repl/hooks/perf.rs
tippenein: the `buffer_data` is mainly for reading. We read from it to return the perf in the SDK. Maybe we can make some sort of ReaderWriter that works differently in native vs wasm

components/clarity-repl/src/repl/hooks/perf.rs
tippenein: Yea, there's a `mainnet` bool on the global context which is even easier 👍

obycode: Oh nice. For any future needs, I added `is_mainnet` and `is_multisig` in this PR - https://github.com/stacks-network/stacks-core/pull/6409/files.

hugoclrd:
@tippenein Can you describe in the PR description how would end-users use it?

tippenein:
@hugoclrd 
> Can you describe in the PR description how would end-users use it?

this will be merged into brice's PR before we go to `main` and his example is good. Probably worth some other external documentation as well though

hugoclrd:
@tippenein My question was mostly regarding the usage in sdk / unit-tests (and the slack discussions that followed)

PR #1958, opened 2025-08-25
fix: custom boot contracts LSP

tippenein:
This allows LSP functionality in custom boot contracts by detecting custom boot contracts in ContractOpened

### DIFF
--- a/components/clarity-lsp/src/common/backend.rs
+++ b/components/clarity-lsp/src/common/backend.rs
@@ -2,6 +2,7 @@ use std::sync::{Arc, RwLock};
 
 use clarinet_files::{FileAccessor, FileLocation, ProjectManifest};
 use clarity_repl::clarity::diagnostic::Diagnostic;
+use clarity_repl::clarity::vm::ClarityVersion;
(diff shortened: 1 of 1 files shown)


components/clarity-lsp/src/common/backend.rs
```diff
@@ -156,15 +157,35 @@ pub async fn process_notification(
...
+                                .unwrap_or_default()
+                                .to_string();
+
+                            if manifest
+                                .project
+                                .override_boot_contracts_source
+                                .contains_key(&contract_name)
+                            {
+                                // Custom boot contracts don't specify clarity versions
+                                ClarityVersion::latest()
```
hugoclrd: Can we use `get_boot_contract_epoch_and_clarity_version()` here?

hugoclrd (changes requested):

components/clarity-lsp/src/common/backend.rs
tippenein: Yes! good call

hugoclrd (approved):
lgtm!

PR #1969, opened 2025-09-02
fix: build

hugoclrd:
### Description

The build was weirdly broken in #1961 (shouldn't have passed CI, to be investigated).
This fixes it. And upgrade clarity-vm again

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -714,7 +714,7 @@ dependencies = [
 [[package]]
 name = "clarity"
 version = "0.0.1"
-source = "git+https://github.com/stacks-network/stacks-core.git?branch=develop#8f56fd4d45a88b25dc1c8869ee892edda579527a"
+source = "git+https://github.com/stacks-network/stacks-core.git?branch=develop#bc9adb35430a893ad1d56dc5234883d6d5147dbc"
 dependencies = [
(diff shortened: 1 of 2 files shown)


Cargo.lock
```diff
@@ -2229,7 +2229,7 @@ checksum = "e04d7f318608d35d4b61ddd75cbdaee86b023ebe2bd5a66ee0915f0bf93095a9"
 dependencies = [
  "hermit-abi",
  "libc",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
```
csgui: Why is this version being downgraded?

hugoclrd: idk. Maybe a change in how cargo resolved dependencies?
Many packages were using v0.59, but now, some use v0.52 and some other use v0.60.
But clarinet had both install anyway

hugoclrd: hmm, I tried installing/uninstalling certain version of certains packages (crossterm and comfy-table), I think the order on which certain packages are installed will resolve in different versions of the lockfile. That fine imo because all the impact package accept windows-sys 0.52 or 0.59.
But at least the changes are smaller now

csgui (approved):
Code LGTM.

PR #1972, opened 2025-09-08
fix: 'latest' serialization

### LINKED ISSUE #1967: Epoch `"latest"` get serialized to actual value
In Clarinet 3.5.0, we introduced a way to set `epoch = "latest"` in the manifesst, instead of an actual epoch ID (eg: 3.2).

But, when Clarinet serializes the manifest to toml (using commands such as `clarinet add | rm | requirements add`), it changes `"latest"` to the actual latest ID `3.2`

tippenein:
- closes #1967

### DIFF
--- a/components/clarinet-files/src/project_manifest.rs
+++ b/components/clarinet-files/src/project_manifest.rs
@@ -6,7 +6,7 @@ use clarity::types::StacksEpochId;
 use clarity::vm::ClarityVersion;
 use clarity_repl::repl;
 use clarity_repl::repl::boot::BOOT_CONTRACTS_NAMES;
-use clarity_repl::repl::{ClarityCodeSource, ClarityContract, ContractDeployer, DEFAULT_EPOCH};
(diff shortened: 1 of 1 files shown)


components/clarinet-files/src/project_manifest.rs
```diff
@@ -427,43 +427,46 @@ impl ProjectManifest {
...
-            "2.5" => StacksEpochId::Epoch25,
-            "3" | "3.0" => StacksEpochId::Epoch30,
-            "3.1" => StacksEpochId::Epoch31,
-            "3.2" => StacksEpochId::Epoch32,
-            _ => return Err(
-                "epoch field invalid (value supported: 2.0, 2.05, 2.1, 2.2, 2.3, 2.4, 3.0, 3.1, 3.2, latest)"
-                    .into(),
-            ),
-        },
+        Some("latest") => clarity_repl::repl::Epoch::Latest,
```
tippenein, line 438: this is the substantive change

PR #1974, opened 2025-09-09
feat: show stack trace when a contract call fails

hugoclrd:
### Description

When a contract call fails in simnet (repl or sdk), it's currently difficult to understand why.
This PR attempts to fix it by showing a stacks trace and a cleanr error when that happens

Screenshot 2025-09-10 at 19 14 49

Note: the sdk is currently limited in terms of stdout or stderr message formatting (color, bold, italic, etc). That should be approached in an other PR.

PR #1982, opened 2025-09-15
feat: show stacks trace on test fail

hugoclrd:
### Description

In #1974, I introduced improvement in the stack trace hook and better support for it in the SDK.
It shows the traces with a better error message whenever a contract-call fails (when there's an error in the vm)

This takes it a step further, and allows the test runner to show the stacks trace whenever a test fails.


tippenein (approved):
LGTM

PR #1986, opened 2025-09-15
chore: warn if falling back to runtime cost perf

tippenein:
Caused confusion if silently falling back to `runtime` perf
Now we warn and give the possible argument names


hugoclrd (approved):
lgtm 👍

PR #1990, opened 2025-09-16
Fix/boot contracts deployment

hugoclrd:
### Description

- Make sure to deploy boot contract and all simnet Sessions. This is handled by directly deploying boot contracts in `Session::new()`

This PR also:
- Funds genesis accounts with STXs in `Session::new()`
- Removes the needs to call `Session::start()`
- Refactors some part of the code:
  - Simplify the logic to deploy boot contract and remove duplicated logic betwen `clarinet-deployments/src/lib.rs` and `components/clarity-repl/src/repl/session.rs`
  - Don't create new Sessions in `generate_default_deployment()`


components/clarinet-deployments/src/lib.rs
```diff
@@ -472,8 +451,6 @@ pub async fn generate_default_deployment(
             let (epoch, clarity_version) =
                 get_boot_contract_epoch_and_clarity_version(contract_name.as_str());
 
-            // Set the session to the correct epoch for validation
-            session.update_epoch(epoch);
```
hugoclrd, line 476: @tippenein We commented a lot on this line during the override boot-contracts PR.

I still think that this is probably not necessary because we only call `interpreter.build_ast(&contract)`, which calls `clarity::vm::ast::build_ast_with_diagnostics`, which takes the epoch as an arguments (provided in `&contract`), and doesn't read if from the state of the session (or the datastore).

So I think we're fine. I still have to run more manuel tests on the PR, and especially makes sure that it properly handles boot contracts override 👍

components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -125,16 +125,6 @@ impl ClarityInterpreter {
...
-        &mut self,
-        contract: &ClarityContract,
-        ast: Option<&ContractAST>,
-        cost_track: bool,
-        eval_hooks: Option<Vec<&mut dyn EvalHook>>,
-    ) -> Result<ExecutionResult, Vec<Diagnostic>> {
-        self.run_interpreter(contract, ast, cost_track, eval_hooks)
-    }
-
-    fn run_interpreter(
```
hugoclrd, line 137: `run` and `run_interpreter` remained from the time we had `run_wasm` that would run `clarity-wasm` and compare the result from `interpreter` and `wasm`

components/clarity-repl/src/repl/session.rs
```diff
@@ -60,25 +103,21 @@ pub struct Session {
...
-                .expect("Unable to parse deployer's address")
-        };
+        let mut interpreter = ClarityInterpreter::new(
+            settings.get_default_sender(),
+            settings.repl_settings.clone(),
+            settings.cache_location.clone(),
+        );
+
+        set_up_accounts(&settings.initial_accounts, &mut interpreter);
+        let boot_contracts = deploy_boot_contracts(&settings, &mut interpreter);
```
hugoclrd, line 108: The main change of this PR.
Directly set up genesis accounts and boot_contracts here

tippenein (approved):
Code looks good. We definitely need to retest the LSP integration with these changes. I don’t see anything that sticks out but it was a game of whack-a-mole last time iirc

components/clarity-repl/src/repl/session.rs
tippenein: 💯

tippenein (approved):

PR #1998, opened 2025-09-19
chore: add overhead costs to perf.data to match get_costs

tippenein:
The overhead cost of contract calls was not represented in perf.data. This adds the overhead cost explicitly to the flamegraph so that it will match the get_costs total

Screenshot 2025-09-19 at 10 27 40 AM

Screenshot 2025-09-19 at 10 27 48 AM

The `get_costs` result for the same contract call
Screenshot 2025-09-19 at 10 29 54 AM

### DIFF
--- a/components/clarity-repl/src/repl/hooks/perf.rs
+++ b/components/clarity-repl/src/repl/hooks/perf.rs
@@ -151,7 +151,7 @@ pub struct PerfHook {
     collected_data: Vec<(String, u64)>,
 }
 
-static OVERHEAD_HEADER: &str = "contract-overhead";
+static OVERHEAD_HEADER: &str = "contract-loading-cost";
 
 impl Clone for PerfHook {
     fn clone(&self) -> Self {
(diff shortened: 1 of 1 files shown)


components/clarity-repl/src/repl/hooks/perf.rs
```diff
@@ -143,8 +143,16 @@ pub struct PerfHook {
...
     cost_field: CostField,
+    /// Sum of all expression costs tracked
+    total_expression_costs: ExecutionCost,
+    /// Contract identifier for overhead tracking
+    contract_identifier: Option<QualifiedContractIdentifier>,
+    /// Collected performance data to rewrite with overhead
+    collected_data: Vec<(String, u64)>,
 }
 
+static OVERHEAD_HEADER: &str = "contract-overhead";
```
hugoclrd: As discussed, this should be renamed

components/clarity-repl/src/repl/hooks/perf.rs
```diff
@@ -311,8 +323,83 @@ impl EvalHook for PerfHook {
...
+                    } else {
+                        // overhead failed for some reason, write the collected data
+                        let mut file = std::fs::File::create("perf.data")
+                            .expect("Failed to create perf output file");
+                        for (call_stack, cost) in &self.collected_data {
+                            writeln!(&mut file, "{} {}", call_stack, cost)
+                                .expect("Failed to write to perf output");
+                        }
+                    }
+                }
```
hugoclrd: I think we can limit some code duplication between wasm32 and no wasm32 code.
What about keeping the writer into a variable, and whole the rest can be target agnostic?

```rust
                #[cfg(not(target_arch = "wasm32"))]
                let mut writer =
                    std::fs::File::create("perf.data").expect("Failed to create perf output file");

                #[cfg(target_arch = "wasm32")]
                self.clear_buffer();
                #[cfg(target_arch = "wasm32")]
                let writer = &mut self.writer;
```


There are already difference between the two implementation (such as the order of the conditions), and I'm afraid it will be hard to maintain in the future

hugoclrd (changes requested):
Some refactoring changes

components/clarity-repl/src/repl/hooks/perf.rs
tippenein: Yes! This works well. Also collapsed those nasty nested if/else 👍

PR #2002, opened 2025-09-23
 fix(devnet): implement graceful shutdown for SIGTERM signals

### LINKED ISSUE #1391: Sending SIGINT/SIGTERM to the process started by clarinet dev start doesn't trigger a graceful shutdown
**Describe the bug**
When sending a SIGINT or SIGTERM to the clarinet process spawned when executing `clarinet dev start` does not trigger a graceful shutdown. Instead, it kills the process ungracefully

**To Reproduce**
1. Run `clarinet dev start`
2. Run `ps -a | grep clarinet` in another terminal session and grab the PID
3. Run `kill -TERM <PID>` or `kill -INT <PID>`

**Expected behavior**
The clarinet process should trigger a graceful shutdown whenever SIGINT or SIGTERM signals are received

 - OS (version): MacOS Ventura 13.2
 - Docker version: Orbstack latest version
 - Clarinet version: 2.2.1

BiliqisO:
### Description
This PR fixes #1391 

I basically added a signal handler for SIGTERM that forwards to the existing SIGINT graceful shutdown mechanism. 

#### Breaking change?
No

### Example

---

### Checklist

[https://drive.google.com/file/d/1dgbkJ4w68xTo-EZX6U9gh2lodIhy2Qqv/view?usp=drivesdk](url)
[https://drive.google.com/file/d/1qiCKPziM8CJK6zzbz9itKAzMU6BVnUj8/view?usp=drivesdk](url)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,27 @@
+# [3.8.1](https://github.com/hirosystems/clarinet/compare/v3.8.0...v3.8.1) (2025-10-10)
+
+##### Chores
+
+- Fix transaction spec contract publish (Clarity 4) (#2020) (8049deab)
+
+# [3.8.0](https://github.com/hirosystems/clarinet/compare/v3.7.0...v3.8.0) (2025-10-09)
+
+##### New Features
+
+- Update clarity-vm with Clarity 4 preview (#2013) (7e53dc5e)
+- Warn on some noop builtins (#2006) (248a1ac3)
+
+##### Bug Fixes
+
+- Contract id parsing (#2016) (4101e3a7)
+- Publish stacking orders regardless of snapshot (#2015) (7b2db50f)
+- Properly execute commands on windows (#2011) (62a32bfb)
+
+##### Chores
+
+- Upgrade stacks node and signer default docker images (#2017) (d36b760b)
+- Remove stacks-devnet-js (#2005) (3761eac0)
+
 # [3.7.0](https://github.com/hirosystems/clarinet/compare/v3.6.1...v3.7.0) (2025-09-23)
 
 ##### New Features
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -575,7 +575,7 @@ checksum = "b94f61472cee1439c0b966b47e3aca9ae07e45d070759512cd390ea2bebc6675"
 
 [[package]]
 name = "clarinet-cli"
-version = "3.7.0"
+version = "3.8.1"
 dependencies = [
  "clap",
  "clap_complete",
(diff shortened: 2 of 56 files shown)


hugoclrd:
Hey @BiliqisO, looks like this PR includes some unwanted commits that have already been merged.
It messes with the files changes makes the PR hard to review.
Can you please fix the history?

BiliqisO:
Hi, I have fixed the history.

BiliqisO:
Hello, The CI workflow tests are failing because of windows compatibility, 
So i was thinking of adding this ctrlc library option for windows compatiblity or just leave it out and allow for windows users to make use of the normal command + c command
What do you think?
```
   #[cfg(windows)]
        {
            if let Err(e) = ctrlc::set_handler(move || {}) {
                eprintln!("Failed to setup signal handler: {}", e);
            }
        }
````

**Comment by hugoclrd** [comment]
Hey @BiliqisO,
So the devnet is already relying on ctrlc [here](https://github.com/hirosystems/clarinet/blob/5c10c0c51327f6e3e3063fd8c5f317ff861aec6d/components/stacks-network/src/lib.rs#L264-L270) and also has [this implementation](https://github.com/hirosystems/clarinet/blob/5c10c0c51327f6e3e3063fd8c5f317ff861aec6d/components/stacks-network/src/ui/mod.rs#L126-L139) to be stopped.
They respectively handle the "--no-dashboard" and "normal" modes.

You might want to build on top of this to properly handle SIGTERM and SIGINT in all situations

File: components/stacks-network/src/lib.rs [hunk 1]
```diff
@@ -259,16 +331,13 @@ async fn do_run_devnet(
...
-        let moved_observer_command_tx = observer_command_tx;
-        let moved_mining_command_tx = mining_command_tx;
-        let _ = ctrlc::set_handler(move || {
-            let _ = moved_orchestrator_terminator_tx.send(true);
-            let _ = moved_observer_command_tx.send(ObserverCommand::Terminate);
-            let _ = moved_mining_command_tx.send(BitcoinMiningCommand::Pause);
-            sleep(Duration::from_secs(3));
-            let _ = devnet_events_tx.send(DevnetEvent::Terminate);
-        });
+        // Set up signal handlers for graceful shutdown (SIGINT and SIGTERM)
```
tippenein: Could we move the `setup_signal_handlers` above the `if` and remove the `else` here? It looks functionally the same without the duplication

BiliqisO: ✅

tippenein: Thanks!

tippenein:
@BiliqisO You can run `cargo clippy --workspace --exclude clarinet-sdk-wasm` to fix up the lint warnings. I think this looks pretty good though!

PR #2010, opened 2025-09-26
chore: bring chainhook back

hugoclrd:
This is a big PR (+26k) because it copies all of [hirosystems/chainhook](https://github.com/hirosystems/chainhook) chainhook-sdk and chainhook-type to the Clarinet repo.
At the same time, it upgrades some dependencies, removes prometheus, and refactor the chainhook server so that it uses Axum instead of Rocket. Rocket is an unmaintained http framework, replacing it with Axum removed 49 dependencies for clarinet, and often old version of those dependencies (hyper 0.14, http 0.2, and more).

In total, this PR removes 55 dependencies from the CLI. See the Cargo.lock

The Devnet was carefully tested to make sure we have regression.

## Context

**The what:**
Bring back [hirosystems/chainhook](https://github.com/hirosystems/chainhook) chainhook-sdk and chainhook-type into the Clarinet repository.
Clarinet rely on chainhook v1 to orchestrate the Devnet.

**The why:**
There is a circular dependency between clarinet <-> chainhook, making it hard to maintain and upgrade both.

**Why now:**
Both of these repositories rely on depend on stacks-network/stacks-core/clarity, which is pulled from github. Upgrading the clarity vm means maintaining two branches in parallel on Clarinet and Chainhook, that depend on each other. 
Because Chainhook v1 will be discontinued and replaced by v2, it makes sense to have in Clarinet for now.

**Side effects and benefits:**
Having chainhook-sdk in the clarient repo will make it way easier and faster for us to maintain and upgrade our dependency tree. We are able to greatly simplify chainhhok and remove around 50 crates for clarinet dependencies.

## Future work

It should be possible for clarinet to fully remove the need for chainhook-sdk, a general purpose tool, and just have a basic event observer that only serves the needs of the Devnet.


Cargo.toml
```diff
@@ -59,14 +65,6 @@ indoc = "2.0.3"
 # dev dependencies
 mockito = { version = "1.6", default-features = false }
 
-[patch.crates-io]
-stacks-codec = { path = "./components/stacks-codec" }
-hiro-system-kit = { path = "./components/hiro-system-kit" }
-chainhook-sdk = { git = "https://github.com/hirosystems/chainhook.git", branch = "chore/update-clarinet-and-clarity" }
-chainhook-types = { git = "https://github.com/hirosystems/chainhook.git", branch = "chore/update-clarinet-and-clarity" }
-# chainhook-sdk = { path = "../chainhook/components/chainhook-sdk" }
-# chainhook-types = { path = "../chainhook/components/chainhook-types-rs" }
```
hugoclrd, line 68: Because everything is now only used internally, it doesn't need to be patched anymore

Cargo.toml
```diff
@@ -1,6 +1,8 @@
 [workspace]
 resolver = "2"
 members = [
+    "components/chainhook-sdk",
+    "components/chainhook-types",
```
csgui, line 5: Ok. Basically this PR brings back `chainhook-sdk` and `chainhook-types`.

hugoclrd: exactly. The code has been copied and simplified (clippy, remove dead code, use axum instead of rocket, this is the most important commit of this PR https://github.com/hirosystems/clarinet/pull/2010/commits/e2de07f9434500b6be7d8ed2a3d37214fe916b30)

csgui:
What would be the effort required to replace Chainhook V1 with V2? I mean, would all the changes in this PR need to be reverted again?

hugoclrd:
@csgui
Discussion around Chainhook 2 are still in progress. What's clear for now is that Clarinet will likely not rely on Chainhook 2 for its internal logic (listen to new block events to orchestrate devnet). Chainhook 2 is a completely different tech (different use case, not Rust, etc).
So for the foreseeable future, Clarinet will keep an internal tool to observe the stacks and burn chains (eg: currently this chainhook-sdk thing, but it could be simplified in the future - see "future work in the PR desc").

If the Chainhook v2 folks want Clarinet to support it for local development and testing, it will likely more likely be an extra docker container in the devnet, rather than something built in like chainhook v1

tl;dr: no, this PR won't be reverted, Clarinet needs a way to observe the chains and react to events

csgui:
Ok! Thanks for the answer @hugoclrd.

As you mentioned it as future work, it would be great to replace the `chainhook` components and references with something in "clarinet-house" solution for event observability. That would help avoid confusion and reliance on external projects (specially a discontinued one).

hugoclrd:
@csgui Agreed, but for now, this is still chainhook, and you can still setup chainhook (v1) predicates in Clarinet https://docs.hiro.so/en/tools/clarinet/chainhook-integration

The rewording / branding should likely happen once Hiro deprecates v1.

Thanks for the review!

PR #2011, opened 2025-09-30
fix: properly execute commands on windows

hugoclrd:
### Description

Adding a run_command helper to properly execute commands as [documented](https://doc.rust-lang.org/std/process/struct.Command.html)

```rs
pub fn run_command(command: &str) -> std::io::Result<std::process::Output> {
    if cfg!(target_os = "windows") {
        Command::new("cmd").args(["/C", command]).output()
    } else {
        Command::new("sh").arg("-c").arg(command).output()
    }
}
```


brice-stacks (approved):
I did not test this on Windows, but it looks good to me.

hugoclrd:
Thanks @obycode 
It was tested on windows and seems to work well 👍

PR #2013, opened 2025-10-02
feat: update clarity-vm with clarity 4

hugoclrd:
### Description

Update the clarity-vm, with preview of the new Clarity 4 features
See the [SIP](https://github.com/obycode/sips/blob/7c78d3af982826b4a4a7e3fd0266af97cdc61802/sips/sip-clarity4/sip-clarity4.md)

- [x] Epoch 3.3
  - [x] In Simnet
  - [x] Optional support in Devnet
- [x] Clarity 4
  - [x] Support new version
  - [x] Test and support certain features in Simnet:
    - [x] `(contract-hash)`
    - [x] `stacks-block-time`
    - [x] `current-contract`
- [x] costs-4

Note: a lot of stacks-core structs don't implement `Copy` anymore, hence the addition of the many `.clone()`

### DIFF
--- a/components/clarity-repl/src/repl/interpreter.rs
+++ b/components/clarity-repl/src/repl/interpreter.rs
@@ -2002,7 +2002,7 @@ mod tests {
         let src = indoc!(
             r#"(define-read-only (get-time) {
                 previous: (get-stacks-block-info? time (- stacks-block-height u1)),
-                current: block-time,
+                current: stacks-block-time,
             })"#
         );
         let contract = ClarityContractBuilder::default()
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -700,7 +700,7 @@ dependencies = [
 [[package]]
 name = "clarity"
 version = "0.0.1"
-source = "git+https://github.com/stacks-network/stacks-core.git?branch=develop#b5e3baf7f03cf79d3f5d956f68ab0745be1ed69f"
+source = "git+https://github.com/stacks-network/stacks-core.git?branch=develop#483f1adef6b71657783b1c7dbd2c28f06a65e134"
 dependencies = [
  "clarity-types",
  "integer-sqrt",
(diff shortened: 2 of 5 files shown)


components/clarity-repl/src/repl/datastore.rs
```diff
@@ -963,18 +981,26 @@ impl Datastore {
...
             let block_info = self.build_next_stacks_block(clarity_datastore);
-            self.stacks_blocks.insert(id, block_info);
+
+            clarity_datastore
+                .put_all_data(vec![(
+                    CLARITY_STORAGE_BLOCK_TIME_KEY.to_string(),
+                    format!("{}", block_info.stacks_block_time),
+                )])
+                .expect("failed to update vm-epoch::block-height");
+            self.stacks_blocks.insert(id.clone(), block_info);
```
hugoclrd: Doew that seem right to you @obycode?

If so, I'll open a quick PR on stacks-core just to make `CLARITY_STORAGE_BLOCK_TIME_KEY` public

components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -276,7 +275,7 @@ impl ClarityInterpreter {
             contract.clarity_version,
             true,
         )
-        .map_err(|(error, _)| error.diagnostic)?;
+        .map_err(|boxed_error| boxed_error.0.diagnostic)?;
```
tippenein, line 278: just curious, is there a performance difference between tuple access and destructuring like it was previously?

hugoclrd: The difference is that `run_analysis` now returns `Result<ContractAnalysis, Box<(CheckError, LimitedCostTracker)>>` instead of `Result<ContractAnalysis, (CheckError, LimitedCostTracker)>`, and Box can not be destructured.

And the error being boxed is also what allows us to remove all of the ` #[allow(clippy::result_large_err)]`

components/clarity-repl/src/repl/datastore.rs
brice-stacks: Yeah, that seems fine to me. There is a `ClarityDatabase` function, `setup_block_metadata`, that puts this data, but I'm not sure if that is convenient to use here.

Issue #1006, opened 2023-05-11
Provide Clarinet.toml schema


hugocaillard:
https://github.com/tamasfe/taplo/tree/master/editors/vscode#completion-and-validation-with-json-schema

hugoclrd:
Hey @BiliqisO,
This one would be a very good issue for you. I've been postponing it for too long.

tl;dr: providing JSON Schema of the various config files that Clarinet, starting with the manifest (`Clarinet.toml`) would be a huge improvement for our developers.
The json schema could be manually written, but it would be awesome if would generate it dynamically based on the `ProjectManifestFile` struct.
https://github.com/hirosystems/clarinet/blob/3761eac0d9d7dea662a17d9150f75df139b4019f/components/clarinet-files/src/project_manifest.rs#L45-L50

Let me know if you are interested in this issue and if it makes sense you

BiliqisO:
Hi..... I'm so sorry, I am just seeing this.  I'll have a look and revert 🙏🏻

BiliqisO:
Hi... Yes, I'm interested in this issue.

I found this crate [https://crates.io/crates/schemars](https://crates.io/crates/schemars)  for JSON Schema derivation, but the  TomlValue enum comes from an external crate, so I can't implement JsonSchema on it. So I think I should write a custom conversion function since the ProjectManifestFile struct data content is defined. What do you think?

hugoclrd:
Hey, I actually had pretty a bad experience with schemars in the past, but seeing that it reached v1.x is a good sign.
I'm also all-in for avoiding adding dependencies, so I think that a custom function conversion will be fine. It's just to convert one struct, so i think we can keep things simple enough.

Don't hesitate to open draft PRs if you need early feedback and let me know if you have any blocker or want to reconsider the approach at some point

BiliqisO:
Okay, got it. Thanks for the swift response😃

BiliqisO:
Hi,  @hugoclrd. Can you have a look? [https://github.com/hirosystems/clarinet/pull/2014](https://github.com/hirosystems/clarinet/pull/2014)

PR #2014, opened 2025-10-02
feat:: Clarinet.toml JSON Schema

BiliqisO:
### Description

This PR  addresses issue #1006 

- I wrote a custom function to generate JSON Schema from ProjectManifestFile struct.

#### Breaking change?

No

### Example

The generated JSON Schema for `Clarinet.toml` project manifest files is in [components/clarinet-files/schemas/clarinet-manifest.schema.json](https://github.com/BiliqisO/clarinet/blob/492e8cecfad6ec4ba706d466c1852a6fe11c590c/components/clarinet-files/schemas/clarinet-manifest.schema.json)

---

### Checklist


hugoclrd:
@BiliqisO 
It looks like the output of `generate_clarinet_manifest_schema` is hardcoded JSON. And not generated from the ProjectManifestFile struct.
Would it be possible to dynamically generate the json instead?

BiliqisO:
Hi, I assumed I was supposed to use the structure of the ProjectManifestFile struct to generate the JSON.  Can you guide me on an alternative approach I could use to generate this Schema, 🙏🏻

hugoclrd:
> use the structure of the ProjectManifestFile struct to generate the JSON.

Right 👍  With a function that dynamically generates the JSON based on `ProjectManifestFile`. So that when we change `ProjectManifestFile`, we can automatically generate the JSON schema.
Basically what schemars offers

BiliqisO:
Hi, @hugoclrd. So I eventually went with using the schemars crate, then created custom struct & enum for the data types that are referenced by TomlValue. eg 

```
  #[schemars(schema_with = "schema::contracts_schema")]
    contracts: Option<TomlValue>,
```

to 👇🏼
```
pub(crate) struct ContractConfig {
    path: String,
    #[serde(skip_serializing_if = "Option::is_none")]
    deployer: Option<String>,
    #[serde(skip_serializing_if = "Option::is_none")]
    #[schemars(schema_with = "clarity_version_schema")]
    clarity_version: Option<u8>,
    #[serde(skip_serializing_if = "Option::is_none")]
    epoch: Option<EpochValue>,
}
```
derived the content for the new struct from  [this block of code](https://github.com/BiliqisO/clarinet/blob/968004f396ef9559de7f4119d8fc29326caf576f/components/clarinet-files/src/project_manifest.rs#L372C18-L401C23)

similarly for this

```
pub(crate) enum EpochValue {
    #[schemars(schema_with = "epoch_string_schema")]
    String(String),
    #[schemars(schema_with = "epoch_number_schema")]
    Number(f64),
}
```

hugoclrd:
Thanks @BiliqisO,

This is looking good. A few questions/remarks:

1️⃣ 
So how did you generate `components/clarinet-files/schemas/clarinet-manifest.schema.json`? Could we add an alias in `.cargo/config.toml` that calls `generate_clarinet_manifest_schema()` and generates it?

2️⃣ 
I think we can add `components/clarinet-files/schemas/clarinet-manifest.schema.json` to the gitignore. It doesn't need to be pushed, we'll generate it in the CI and can make it avaialble in the vscode extension, or push it to schemastore.org.

3️⃣ 
Because the schema is only generated one time, on te fly, we actually don't need to have schemars always enabled.
What I mean is that it could be optional, behind a feature flag:
```toml
# components/clartinet-files/cargo.toml
schemars = { workspace = true, optional = true }

#... 

[features]
json_schema = ["schemars"]
```

And then in the structs definitions:
```rs
#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize)]
#[cfg_attr(feature = "json_schema", derive(JsonSchema))]
pub struct Settings {
}
```

This would make it more efficient, and only enable Schemars when we need to generate schema

---

Note: We merged a PR that involves a few changes due to a clarity-vm upgrade, leading to a few conflicts, but nothing major I believe. Let me know if you need help with that.

components/clarinet-files/schemas/clarinet-manifest.schema.json
```diff
@@ -0,0 +1,332 @@
...
+      "$ref": "#/$defs/SettingsFile",
+      "description": "REPL and analysis settings"
+    }
+  },
+  "required": [
+    "project",
+    "contracts",
+    "repl"
+  ],
+  "title": "ProjectManifestFile",
```
hugoclrd: Does schemars let you rename the title to something like `Clarinet Manifest files`?

components/clarinet-files/src/project_manifest.rs
```diff
@@ -26,26 +28,50 @@ pub struct ClarityContractMetadata {
...
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, JsonSchema)]
+#[schemars(rename_all = "snake_case")]
 pub struct ProjectConfigFile {
+    /// Name of the project
     name: String,
+    /// List of project authors
+    #[schemars(skip_serializing_if = "Option::is_none")]
```
hugoclrd: I don't understand why there is `#[schemars(skip_serializing_if = "Option::is_none")]`.

Schemars serializes the schema of the struct, the values shouldn't be relevant

hugoclrd (changes requested):

BiliqisO:
> So how did you generate `components/clarinet-files/schemas/clarinet-manifest.schema.json`? Could we add an alias in `.cargo/config.toml` that calls `generate_clarinet_manifest_schema()` and generates it?

1. I generated it by creating this [test](https://github.com/BiliqisO/clarinet/blob/492e8cecfad6ec4ba706d466c1852a6fe11c590c/components/clarinet-files/src/schema.rs#L261-L275) , its supposed to be a test file,  I didnt remove or delete it, cos I wanted to share the generated schema as an example on the PR, but I'd remove it in my next commit.
2. Got that, will do. Thanks.
3. RIght., I made this PR/commit for you to have a look and see if my approach was okay basically... I'll make the corrections. Thanks sm

components/clarinet-files/schemas/clarinet-manifest.schema.json
BiliqisO: Yes, it does.. but this one was generated automatically with the schemars crate.

Could you have a look at my latest changes

hugoclrd (changes requested):
Hey @BiliqisO 
Sorry for the delay between reviews, and thanks for your reactivity.

I tested the schema, and it works well.
Here's how I tested, in the [hirosystems/clarity-starter](https://github.com/hirosystems/clarity-starter/) project, a copied the schema under `./schema.json`, then added `#:schema ./manifest-schema.json` at the beginning of `Clarinet.toml`.
I have a toml extension in vscode ("even better toml") that picks it up and validates the schema.

I noticed the schema has too many `required` fields, indeed, the Clarinet.toml is pretty flexible,

`"contracts"`, `requirements`, and `"repl"` properties are not required.

In other words, this is a valid manifest
```toml
[project]
name = 'project-template'
```

Can you make the changes?
Then we'll be good to merge 👍

BiliqisO:
Yes, I noticed the required fields as well; it was generated from the Schemars crate.  I noticed they're coming from the option enum in the struct. I've made them optional by adding #[serde(default)] to the fields generated from the Option enum type.

hugoclrd (approved):
LGTM! Thanks a lot @BiliqisO

FuzzysTodd (commented):
#[derive(Debug, Default, Clone, Copy, Serialize, Deserialize)]
#[cfg_attr(feature = "json_schema", derive(JsonSchema))]
pub struct Settings {
}

PR #2016, opened 2025-10-07
fix: contract id parsing

hugoclrd:
The previous condition would (obviously) fail to parse contract_id that started with an `S` (eg: `Super_NFT`). I guess we never had the issue because contract names tend to start with a lower case.

This method should be more robust, and has a better error handling (one less `unwrap`


tippenein (approved):
lgtm

PR #2023, opened 2025-10-20
fix: match lining

brady-stacks:
fixes an issue where match results aren't lined properly


brady-stacks:
https://github.com/stx-labs/clarinet/pull/2029

PR #2031, opened 2025-10-23
fix: requirements fetching

hugo-stacks:
### Description

- The contract source rpc endpoint actually doesn't return the clarity version
- So Clarinet was relying solely on the published block height to determine the clarity version of a contract (using `ClarityVersion::default_for_epoch()`

Here is a contract that was published in epoch 3.x but using Clarity v2
https://explorer.hiro.so/txid/SP4SZE494VC2YC5JYG7AYFQ44F5Q4PYV7DVMDPBG.ststxbtc-token-v2?chain=mainnet

It was leading to a bug where Clarinet couldn't parse the contract since it contains "block-height" - which has been removed in clarity 3.

Ideally, we'd only want to fetch the stacks node RPC endpoints Clarinet, but here it would mean making two api calls also to fetch `/v2/contracts/interface/` - so we can fetch the Hiro API instead.

--- 

This part would deserve a unit or integration test, but it's not.. let's do it as a hotfix for now


hugo-stacks:
Thanks @y0s0n for tracking down this bug 🙏

Issue #1989, opened 2025-09-16
Allow to mint any kind of SIP 010 tokens in Simnet


hugoclrd:
For testing purpose, the Simnet (Repl and SDK) should offer a way for developers to mint any sort of SIP 010 tokens.

## Implementation details

It might be possible to directly call `Interpreter::credit_token()` for any token (it's currently only used for STX)

xoloki-stacks:
The current PR #2033 is in review, and as suggested only adds `repl` support.  The commit which adds SDK support is also done, and is pretty small: 

https://github.com/xoloki-stacks/clarinet/commit/f0bb2e38f387cbb146893521bb000b4472c83e27

It would seem to be reasonable to fold that commit into #2033 rather than have two PRs.

PR #2033, opened 2025-10-25
Implement `mint_ft` functionality in interpreter and console

xoloki-stacks:
### Description

This PR addresses #1989 by adding `mint_ft` functionality to the interpreter and console.  It treats fungible tokens the same as `STX` in terms of manipulating both the interpreter and clarity vm db balances when minting, and this is confirmed in the added test.

One oddity which I discovered during testing is that when a contract is deployed, if there are functions called that fail, nothing indicates this in the `ExecutionResult` returned from `deploy_contract`.  Successful function calls do return a `StacksTransactionEvent`.

#### Breaking change?

No APIs have been altered, so this change should not break anything.

### Checklist

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -1229,7 +1229,8 @@ impl Session {
 
         let (contract_identifier_str, asset_name_str) = identifier.split_at(index);
 
-        let Ok(contract_identifier) = Self::desugar_contract_id(deployer, contract_identifier_str)
+        let Ok(contract_identifier) =
+            Self::desugar_contract_id(default_deployer, contract_identifier_str)
         else {
             return Err(AssetIdentifierParseError::ContractIdentifierParseError);
         };


components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -817,6 +819,66 @@ impl ClarityInterpreter {
...
+                &contract_identifier,
+                &token_name,
+                recipient,
+                cur_balance + amount as u128,
+            ).map_err(|e| format!("failed to set_ft_balance for {contract_identifier}.{token_name} {recipient}: {e:?}"))?;
+
+            let final_balance = global_context
+                .database
+                .get_ft_balance(&contract_identifier, &token_name, recipient, Some(&metadata))
+                .map_err(|e| format!("failed to get_ft_balance for {contract_identifier}.{token_name} {recipient}: {e:?}"))?;
```
hugo-stacks, line 857: So the direct interaction with the DB _is_ needed, right?

components/clarity-repl/src/repl/session.rs
```diff
@@ -1199,6 +1208,66 @@ impl Session {
...
+            contract_identifier,
+            asset_name: asset_name.into(),
+        })
+    }
+
+    fn mint_ft(&mut self, command: &str) -> String {
+        let args: Vec<_> = command.split(' ').collect();
+
+        if args.len() != 4 {
+            return "Usage: ::mint_ft <recipient address> <assert identifier> <amount>"
```
hugo-stacks: I feel like `<asset identifier> <recipient> <amount>` would be more natural, what do you think?

also `assert` -> `asset`

components/clarity-repl/src/repl/session.rs
```diff
@@ -1199,6 +1208,66 @@ impl Session {
...
+            return Err(AssetIdentifierParseError::ContractIdentifierParseError);
+        };
+        let Some(asset_name) = asset_name_str.strip_prefix('.') else {
+            return Err(AssetIdentifierParseError::NoPrefixPeriod);
+        };
+        Ok(AssetIdentifier {
+            contract_identifier,
+            asset_name: asset_name.into(),
+        })
+    }
```
hugo-stacks: In many places in the REPL, we support the short syntax for contract ID so that this:
```
ST1PQHQKV0RJXZFY1DGX8MNSNYVE3VGZJSRTPGZGM.my-contract
```

can be written like that `.my-contract` when called with the same address (typically, the default deployer).

So it would be nice to support

```
::mint_ft .mycontract.mytoken <...>
```

Does that make sense?

components/clarity-repl/src/repl/session.rs
```diff
@@ -201,6 +209,7 @@ impl Session {
             cmd if cmd.starts_with("::toggle_timings") => self.toggle_timings(),
 
             cmd if cmd.starts_with("::mint_stx") => self.mint_stx(cmd),
+            cmd if cmd.starts_with("::mint_ft") => self.mint_ft(cmd),
```
hugo-stacks, line 212: To be added to `display_help()`

components/clarity-repl/src/repl/session.rs
```diff
@@ -1966,4 +2035,108 @@ mod tests {
...
+            .expect("failed to parse QualifiedContractIdentifier from {contract_identifier_str}");
+
+        let asset_identifier = format!("{contract_identifier_str}.{token_name}");
+
+        let recipient = contract_identifier.issuer.to_string();
+
+        // more than 4 tokens
+        let cmd = format!("::mint_ft {recipient} {asset_identifier} 1000 foo");
+        let result = session.handle_command(&cmd);
+        println!("{result}");
```
hugo-stacks, line 2130: Do we need those prints here in the tests?

hugo-stacks (changes requested):
Looking good, and well tested!
Requested a few changes.

Also, I think that there should be instructions somewhere on the expect token identifier format (either in `::help` or when a wrong format is provided). Devs often get confused with that

components/clarity-repl/src/repl/interpreter.rs
xoloki-stacks: It appears to be needed.  When contracts run, they interact with the token balances in the clarity vm db, not the `repl`.  So if we want contracts to behave as expected we need to manipulate the vm db, as the existing `mint_stx` call does.

components/clarity-repl/src/repl/session.rs
xoloki-stacks: Good catch!  Fixed

components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -1173,6 +1235,115 @@ mod tests {
...
+                    "(define-private (test-transfer-1000)",
+                    "  (ft-transfer? ctb u1000 tx-sender (as-contract tx-sender)))",
+                    "(define-private (test-burn-1000)",
+                    "  (ft-burn? ctb u1000 tx-sender))",
+                    "(test-mint)",
+                    "(test-burn)",
+                    "(test-transfer)",
+                    "(test-transfer-1000)",
+                    "(test-burn-1000)",
+                ]
```
hugo-stacks: Thinking about that now, we've been using `indoc` more and more lately to have beautiful source code in tests, eg:
https://github.com/stx-labs/clarinet/blob/43a35feb753fcbaa08ee7b8959b86dd0c1731166/components/clarity-repl/src/repl/interpreter.rs#L2002-L2007

You can use it if you want

components/clarity-repl/src/repl/session.rs
xoloki-stacks: Done, including an explanation of `<asset identifier>` format

components/clarity-repl/src/repl/session.rs
xoloki-stacks: The prints are only output if the test fails, and if that happens I think it would be useful to see if any of the cases which were expected to fail actually succeeded.  `Session::handle_command` simply returns `String`, not `Result<String, String>` so there's no way other than printing the output to see what succeeded and what failed.

components/clarity-repl/src/repl/session.rs
xoloki-stacks: Makes sense.  I updated the code so we pass an optional default_sender to when parsing the asset identifier, and if it fails to parse the contract identifier and it was passed a default sender it tries again with that prepended.  I then added tests on both the parser and the session command execution.

Lmk if that's what you wanted.

components/clarity-repl/src/repl/interpreter.rs
xoloki-stacks: Done.  It looks so much nicer!

components/clarity-repl/src/repl/session.rs
```diff
@@ -1199,6 +1212,91 @@ impl Session {
...
+                    } else {
+                        return Err(AssetIdentifierParseError::ContractIdentifierParseError);
+                    }
+                } else {
+                    return Err(AssetIdentifierParseError::ContractIdentifierParseError);
+                }
+            } else {
+                return Err(AssetIdentifierParseError::ContractIdentifierParseError);
+            }
+        };
```
hugo-stacks: Can you use `Self::desugar_contract_id` instead?

hugo-stacks (changes requested):
One minor request

components/clarity-repl/src/repl/session.rs
xoloki-stacks: I'm not sure that `desugar_contract_id` is behaving as expected.  A sugared contract_id should be in the form `.contract`, but if you pass that to `desugar_contract_id` it will split on the `.` which gives two parts, not one.  So it will not prepend the `deployer`, and `QualifiedContractIdentifier::parse` will fail.

Am I missing something here?

hugo-stacks: Ah, you're right.
That's because `desugar_contract_id` is mostly used in the JS SDK, that enabled
```ts
simnet.callPublicFn("my-contract", ...)
// and not `simnet.callPublicFn(".my-contract", ...) - with the dot
```

That's a bit confusing then 🤔 
I'd like to avoid the duplication of this `{deployer}.{contract_name}` logic, maybe it would be better to support both with the leading dot and without in `desugar_contract_id` if that's doesn't make it too confusing.

What do you think?

--- 

This concept of "sugar contract id" is trying to mimic the clarity syntax where you can do this if contract-1 and 2 have the same deployer
```clar
;; contract-2
(contract-call? .contract-1 ...)
```

xoloki-stacks: I think I can support both in `desugar_contract_id`.  I'll try that now.  I agree that we shouldn't duplicate this logic, and it's especially bad if we do so in divergent ways.

xoloki-stacks: Ok, seems to be working now

components/clarity-repl/src/repl/session.rs
```diff
@@ -1199,6 +1216,71 @@ impl Session {
         }
     }
 
+    fn parse_asset_identifier(
+        identifier: &str,
+        deployer: &str,
+    ) -> Result<AssetIdentifier, AssetIdentifierParseError> {
```
hugo-stacks, line 1220: Nitpicking here, but we may want to inverse the order of the arguments here, since this would make more sense to me
```rs
parse_asset_identifier("ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG", ".contract.asset")
```
vs
```rs
parse_asset_identifier(".contract.asset", "ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG")
```

Additionally, i'd rename `deployer` to `default_deployer`, both here and in `desugar_contract_id`, to make it more explicit that is not necessarily the deployer of the contract/asset, but the one to be used if needed.

Does that make sense?

hugo-stacks (changes requested):
- Need to update formatting (`cargo fmt-stacks`) for the CI
- Requested one last change

components/clarity-repl/src/repl/session.rs
xoloki-stacks: Swapping the order makes sense, since that's the order the args are in the string, though it does seem a little strange to have the optional deployer argument at the beginning.  And it's good to have the same order for both `desugar_contract_id` and `parse_asset_identifier`.  

I don't agree though about renaming `deployer` to `default_deployer`.  My understanding is that this does need to be the deployer of the contract/asset, or else the asset calls won't work on the clarity backend.  We may in fact be using what clarinet calls the default deployer, but only if we actually did use the default when deploying.

components/clarity-repl/src/repl/session.rs
```diff
@@ -248,12 +248,10 @@ impl Session {
         let is_qualified = parts_count == 2 && &contract[0..1] != ".";
         let contract_id = if is_qualified {
             contract.to_string()
+        } else if &contract[0..1] != "." {
+            format!("{}.{}", deployer, contract,)
```
hugo-stacks: 👌

components/clarity-repl/src/repl/session.rs
hugo-stacks: Well it's kind of optional, but in practice, it's always called with a value, since `self.get_tx_sender()` will always return something.

> My understanding is that this does need to be the deployer of the contract/asset

It doesn't if you call it with a fully qualified identifier.

So you can do
```rs
let caller = "ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG";

parse_asset_identifier(caller, ".contract.asset")
// resolve `ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG.contract.asset
parse_asset_identifier(caller, "STNHKEPYEPJ8ET55ZZ0M5A34J0R3N5FM2CMMMAZ6.contract.asset")
// resolve `STNHKEPYEPJ8ET55ZZ0M5A34J0R3N5FM2CMMMAZ6.contract.asset
```

This is just to provide a shorthand syntax to user, since the often just use to default deployer.
Does that make sense?

xoloki-stacks: Oh, I see what you're saying: we will always call it with the default deployer in the clarinet repl, but it will only be used if we passed a sugared contract identifier.  So in this context it is more correct to say default_deployer.

hugo-stacks (dismissed):
lgtm 🎉

components/clarity-repl/src/repl/session.rs
xoloki-stacks: Fixed

PR #2026, opened 2025-10-22
chore: stacks-labs migrations

hugo-stacks:
Let's merge #2049 with the packages renaming first.
This will allow us to set up trusted publishers on NPM

- update website, docs, and repositories URL
- update CI to handle stacks-labs permissions

PR #2034, opened 2025-10-27
ci: fix cache

hugo-stacks:
### Description

- `~/.cargo/bin/` is platform specific and should **not** be set in the global cache (see #2027)
- removev to obsoloete files

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -201,7 +201,7 @@ jobs:
           target-cache-key: ${{ matrix.os }}-${{ matrix.target }}-${{ github.job }}
 
       - name: Install wix (Windows)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -201,7 +201,7 @@ jobs:
           target-cache-key: ${{ matrix.os }}-${{ matrix.target }}-${{ github.job }}
 
       - name: Install wix (Windows)
-        if: matrix.os == 'windows-latest' && steps.setup-rust-cache.outputs.cargo-cache-hit != 'true'
+        if: matrix.os == 'windows-latest'
```
hugo-stacks, line 204: Slightly less optimized, but `wix` will be removed with #2026 anyway

PR #2035, opened 2025-10-27
Add `mintFT` to SDK

### LINKED ISSUE #1989: Allow to mint any kind of SIP 010 tokens in Simnet

xoloki-stacks:
### Description

This PR fixes #1989 by adding the ability to mint a fungible token to the SDK.

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -1133,8 +1133,8 @@ impl SDK {
     #[wasm_bindgen(js_name=mintFT)]
     pub fn mint_ft(
         &mut self,
-        recipient: String,
         token: String,
+        recipient: String,
         amount: u64,
     ) -> Result<String, String> {
         if PrincipalData::parse(&recipient).is_err() {
--- a/components/clarinet-sdk/node/tests/simnet-usage.test.ts
+++ b/components/clarinet-sdk/node/tests/simnet-usage.test.ts
@@ -522,6 +522,29 @@ describe("simnet can transfer stx", () => {
   });
 });
 
+describe("simnet can mint stx", () => {
+  it("can mint stx", () => {
+    simnet.mintSTX(address1, BigInt(1000));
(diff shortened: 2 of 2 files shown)


components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -1130,6 +1130,34 @@ impl SDK {
...
+        amount: u64,
+    ) -> Result<String, String> {
+        if PrincipalData::parse(&recipient).is_err() {
+            return Err(format!("Invalid recipient address '{recipient}'."));
+        }
+
+        let session = self.get_session_mut();
+
+        let default_deployer = session.get_tx_sender();
+        let asset_identifier = match Session::parse_asset_identifier(&default_deployer, &token) {
```
hugo-stacks: What about `.map_err()?;` instead of match here?

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -1130,6 +1130,34 @@ impl SDK {
...
+            Ok(asset_identifier) => asset_identifier,
+            Err(err) => return Err(format!("Invalid asset identifier '{token}': {err}")),
```
hugo-stacks: I'm afraid that this error messages might be a bit hard to understand for users if it just says `"Invalid [...]: no period"`.

Showing the expected format might be useful as well

hugo-stacks (changes requested):

components/clarinet-sdk-wasm/src/core.rs
xoloki-stacks: Done

components/clarinet-sdk-wasm/src/core.rs
xoloki-stacks: Done

PR #2039, opened 2025-10-29
fix: trick cost tracking into using deployed costs contracts

brady-stacks:
By default the costs are computed via [compiled cost function references](https://github.com/stacks-network/stacks-core/blob/master/clarity/src/vm/costs/mod.rs#L1158-L1173) so in order to actually use the custom boot contracts for costs, we need to trick clarity. We do this by hijacking the state_summary and adding our references to cost functions from the dummy contract we deploy if overrides are present.

This successfully alters the result of `::get_costs` in the repl, but the tests seemingly need an extra step similar to this as well..

### DIFF
--- a/components/clarity-repl/src/repl/boot/mod.rs
+++ b/components/clarity-repl/src/repl/boot/mod.rs
@@ -216,6 +216,10 @@ pub fn get_boot_contracts_data_with_overrides(
             continue;
         }
 
+        eprintln!(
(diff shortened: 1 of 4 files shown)


brady-stacks:
Closing this horrible hack and ignoring this for now 
Screenshot 2025-10-30 at 11 10 52 AM

PR #2040, opened 2025-10-30
fix: use clarity_version in ast visitor

hugo-stacks:
### Description

- add `fn get_clarity_version()` to the ASTVisitor trait.
  - add it to all ASTVisitor implementation (that's 80% of the PR)
- fix `traverse_expr` so that it uses the actual clarity version
- fix `match_tuple` so that it uses the actual clarity version

### DIFF
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -1130,6 +1130,32 @@ impl SDK {
         )
     }
 
+    #[wasm_bindgen(js_name=mintFT)]
+    pub fn mint_ft(
+        &mut self,
+        token: String,
+        recipient: String,
+        amount: u64,
+    ) -> Result<String, String> {
+        if PrincipalData::parse(&recipient).is_err() {
+            return Err(format!("Invalid recipient address '{recipient}'."));
+        }
+
+        let session = self.get_session_mut();
+
+        let default_deployer = session.get_tx_sender();
+        let asset_identifier = Session::parse_asset_identifier(&default_deployer, &token)
+            .map_err(|e| format!("Expected 'contract_id.asset_name', got '{token}': {e}"))?;
+
+        session.interpreter.mint_ft_balance(
+            &asset_identifier,
+            &PrincipalData::Standard(StandardPrincipalData::from(
+                StacksAddress::from_string(&recipient).unwrap(),
+            )),
+            amount,
+        )
+    }
+
     #[wasm_bindgen(js_name=setCurrentTestName)]
     pub fn set_current_test_name(&mut self, test_name: String) {
         let session = self.get_session_mut();
--- a/components/clarinet-sdk/node/tests/simnet-usage.test.ts
+++ b/components/clarinet-sdk/node/tests/simnet-usage.test.ts
@@ -522,6 +522,29 @@ describe("simnet can transfer stx", () => {
   });
 });
 
+describe("simnet can mint stx", () => {
+  it("can mint stx", () => {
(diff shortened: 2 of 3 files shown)


components/clarity-events/src/analysis/mod.rs
```diff
@@ -9,14 +9,15 @@ use clarity_repl::clarity::vm::types::{
...
 use clarity_repl::repl::clarity_values::value_to_string;
 use serde::ser::SerializeMap;
 use serde::{Serialize, Serializer};
 use serde_json::Value as JsonValue;
 
-#[derive(Default)]
-pub struct Settings {}
+pub struct Settings {
+    pub clarity_version: ClarityVersion,
+}
```
xoloki-stacks, line 21: You did a manual `impl Default for ContractDefinedData` below.  Why not do so here, using `DEFAULT_CLARITY_VERSION`?

components/clarity-lsp/src/common/requests/definitions.rs
```diff
@@ -22,19 +23,23 @@ pub enum DefinitionLocation {
...
 // when a user-defined keyword is used in the code, its position and definition location are stored in `tokens`
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 pub struct Definitions {
     pub tokens: HashMap<(u32, u32), DefinitionLocation>,
+    clarity_version: ClarityVersion,
     global: HashMap<ClarityName, Range>,
     local: HashMap<u64, HashMap<ClarityName, Range>>,
     deployer: Option<StandardPrincipalData>,
 }
```
xoloki-stacks, line 34: Again, why not `impl Default` using `DEFAULT_CLARITY_VERSION`?

components/clarity-lsp/src/common/requests/document_symbols.rs
```diff
@@ -553,7 +560,7 @@ mod tests {
 
     fn get_symbols(source: &str) -> Vec<DocumentSymbol> {
         let expr = get_ast(source);
-        let ast_symbols = ASTSymbols::new();
+        let ast_symbols = ASTSymbols::new(ClarityVersion::Clarity1);
```
xoloki-stacks, line 563: Why default to `ClarityVersion::Clarity1`?

components/clarity-events/src/analysis/mod.rs
hugo-stacks: Mostly because Settings::default() wouldn't be called

components/clarity-lsp/src/common/requests/definitions.rs
hugo-stacks: It would only be used to do

```rust
pub fn new(clarity_version: ClarityVersion, deployer: Option<StandardPrincipalData>) -> Self {
        Self {
            deployer,
            clarity_version,
            ..Default::default()
        }
    }
```

But the clarity_version default value would actually never be used

components/clarity-lsp/src/common/requests/document_symbols.rs
hugo-stacks: This is in the tests.
The `get_ast` function defined just above uses Clarity1

xoloki-stacks:
> @xoloki-stacks Now I'm actually second-guessing if you should have implemented default ContractDefinedData. In all of these scenarios, we never want to use a default clarity version. The actual value is actually specified.
> 
> What do you think?

If we never want to use a default clarity version, we definitely shouldn't `impl Default` with one for any struct @hugo-stacks

hugo-stacks:
Updated @xoloki-stacks 
Good catch

PR #2038, opened 2025-10-29
feat: upgrade clarity-vm with clarity 4 support - part 2

hugo-stacks:
### Description

Follow up of #2013

- upgrade clarity vm
- set basic support for new features

Related issue regarding the `todo` comment: https://github.com/stx-labs/clarinet/issues/2036

PR #2044, opened 2025-10-31
feat: restrict-assets? format

### LINKED ISSUE #2036: Clarity 4 support in formatter
The formatter will need to properly format the new `restrict-assets?` functions.<br>See the spec of this function in the [SIP](https://github.com/stacksgov/sips/blob/c2e7310ac34f30a8fa33b6805e705b2368a58cad/sips/sip-033/sip-033-clarity4.md#limiting-asset-access-restrict-assets).<br>

## Implementation details

The signature is:

```clar
(restrict-assets? asset-owner ((with-stx|with-ft|with-nft|with-stacking)*) expr-body1 expr-body2 ... expr-body-last)
```

We may way to format it like so:

* If `asset-owners` followed by the `allowances` (the `with-*`) methods fit one line

```clar
(restrict-assets? asset-owner ((with-stx u10000))
  expr-body1
  expr-body2
  ...
  expr-body-last
)
```

* If there are multiple allowances, put each one line, with an extra indentation (as for `let`)

```clar
(restrict-assets? asset-owner (
    (with-stx u10000)
    (with-ft (contract-of token-trait) "stackaroo" u50)
  )
  expr-body1
  expr-body2
  ...
  expr-body-last
)
```

brady-stacks:
- closes #2036 

follow up to #2038

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -241,6 +241,7 @@ impl<'a> Aggregator<'a> {
 
                             NativeFunctions::ListCons => self.format_list(list, previous_indentation),
 
+                            | NativeFunctions::RestrictAssets => self.format_restrict_assets(list, previous_indentation),
                             // everything else that's not special cased
                             NativeFunctions::Add
                             | NativeFunctions::Subtract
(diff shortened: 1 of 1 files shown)


hugo-stacks (changes requested):
@brady-stacks 

The body needs formatting as well
```clar
(restrict-assets? tx-sender
  (
    (with-stx u10000)
    (with-ft (contract-of token-trait) "stackaroo" u50)
  )
  (print u1) (print u1) (print u1) (print u1) (ok true)
)
```

I guess it basically need to be handle like the `let` body

brady-stacks:
@hugo-stacks that code is formatted currently like this:

```
            (restrict-assets? tx-sender (
                (with-stx u10000)
                (with-ft (contract-of token-trait) "stackaroo" u50)
              )
              (print u1)
              (print u1)
              (print u1)
              (ok true)
            )"#
```
Seems correct to me. Only difference being the argument opening paren. open to changing that

PR #2045, opened 2025-11-02
fix:: empty trait format bug

### LINKED ISSUE #2008: Minor formatting issue with empty trait definition
**Describe the bug**
For this Clarity code:
```clarity
(define-trait my-trait ())
```

The formatter adds a bunch of unnecessary whitespace:
```clarity
(define-trait my-trait (
  
))
```

**To Reproduce**
Steps to reproduce the behavior, create that contract, then format it using the LSP in VSCode.

**Expected behavior**
The original formatting should be left as-is.

BiliqisO:
### Description

This PR fixes #2008 - Empty trait definitions no longer add unnecessary whitespace.

 
  - I added an early-exit check in the define_trait function (formatter/mod.rs:460-464) to detect
  empty method lists and format them inline without adding whitespace.

  - I added the test_empty_trait test case to verify the fix

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -2394,6 +2394,7 @@ mod tests_formatter {
               () ;; empty
             )
             "#
+        );
         let result = format_with_default(src);
         assert_eq!(src, result);
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,38 @@
+# [3.9.0](https://github.com/stx-labs/clarinet/compare/v3.8.1...v3.9.0) (2025-11-05)
+
+This release has full support for Epoch 3.3 and Clarity 4. It still defaults to Epoch 3.2 and
+Clarity 3. A version 3.10.0 will be released shortly to default the the newer versions.
+
+The Clarinet JS SDK packages are now publised in the `@stacks` GitHub organization. Clarinet
+projects will need to migrate the project to use:
+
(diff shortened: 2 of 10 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -453,11 +453,20 @@ impl<'a> Aggregator<'a> {
         // name
         acc.push_str(&self.format_source_exprs(slice::from_ref(&exprs[1]), previous_indentation));
         acc.push_str(" (");
+
+        // methods
+        let methods = exprs[2].match_list().unwrap();
+
+        // if there are no methods, keep it on one line (no whitespace)
+        if methods.is_empty() {
+            acc.push_str("))");
```
brady-stacks: this should also handle the case where the methods are followed by comments
```
(define-trait my-trait 
  () ;; empty
)
```

brady-stacks:
Looks good @BiliqisO. If you can add a test for the case I mentioned we can merge it 👍

components/clarinet-format/src/formatter/mod.rs
BiliqisO: Actually, this fix doesn't account for comments (my bad). I'll make a correction on that now 🙏🏻

brady-stacks:
Sorry @BiliqisO I tried to fix the conflicts in the gh editor and failed. If you can fix those we'll get it merged asap

BiliqisO:
> Sorry @BiliqisO I tried to fix the conflicts in the gh editor and failed. If you can fix those we'll get it merged asap

Okay, I'll do that now

PR #2047, opened 2025-11-03
fix: publish -> deployments in completions

brady-stacks:
fixes outdated "publish" references


brady-stacks:
Created an issue label good-first-issue [here](https://github.com/stx-labs/clarinet/issues/2050)

PR #2051, opened 2025-11-04
fix: formatter string escapes

### LINKED ISSUE #2043: Formatter incorrectly escapes `'`
**Describe the bug**

If your contract contains a `'` in a string, the formatter will try to escape it, which is then not legal Clarity. For example the contract:

```clarity
"'"
```

becomes

```clarity
"\'"
```

brady-stacks:
closes #2043 

The issue didn't mention utf8 strings, but while handling the cases for `SequenceData::String` I noticed we also were mangling `CharType::UTF8` with the `to_string()`
This uses source extraction to skip dealing with the unicode codepoints at all

### DIFF
--- a/components/clarinet-format/src/formatter/ignored.rs
+++ b/components/clarinet-format/src/formatter/ignored.rs
@@ -4,12 +4,12 @@ use clarity::vm::representations::PreSymbolicExpression;
 /// This is useful for strings where we don't want to touch the formatting
 pub fn extract_expr_source(expr: &PreSymbolicExpression, source: &str) -> String {
     let span = &expr.span;
-    let start_line = usize::try_from(span.start_line).unwrap_or(0);
-    let start_column = usize::try_from(span.start_column).unwrap_or(0);
+    let start_line = usize::try_from(span.start_line).unwrap();
+    let start_column = usize::try_from(span.start_column).unwrap();
 
     if let Some(line) = source.lines().nth(start_line - 1) {
         let start_col = (start_column).saturating_sub(1);
-        let end_col = usize::try_from(span.end_column).unwrap_or(0);
+        let end_col = usize::try_from(span.end_column).unwrap();
 
         if start_col < line.len() && end_col <= line.len() {
             return line[start_col..end_col].to_string();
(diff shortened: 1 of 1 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1188,7 +1188,33 @@ impl<'a> Aggregator<'a> {
...
                     format!("'{c}")
                 }
-                // Fill in these explicitly
+                clarity::vm::types::Value::Sequence(clarity::vm::types::SequenceData::String(
+                    ref string_data,
+                )) => {
+                    match string_data {
+                        clarity::vm::types::CharType::ASCII(ascii_data) => {
+                            let content = String::from_utf8_lossy(&ascii_data.data);
+                            format!("\"{}\"", content)
```
jbencin-stacks: I thought Clippy made you inline these:

```suggestion
                            format!("\"{content}\"")
```

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1188,7 +1188,33 @@ impl<'a> Aggregator<'a> {
...
+                            if let Some(source) = self.source {
+                                let extracted = extract_expr_source(pse, source);
+                                if !extracted.is_empty() {
+                                    extracted
+                                } else {
+                                    value.to_string()
+                                }
+                            } else {
+                                value.to_string()
+                            }
```
jbencin-stacks: Can be simplified:

```suggestion
                            self.source
                                .map(|source| extract_expr_source(pse, source))
                                .filter(|extracted| !extracted.is_empty())
                                .unwrap_or_else(|| value.to_string())
```

components/clarinet-format/src/formatter/ignored.rs
```diff
@@ -1,5 +1,23 @@
 use clarity::vm::representations::PreSymbolicExpression;
 
+/// Extract the original source text using its span
+/// This is useful for strings where we don't want to touch the formatting
+pub fn extract_expr_source(expr: &PreSymbolicExpression, source: &str) -> String {
+    let span = &expr.span;
+    let start_line = span.start_line as usize;
```
jbencin-stacks: In general I try to avoid using `as` because it can silently truncate. Not going to be an issue going from `u32` to `usize` on any modern platform, but still better to use `usize::try_from()` or just make everything a `usize` or `u32` to avoid the conversion

brady-stacks: the u32 is from the PreSymbolicExpression span so we can't change that, but I can use the `try_from`

jbencin-stacks: I think it's better to panic here (just use `unwrap()`). This condition shouldn't happen, but if it does it shouldn't fail silently

brady-stacks: there were a couple other usize casts I changed as well 👍

PR #2052, opened 2025-11-05
Implement support for encrypted mnemonics

xoloki-stacks:
### Description

This PR adds the ability to encrypt a mnemonic via `clarinet deployments encrypt`, which the user can then store in the config file as `encrypted_mnemonic` where they would currently store a `mnemonic`.  Then when calling `clarinet deployments apply`, if an `encrypted_mnemonic` is present the user will be asked for a password, and the mnemonic will be decrypted and used during deployment.

`encrypt`/`decrypt` functions have been pulled from `wsts` into `clarinet-utils`; the `argon2` crate is now used for a  password-appropriate key derivation function.  Tests have been added to `clarinet-utils` for the new functionality, though not all error paths could be exercised.  

I didn't see an obvious way to directly test the new `clarinet deployments encrypt` functionality.  The top level `clarinet-cli/tests` directory did have some ability to run `clarinet console` commands while manipulating the input stream.  I could probably do something similar to run `clarinet deployments encrypt` followed by `clarinet deployments apply`, but the test fixtures are not designed to do such so it would be a massive change.  

#### Breaking change?

No breaking change.  However, if both `mnemonic` and `encrypted_mnemonic` are present, the `mnemonic` will be ignored.

### Checklist

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ members = [
 default-members = ["components/clarinet-cli"]
 
 [workspace.package]
-version = "3.9.2"
+version = "3.10.0"
 
 [workspace.dependencies]
 clarity = { git = "https://github.com/stacks-network/stacks-core.git", branch="develop", package = "clarity", default-features = false }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,17 @@
+# [3.10.0](https://github.com/stx-labs/clarinet/compare/v3.9.2...v3.10.0) (2025-11-13)
+
+##### New Features
+
+* Default to epoch 3.3 (#2063) (8aa982ee)
+* **Linter:**
+  *  Add `unused_map` lint (#2077) (a2e2e70f)
+  *  Add `unused_data_var` lint (#2070) (bd0a5939)
+  *  Add `unused_const` lint (#2066) (a36fc9a3)
+
+##### Chores
+
+*  New vscode extension logo (#2079) (d995f7cc)
+
 # [3.9.2](https://github.com/stx-labs/clarinet/compare/v3.9.1...v3.9.2) (2025-11-10)
 
 ##### Continuous Integration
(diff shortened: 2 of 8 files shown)


Cargo.toml
```diff
@@ -48,6 +51,7 @@ reqwest = { version = "0.12.22", default-features = false, features = [
     "json",
     "rustls-tls",
 ] }
+scanpw = "1.0.0"
```
hugo-stacks: Looks like `scanpw` brings a bunch of old dependencies (crossterm, mio, windows), could we maybe use something like https://github.com/conradkleinespel/rpassword instead?

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -873,22 +877,39 @@ pub fn main() {
...
+                );
+                let mut network_manifest = match res {
+                    Ok(network_manifest) => network_manifest,
+                    Err(e) => {
+                        let _ = event_tx.send(DeploymentEvent::Interrupted(e));
+                        return;
+                    }
+                };
+
+                // decrypt any encrypted mnemonics
```
hugo-stacks: This comment doesn't seem that helpful to me.
Or maybe it's just because there's an empty line below so it seems to be attached to nothing

hugo-stacks (changes requested):
Let me what you think about the scanpw dependency

Cargo.toml
xoloki-stacks: Done.  Significantly fewer dependencies as seen in `Cargo.lock`

components/clarinet-cli/src/frontend/cli.rs
xoloki-stacks: I removed it, the context below makes it extremely obvious what's going on in this block.

hugo-stacks (approved):
lgtm @xoloki-stacks 
Super cool 🙏

PR #2056, opened 2025-11-05
ci: upload release as draft

albu-stacks:
Changed the release action used from `ncipollo/release-action` to `softprops/action-gh-release`, also changed the release to be created as draft in order for the files to be uploaded correctly in the release.

Related issue: https://github.com/stx-labs/clarinet/issues/2054

### DIFF
--- a/.github/workflows/release-cli.yaml
+++ b/.github/workflows/release-cli.yaml
@@ -47,11 +47,11 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          password: ${{ github.token }}
 
       - name: Download release asset
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ github.token }}
           TAG: ${{ github.event.client_payload.tag || github.event.inputs.tag }}
         run: |
           gh release download $TAG --pattern "clarinet-linux-x64-glibc.tar.gz"


.github/workflows/ci.yaml
```diff
@@ -301,11 +301,19 @@ jobs:
         uses: actions/download-artifact@v4
 
       - name: Tag and Release
-        uses: ncipollo/release-action@v1
+        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2.2.1
+        env:
+          GITHUB_TOKEN: ${{ github.TOKEN }}
```
hugo-stacks: isn't it `${{ secrets.GITHUB_TOKEN }}`?


```suggestion
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

.github/workflows/ci.yaml
```diff
@@ -301,11 +301,19 @@ jobs:
...
         with:
-          artifacts: "**/*.tar.gz,**/*.exe"
-          tag: ${{ needs.get_release_info.outputs.tag }}
-          commit: ${{ env.GITHUB_SHA }}
+          name: "${{ needs.get_release_info.outputs.tag }}"
+          tag_name: ${{ needs.get_release_info.outputs.tag }}
+          draft: true
+          prerelease: true
```
hugo-stacks: What's the difference between draft and prelease?
Don't we just want `draft`?

.github/workflows/ci.yaml
```diff
@@ -301,11 +301,19 @@ jobs:
         uses: actions/download-artifact@v4
 
       - name: Tag and Release
-        uses: ncipollo/release-action@v1
+        uses: softprops/action-gh-release@c95fe1489396fe8a9eb87c0abf8aa5b2ef267fda # v2.2.1
```
hugo-stacks: Can we use the latest? (2.4.1)

.github/workflows/ci.yaml
albu-stacks: They functionally work the same, but with secrets you can also put a PAT in there - which is not needed for us: https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

On that note, I might have to allow the workflow some permissions to create the release

.github/workflows/ci.yaml
albu-stacks: Draft is visible only by repo maintainers - i.e. a "keep it private until I manually publish it", while prerelease is public marked as pre-release (experimental). When a draft prerelease is published, it will be just prerelease. I'll remove it, only leaving it as draft

.github/workflows/ci.yaml
albu-stacks: https://github.com/stx-labs/clarinet/pull/2056/commits/f80c1fd73c1a5e105a9c3a48fca53ad93167e827

.github/workflows/ci.yaml
albu-stacks: https://github.com/stx-labs/clarinet/pull/2056/commits/f80c1fd73c1a5e105a9c3a48fca53ad93167e827

.github/workflows/ci.yaml
albu-stacks: https://github.com/stx-labs/clarinet/pull/2056/commits/f80c1fd73c1a5e105a9c3a48fca53ad93167e827

.github/workflows/ci.yaml
```diff
@@ -18,6 +18,9 @@ concurrency:
   group: "${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.ref }}"
   cancel-in-progress: true
 
+permissions:
+  contents: write
```
hugo-stacks: the release job already has the permission:
https://github.com/stx-labs/clarinet/pull/2056/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR298

.github/workflows/ci.yaml
```diff
@@ -301,11 +304,18 @@ jobs:
         uses: actions/download-artifact@v4
 
       - name: Tag and Release
-        uses: ncipollo/release-action@v1
+        uses: softprops/action-gh-release@6da8fa9354ddfdc4aeace5fc48d7f679b5214090 # v2.4.1
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
```
hugo-stacks: Per the doc link you shared previously, isn't it `GH_TOKEN` instead `GITHUB_TOKEN`? For the key, not the value?

.github/workflows/ci.yaml
albu-stacks: Missed it, thank you for sharing! Removed in https://github.com/stx-labs/clarinet/pull/2056/commits/901c2c31617d114551d1d6c9a9f7ac98d1393f0b

.github/workflows/ci.yaml
albu-stacks: Hm actually not sure about this, as it seems to work with GITHUB_TOKEN on the stacks-core repo (from which this was taken). Will update it though, hope it works

albu-stacks: https://github.com/stx-labs/clarinet/pull/2056/commits/52d17673969411686f10b8431619b8c6fd758300

PR #2057, opened 2025-11-05
ci: update github token in docker image release job

hugo-stacks:
Use `github.token` for consistency with `github.TOKEN` with #2056

### DIFF
--- a/.github/workflows/release-cli.yaml
+++ b/.github/workflows/release-cli.yaml
@@ -47,7 +47,7 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ github.TOKEN }}
+          password: ${{ github.token }}
 
       - name: Download release asset
         env:


.github/workflows/release-cli.yaml
```diff
@@ -47,11 +47,11 @@ jobs:
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          password: ${{ github.TOKEN }}
```
albu-stacks:
```suggestion
          password: ${{ github.token }}
```

.github/workflows/release-cli.yaml
```diff
@@ -47,11 +47,11 @@ jobs:
...
 
       - name: Download release asset
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_TOKEN: ${{ github.TOKEN }}
```
albu-stacks:
```suggestion
          GH_TOKEN: ${{ github.token }}
```

albu-stacks (approved):
LGTM

PR #2059, opened 2025-11-05
ci: fix release dispatch

hugo-stacks:
### Description

With the changes of #2056, the CI now creates releases as Drafts.
Which triggered these release jobs during the 3.9.1 release.
We instead want to run it when the release is published


albu-stacks (approved):
LGTM

PR #2060, opened 2025-11-06
chore: update project schema with new epoch and clarity version

hugo-stacks:
### Description

Support for clarity 4 and epoch 3.3 in schema


brady-stacks:
Looks like we can close this 👌

PR #2063, opened 2025-11-06
feat: default to epoch 3.3

hugo-stacks:
### Description

Update:
- default epoch
- default clarity version
- default stacks-node and signer Docker images
- default config for the stacks node

### DIFF
--- a/components/clarity-repl/src/repl/remote_data/mod.rs
+++ b/components/clarity-repl/src/repl/remote_data/mod.rs
@@ -22,7 +22,7 @@ pub const MAINNET_25_START_HEIGHT: u32 = 147_290;
 pub const MAINNET_30_START_HEIGHT: u32 = 171_833;
 pub const MAINNET_31_START_HEIGHT: u32 = 340_555;
 pub const MAINNET_32_START_HEIGHT: u32 = 2_401_415;
-pub const MAINNET_33_START_HEIGHT: u32 = u32::MAX;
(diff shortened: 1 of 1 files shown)


components/clarinet-deployments/src/types.rs
```diff
@@ -1492,74 +1491,7 @@ mod tests {
...
-                [[burnchain.epochs]]
-                epoch_name = "3.2"
-                start_height = 10
-                "#
-            }
-        );
-    }
-
-    #[test]
-    fn test_epoch_config_with_optional_3_3() {
```
hugo-stacks, line 1550: I think this is a useful when we have an epoch transition (and we want to handle optional epoch setting in the stacks node config - depending in the image)

But also, we should better handle the epoch setting

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -14,16 +14,15 @@ use clarinet_deployments::{
 use clarinet_files::{
     FileAccessor, FileLocation, ProjectManifest, StacksNetwork, WASMFileSystemAccessor,
 };
+use clarity::types::{Address, StacksEpochId};
+use clarity::vm::{ClarityVersion, EvaluationResult, ExecutionResult, SymbolicExpression};
```
hugo-stacks, line 18: Slightly reworked the import for consistency

Issue #2028, opened 2025-10-23
Clarity Linter (updated)


jbencin-stacks:
Replaces #1792

### Description

Clarinet currently checks Clarity code for errors, but Clarity also  needs a full-featured linter, similar to `eslint` or `clippy`, which will flag code that is valid, but does not follow best practices or conventions

#### Config File

I plan to rename and refactor the current config file because as it is now, these names are too long and do not make it clear what the lint does. Example:

```toml
[repl.analysis]
passes = ['call_checker', 'check_checker', 'noop_checker', 'all']
```

The issues with this:
- Having a `_checker` suffix on every lint is extremely verbose
- Names like `check_checker` are not descriptive
- Enabling all lints should be specified a different way than enabling a single lint

My proposal:

```toml
[repl.analysis]
lints = ['arg_count', 'args_checked', 'noop']
```

or to enable all lints:

```toml
[repl.analysis]
lints = true
```

#### Planned Lints

Add a linter to the LSP with the possibility to easily add rules over time. The first few essential rules could include:

- Dead code
  - [ ] Unused function parameter
  - [ ] Unused private function
  - [ ] Unused `let` binding
  - [ ] Unused `define-constant`
  - [ ] Unused `use-trait`
  - [ ] Unused `define-data-var`
  - [ ] Unused `define-map`
  - [ ] Unused `define-fungible-token`
  - [ ] Unused `define-non-fungible-token`
  - [ ] Unnecessary `as-max-len?`
- Safety
  - [ ] Use of `unwrap-panic` 
  - [ ] Use of `unwrap-err-panic`
  - Prefer native function aliases (introduced in clarity 2 and versions to come)
    - [ ] Prefer `element-at?` (instead of `element-at`)
    - [ ] Prefer `index-of?` (instead of `index-of`)
    - [ ] Prefer bit-xor (instead of xor) (warning)
- Style/Convention
  - [ ] Invalid naming case
  - [ ] Multiple error constants with same value
- Misc
  - [x] Expression with no effect (no-op)
  - [ ] Public function that could be read-only

#### Configuration/Customization

- [ ] Allow prefixing identifier with `_` to disable "unused" warnings

#### Refactoring

I have a couple ideas here

- **Renaming and Moving Files:** Rename existing files and directories to match new names in config file, and move them all under single `lints` directory
[...]

PR #2066, opened 2025-11-08
feat(linter): Add `unused_const` lint

jbencin-stacks:
### Description

Add a single lint to linter (#2028)

Generate diagnostic messages for constants declared using `declare-constant` but never used

#### Breaking change?

No

### Example

```clarity
(define-constant MINUTES_PER_HOUR u60)

(define-read-only (hours-to-minutes (hours uint))
  (* hours u60))
```

```clarity
;; #[allow(unused_const)]
```

on the line above it

---

### Checklist

### DIFF
--- a/components/clarity-repl/src/analysis/mod.rs
+++ b/components/clarity-repl/src/analysis/mod.rs
@@ -101,11 +101,11 @@ impl TryFrom<String> for Lint {
 #[serde(rename_all = "snake_case")]
 pub enum LintLevel {
     #[default]
-    #[serde(alias = "allow", alias = "false", alias = "off", alias = "none")]
+    #[serde(alias = "allow", alias = "off", alias = "none")]
     Ignore,
     #[serde(alias = "note")]
     Notice,
-    #[serde(alias = "warn", alias = "true", alias = "on")]
+    #[serde(alias = "warn", alias = "on")]
     Warning,
     #[serde(alias = "err")]
     Error,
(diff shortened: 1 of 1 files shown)


components/clarity-repl/src/analysis/lints/noop/mod.rs
```diff
@@ -186,15 +187,17 @@ mod tests {
...
-(define-public (test-func)
-    (begin
-        (is-eq true)
-        (ok true)
-    )
-)
-"
-        .to_string();
+
+        #[rustfmt::skip]
```
brady-stacks, line 191: unfortunate 😢

jbencin-stacks: I really try not to use this, but if I don't it puts the `"`s on their own lines, which looks soooo bad

brady-stacks: Yeah, I've run into this often and it never fails to disappoint. Oh well

jbencin-stacks:
> The only thing I'm questioning is the rust format for ignoring

This syntax is not new, the check-checker and call-checker have used it probably for years (see `components/clarity-repl/src/analysis/annotation.rs`). But I agree that consistency would be good, we should talk with the rest of the team to decide which style we should standardize on

hugo-stacks:
> This syntax is not new, the check-checker and call-checker have used it probably for years [...] consistency would be good

We can definitely make breaking changes in this new linter spec (potentially keep backward compatibility with old annotations for a while).
But "check-checker" shouldn't hold us back from having a well-defined and well-thought spec

jbencin-stacks:
The existing `;; #[allow(...)]` syntax works well for the linter because there may be multiple violations per line. For example, a constant may not follow the naming convention, and also be unused. The user may only want to one or both of those warnings, and can specify which in the allow list

We should start discussion on unifying the syntax, but since `;; #[allow(...)]` is already established,  I don't want this to hold up the linter PRs

brady-stacks:
Yea, agreed. No need to hold it up, was just curious 👍

PR #2069, opened 2025-11-10
chore: release 3.9.2

hugo-stacks:
### Description

Clarinet 3.9.2 with many bug fixes.

Fixes new projects generation using Vitest v4


components/clarinet-sdk/node/package.json
```diff
@@ -68,6 +68,6 @@
     "@stacks/encryption": "^7.0.0",
     "@stacks/network": "^7.0.0",
     "@stacks/stacking": "^7.0.0",
-    "vitest": "^3.0.5"
+    "vitest": "^4.0.8"
```
hugo-stacks, line 71: This is not directly related to this release, but I figured that since clarinet project will now use vitest@4, we should be using it as well to run the sdk unit tests

PR #2070, opened 2025-11-11
feat(linter): Add `unused_data_var` lint

jbencin-stacks:
### Description

Add `unused_data_var` to the linter (#2028)

For a global variable defined using `define-data-var`:
- If never used at all, suggest user remove variable declaration
- If never written to with `var-set`, suggest user declare as constant instead
- If never read with `var-get`, suggest user remove variable declaration. In some cases this may be intentional, as variables can be read via API, but if so user should mark with `# [allow(unused_data_var)]` to ignore that specific variable

This PR also contains some refactoring, including a new `linter.rs` file and a new `LintPass` trait

#### Breaking change?

No

### Example

```clarity
(define-data-var counter uint u0)

(define-public (read-counter)
  (ok (var-get counter)))
```

```clarity
;; #[allow(unused_data_var)]
```

on the line above it

---

### Checklist


hugo-stacks (approved):
LGTM!

PR #2073, opened 2025-11-12
chore: upgrade clarity

hugo-stacks:
### Description

Upgrade clarity vm following this stacks-core PR: https://github.com/stacks-network/stacks-core/pull/6655

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -21,7 +21,7 @@
 
         # Hash for clarity git dependency - update this when Cargo.lock changes
         # Run `nix run .#check-git-dependencies-hash` to verify or get the new hash
-        clarityHash = "sha256-zVwVvLyh/hObWWKN6bIe4J5mf+6N8SyWZozzpKStcME=";
+        clarityHash = "sha256-QMkAqWyF3tJAJRkhvXEJnJiKvKd1OoAqwTZlpm5vojE=";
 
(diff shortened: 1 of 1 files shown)


components/clarity-repl/src/repl/datastore.rs
```diff
@@ -18,8 +18,9 @@ use clarity::vm::database::clarity_store::ContractCommitment;
 use clarity::vm::database::{
     BurnStateDB, ClarityBackingStore, ClarityDatabase, ClaritySerializable, HeadersDB, StoreType,
 };
-use clarity::vm::errors::{InterpreterError, InterpreterResult as Result};
+use clarity::vm::errors::InterpreterResult as Result;
```
brady-stacks, line 21: InterpreterResult doesn't exit  in `develop` anymore, right?
https://github.com/stacks-network/stacks-core/blob/develop/clarity/src/vm/errors.rs

hugo-stacks: Well, this is not exactly `develop`. This is `develop` two weeks ago, so this commit: `17a221c9ca051784c750c4b7aec9013e03280856`, as it is set in the Cargo.lock

brady-stacks: 👌

brady-stacks (approved):

PR #2077, opened 2025-11-12
feat(linter): Add `unused_map` lint

jbencin-stacks:
> [!IMPORTANT]
> This PR depends on #2070. Please review first!

### Description

Add `unused_map` to linter (#2028)

A map is considered used if the contract contains either of the following:
- A `map-insert` call
- Both a `map-set` and a `map-get?` call

#### Breaking change?

No

### Example

```clarity
(define-map admins principal bool)

(define-public (set-admin (p principal))
  (ok (map-set admins p true)))
```

```clarity
;; #[allow(unused_map)]
```

on the line above it

---

### Checklist

### DIFF
--- a/components/clarity-repl/src/analysis/lints/unused_map.rs
+++ b/components/clarity-repl/src/analysis/lints/unused_map.rs
@@ -173,7 +173,6 @@ impl<'a> UnusedMap<'a> {
         let mut diagnostics = vec![];
 
         for (name, data) in &self.maps {
-            // Map is unused, now decide which message/suggestion to show user
             let (message, suggestion) = match Self::compute_map_usage(data) {
                 Usage::Used => continue,
                 Usage::Unset => Self::make_diagnostic_strings_unset(name),


components/clarity-repl/src/analysis/lints/unused_map.rs
```diff
@@ -0,0 +1,407 @@
...
+    fn generate_diagnostics(&mut self) -> Vec<Diagnostic> {
+        let mut diagnostics = vec![];
+
+        for (name, data) in &self.maps {
+            if Self::is_map_used(data) {
+                continue;
+            }
+
+            // Map is unused, now decide which message/suggestion to show user
+            let (message, suggestion) = match (data.map_set, data.map_get) {
```
brady-stacks: This looks correct (so this is non-blocking), but I'm trying to imagine a "correct by construction" approach to this. Something like an enum `Unset | Unread | Unused`  that also removes the (true, true) case. I think you could also `case` on it in a generic `make_diagnostic_string` function rather than 3 different ones.

jbencin-stacks: Yeah I could move all this logic into `is_map_used()`, rename it to `get_map_usage()`, and have it return an enum. Would be a bit cleaner

jbencin-stacks: See if you like 198b669f better

PR #2079, opened 2025-11-13
chore: new vscode extension logo

hugo-stacks:
### Description

New vscode extension logo


jannik-stacks:
1074px is an interesting size, but lgtm

hugo-stacks:
Yeaahh... I don't know if we care about it 😅

PR #2084, opened 2025-11-18
Add flake.nix to enable Nix builds

adriano-stacks:
This PR also adds a nightly Nix CI build so that:

1. it is out of the way; but 
2. we will know if the Nix build breaks.

### Description

Once merged, people will be able to run Clarinet with:

```bash
nix run github:stx-labs/clarinet
```

### DIFF
--- a/.github/workflows/nix.yaml
+++ b/.github/workflows/nix.yaml
@@ -31,4 +31,5 @@ jobs:
       - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
       - uses: DeterminateSystems/determinate-nix-action@dbda91f6efef3ee627f56175120aa9543687d830  # v3.11.2
       - uses: DeterminateSystems/flake-checker-action@3164002371bc90729c68af0e24d5aacf20d7c9f6  # v12
+      - run: nix run .#check-git-dependencies-hash
       - run: nix flake check
--- a/flake.nix
+++ b/flake.nix
@@ -19,13 +19,17 @@
         pname = "clarinet";
         inherit (cargoToml.workspace.package) version;
 
+        # Hash for clarity git dependency - update this when Cargo.lock changes
+        # Run `nix run .#check-git-dependencies-hash` to verify or get the new hash
+        clarityHash = "sha256-zVwVvLyh/hObWWKN6bIe4J5mf+6N8SyWZozzpKStcME=";
+
         clarinet = pkgs.rustPlatform.buildRustPackage {
           inherit pname version;
           src = self;
           cargoLock = {
             lockFile = ./Cargo.lock;
             outputHashes = {
-              "clarity-0.0.1" = "sha256-zVwVvLyh/hObWWKN6bIe4J5mf+6N8SyWZozzpKStcME=";
+              "clarity-0.0.1" = clarityHash;
             };
           };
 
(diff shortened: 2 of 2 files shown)


flake.nix
```diff
@@ -0,0 +1,66 @@
...
+        pname = "clarinet";
+        inherit (cargoToml.workspace.package) version;
+
+        clarinet = pkgs.rustPlatform.buildRustPackage {
+          inherit pname version;
+          src = self;
+          cargoLock = {
+            lockFile = ./Cargo.lock;
+            outputHashes = {
+              "clarity-0.0.1" = "sha256-zVwVvLyh/hObWWKN6bIe4J5mf+6N8SyWZozzpKStcME=";
```
hugo-stacks: What's this 0.0.1 version?

flake.nix
```diff
@@ -0,0 +1,66 @@
...
+            };
+          };
+
+          preBuild =
+            let
+              versionsToml = pkgs.writeText "versions.toml" ''
+                stacks_node_version = "0.0.0.0.1"
+                stacks_signer_version = "0.0.0.0.1.0"
+              '';
```
hugo-stacks, line 41: What about this?

flake.nix
```diff
@@ -0,0 +1,66 @@
...
+            in
+            # Minimal versions.toml for stacks-common build script
+            # The build.rs in stacks-common expects it at ../versions.toml
+            # relative to the stacks-common crate in the cargo vendor directory
+            ''
+              cp ${versionsToml} "$NIX_BUILD_TOP/cargo-vendor-dir/versions.toml"
+            '';
+
```
hugo-stacks, line 49: copy-pasted stuff?

.github/workflows/nix.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - uses: DeterminateSystems/flake-checker-action@3164002371bc90729c68af0e24d5aacf20d7c9f6  # v12
+      - run: nix flake check
+
+  nix-darwin-ci:
+    runs-on: macos-26  # arm, beta
+    steps:
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+      - uses: DeterminateSystems/determinate-nix-action@dbda91f6efef3ee627f56175120aa9543687d830  # v3.11.2
+      - uses: DeterminateSystems/flake-checker-action@3164002371bc90729c68af0e24d5aacf20d7c9f6  # v12
+      - run: nix flake check
```
hugo-stacks, line 35: Could use a matrix here
```suggestion
jobs:
  nix-ci:
    strategy:
      matrix:
        os:
          - ubuntu-24.04-arm
          - macos-26
    runs-on: ${{ matrix.os }}
    steps:
      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
      - uses: DeterminateSystems/determinate-nix-action@dbda91f6efef3ee627f56175120aa9543687d830  # v3.11.2
      - uses: DeterminateSystems/flake-checker-action@3164002371bc90729c68af0e24d5aacf20d7c9f6  # v12
      - run: nix flake check
```

adriano-stacks:
(I don't think the CI failures are related to my changes)

adriano-stacks:
> @adriano-stacks To make sure I understand correctly, you specifically want `nix run github:stx-labs/clarinet` to run whatever is on `main`, rather than a tagged release? What's the use case?

You can actually run tagged versions or even specific commits, Nix allows that (see [here for the docs](https://nix.dev/manual/nix/2.18/command-ref/new-cli/nix3-flake.html#examples)). So running the currently newest release would be:

```bash
nix run github:stx-labs/clarinet/v3.10.0
```

> 
> Regarding
> 
> > it is out of the way
> 
> Do you expect it to be long-running or heavy? Otherwise, we could just run it in the regular CI or whenever a PR is merged

My thinking here is: maybe not a lot of people will use Nix. So I don't want devs to wait on CI to build Nix or fix it if broken. If it breaks, I can just fix it when I have cycles.

We can also make it a CI check for PRs, it should take roughly like a Rust build (since it's what it does).

flake.nix
adriano-stacks: It comes from here:

https://github.com/stx-labs/clarinet/blob/7ff149f8a83c8e4ef0510cef386108884b3e0849/Cargo.lock#L786

Which in turns comes from here:

https://github.com/stacks-network/stacks-core/blob/b3efa588571e168d2d154790e1c57d9e3e64adc5/clarity/Cargo.toml#L3

If you are wondering about the hash, is because the `cargo.lock` doesn't specify a checksum for git dependencies.

flake.nix
adriano-stacks: It's a hack to make `stacks-common` build in Nix. It requires [this file](https://github.com/stacks-network/stacks-core/blob/master/versions.toml) to exist on its parent directory, but because it's not part of the crate Nix doesn't automatically fetch it and the build breaks.

There are cleaner ways to do it e.g. overriding how that particular crate is built or fetching the file separately and making it available to Nix with the correct versions.

This _shouldn't_ be an issue and it's a quick fix that makes it work.

flake.nix
adriano-stacks: No it's actually required. The line `versionsToml = pkgs.writeText "versions.toml"` writes the file to the Nix store (Nix is kinda weird in that), this one copies it where cargo will find it.

flake.nix
hugo-stacks: Do you know why it cares about the `clarity` version? And not like the `stacks-common`, `pox-locking`, or `stackslibs` one? That also are git dependencies

And does it mean we'll have to manually update this file when we update the stacks-core dependencies version?

adriano-stacks: I think it's because they are all part of the same `git` workspace and so you pull them all together?

> And does it mean we'll have to manually update this file when we update the stacks-core dependencies version?

Unfortunately I think so, it wants know in advance the hash of what it will fetch. There might be ways around it since it's a git repo and we have the commit, but it might not be easy.

If we feel that's too much burden, I can also have the `flake.nix` file up somewhere outside this repo and update it every once in a while (eg for new releases).

hugo-stacks:
@adriano-stacks 

> My thinking here is: maybe not a lot of people will use Nix. So I don't want devs to wait on CI to build Nix or fix it if broken.  If it breaks, I can just fix it when I have cycles.

I worry that if the job fails every day in the middle of the night, I'm not sure who will be notified and will fix it, and I don't think it should rely on one person to fix it (you).

But indeed (also re: "_If we feel that's too much burden, I can also have the flake.nix file up somewhere outside this repo and update it every once in a while_"), if we do this in this repo, I'd rather make sure that we have first-class support for it, and that the team actually maintains it.

Which kind of means running the ci check in every PR

hugo-stacks:
We already build the cli 8 times and the sdk 6 times, so I think we can build it 2 more times

adriano-stacks:
Alright, https://github.com/stx-labs/clarinet/pull/2084/commits/6d9b961d6aced4ca2cbea1cea818a170bc6cedcf makes it a PR-triggered workflow. Be ruthless about it, if it ever slows you down, we can kill it. 

EDIT: I have just pushed [this](https://github.com/stx-labs/clarinet/pull/2084/commits/ae12f1b9ba07d7fa400a82a3d3422efa8dd11edf) so that when the lockfile changes the CI will just suggest which "hash" value to add to the flake.nix file.

hugo-stacks:
> I have just pushed [this](https://github.com/stx-labs/clarinet/pull/2084/commits/ae12f1b9ba07d7fa400a82a3d3422efa8dd11edf) so that when the lockfile changes the CI will just suggest which "hash" value to add to the flake.nix file

Awesome! It was the thing I was worried about regarding maintaining it

>  if it ever slows you down, we can kill it.

The job runs in 2min, that's perfect 👌

PR #2087, opened 2025-11-20
feat: add static code analysis CodeLens

brady-stacks:
- [x] CodeLens above functions showing static max cost

Screenshot 2026-03-02 at 10 55 24 AM

- [x] config toggle

Screenshot 2026-03-02 at 10 55 04 AM

depends on #2210 

Additional fixes:
- removes insights panel view (unused)
- adds a flag for checking exact values in dynamic vs static (the range test allowed wrong values for static fns)

### DIFF
--- a/components/clarinet-files/src/network_manifest.rs
+++ b/components/clarinet-files/src/network_manifest.rs
@@ -15,8 +15,9 @@ use crate::paths;
 
 pub const DEFAULT_DERIVATION_PATH: &str = "m/44'/5757'/0'/0/0";
 
-pub const DEFAULT_STACKS_NODE_IMAGE: &str = "blockstack/stacks-blockchain:3.3.0.0.5-alpine";
-pub const DEFAULT_STACKS_SIGNER_IMAGE: &str = "blockstack/stacks-signer:3.3.0.0.5.0-alpine";
+pub const DEFAULT_STACKS_NODE_IMAGE: &str = "ghcr.io/stacks-network/stacks-core:3.3.0.0.6-alpine";
+pub const DEFAULT_STACKS_SIGNER_IMAGE: &str =
+    "ghcr.io/stacks-network/stacks-signer:3.3.0.0.6.0-alpine";
 pub const DEFAULT_STACKS_API_IMAGE: &str = "hirosystems/stacks-blockchain-api:latest";
 
 pub const DEFAULT_POSTGRES_IMAGE: &str = "postgres:alpine";


brady-stacks:
dependent on https://github.com/stacks-network/stacks-core/pull/6704

components/clarity-lsp/src/common/backend.rs
```diff
@@ -576,6 +591,41 @@ pub fn process_request(
...
+                .try_read(|es| es.get_code_lenses(&contract_location))
+                .unwrap_or_default();
+            Ok(LspRequestResponse::CodeLens(code_lenses))
+        }
+
+        LspRequest::FunctionAnalysis(params) => {
+            eprintln!(
+                "[LSP] FunctionAnalysis request received: path={}, line={}, char={}",
+                params.path, params.line, params.char
+            );
```
hugo-stacks: To be removed?

components/clarity-lsp/src/common/requests/capabilities.rs
```diff
@@ -29,6 +30,7 @@ impl Default for InitializationOptions {
             go_to_definition: true,
             hover: true,
             signature_help: true,
+            static_cost_analysis: true,
```
hugo-stacks: Shouldn't it be off by default?

components/clarity-lsp/src/common/backend.rs
```diff
@@ -284,6 +284,17 @@ pub async fn process_notification(
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct FunctionAnalysisParams {
+    pub path: String,
+    pub line: u32,
+    pub char: u32,
+}
+
+impl FunctionAnalysisParams {
+    pub const METHOD: &'static str = "clarity/getFunctionAnalysis";
+}
```
hugo-stacks: Is this adding a new non-standard lsp request type?
Can you add a comment the explains why / how it works?

I dont get how they are sent by the client 🤔

components/clarity-vscode/client/src/common.ts
```diff
@@ -159,6 +160,10 @@ export async function initClient(
     }
   }
 
+  // Register no-op command for static cost code lens (display-only, no action on click)
+  context.subscriptions.push(
+    vscode.commands.registerCommand("clarity.staticCostLens", () => {}),
+  );
```
hugo-stacks: What does this do?

components/clarity-lsp/src/common/requests/capabilities.rs
brady-stacks: yes, The config has it defaulted, but this should be too 👍

hugo-stacks: Right, also because the default config is only for vscode, but regular lsp needs to reflect that

components/clarity-lsp/src/common/state.rs
```diff
@@ -431,6 +449,103 @@ impl EditorState {
...
+
+        if !self.settings.static_cost_analysis {
+            return code_lenses;
+        }
+
+        // Get the contract metadata to find the manifest
+        let contract_metadata = match self.contracts_lookup.get(contract_location) {
+            Some(md) => md,
+            None => return code_lenses,
+        };
```
hugo-stacks, line 455: The `match` here and below can be replaced with more idiomatic `let else`

components/clarity-lsp/src/common/state.rs
```diff
@@ -128,17 +130,20 @@ impl ActiveContractData {
...
     contract_calls: Vec<CompletionItem>,
     errors: Vec<ClarityDiagnostic>,
     warnings: Vec<ClarityDiagnostic>,
     notes: Vec<ClarityDiagnostic>,
+    #[allow(dead_code)]
     contract_id: QualifiedContractIdentifier,
     analysis: Option<ContractAnalysis>,
     definitions: HashMap<ClarityName, Range>,
+    #[allow(dead_code)]
     location: PathBuf,
```
hugo-stacks: Let's delete this properties in a follow up PR once this is merged #2256

components/clarity-lsp/src/utils/mod.rs
```diff
@@ -17,6 +17,21 @@ macro_rules! log {
...
+        #[cfg(target_arch = "wasm32")]
+        {
+            web_sys::console::log_1(&format!( $( $t )* ).into());
+        }
+        #[cfg(not(target_arch = "wasm32"))]
+        {
+            eprintln!( $( $t )* );
+        }
+    };
+}
```
hugo-stacks: I think in the lsp you can use `uprint!` from `clarity_repl`
If we want to keep those logs

https://github.com/brady-stacks/clarinet/blob/9f20d43e82fba38f51e910bd38dc8e86f22d4f18/components/clarity-repl/src/uprint.rs#L5

components/clarity-lsp/src/common/state.rs
```diff
@@ -705,7 +861,69 @@ pub async fn build_state(
...
+        "[LSP] get_cost_analysis called for contract: {} (clarity version: {:?})",
+        contract_id,
+        clarity_version
+    );
+
+    let tx_sender: clarity_types::types::PrincipalData = session.interpreter.get_tx_sender().into();
+
+    let mut global_context = session
+        .interpreter
+        .get_global_context(clarity_repl::clarity::StacksEpochId::Epoch21, false)
```
hugo-stacks: I'm not sure if it would make a difference, but do we need to pass the actual contract epoch here?

components/clarity-vscode/server/src/common.ts
```diff
@@ -31,9 +32,16 @@ export function initConnection(
...
+    if (!notification) return;
+    const [method] = notification;
     try {
       await bridge.onNotification(...notification);
+      if (
+        method === DidSaveTextDocumentNotification.method ||
+        method === DidOpenTextDocumentNotification.method
+      ) {
+        connection.sendRequest("workspace/codeLens/refresh").catch(() => {});
+      }
```
hugo-stacks: We want to have as little logic as possible in server/common.ts, because it means that it will only be enabled in vscode and not in other editors using the native lsp. This file should just handle requests and notifications routing, but not have any logic.

I think the `"workspace/codeLens/refresh"` should be sent from the Rust side, both in vscode_bridge and native_bridge

brady-stacks: Good call. Now sends from vscode bridge and native 👍

components/clarity-lsp/src/common/state.rs
```diff
@@ -129,17 +131,20 @@ impl ActiveContractData {
     }
 }
 
-#[derive(Debug, Clone, PartialEq)]
```
xoloki-stacks, line 132: Why remove `PartialEq`?

components/clarity-lsp/src/common/state.rs
```diff
@@ -129,17 +131,20 @@ impl ActiveContractData {
...
+#[derive(Debug, Clone)]
 pub struct ContractState {
     contract_calls: Vec<CompletionItem>,
     errors: Vec<ClarityDiagnostic>,
     warnings: Vec<ClarityDiagnostic>,
     notes: Vec<ClarityDiagnostic>,
+    #[allow(dead_code)]
```
xoloki-stacks: Why keep this (and the one below) if they are dead code?  Or is the compiler just being picky?

components/clarity-lsp/src/common/state.rs
brady-stacks: unused. I think it's related to the dead code. I just removed those too. I figured it was going to be too much changes for this already large PR, but it was only a few places

PR #2090, opened 2025-11-24
fix: shim InterpreterResult for future stacks-core

brady-stacks:
`InterpreterResult` alias was removed from stacks-core [here](https://github.com/stacks-network/stacks-core/commit/3ba954dd0b231adc1f6406f5ef8b4ec22e5768cb#diff-236aa49dfd8ede88d351e87f5df9ee9918a246a8e6a4a48136190385853503feR52-R78) so when we bump the version it would break. This shims it back in for clarinet


brady-stacks:
Ah, there's a bunch of others too.. @hugo-stacks We don't want to use develop in clarinet, but the static cost is using it, so we're sort of stuck. I could shim all of it, but it'll be messy. It's largely module moves and renames

PR #2098, opened 2025-12-08
Fix: update clarinet to parse new tenure extends

aaronb-stacks:
### Description

This lets Clarinet parse the new tenure extend variants.

It also updated a couple of imports which was necessary for getting the CLI to compile with the latest from `stacks-core:develop`.

One note on this patch is that it seems like clarinet's build would benefit from using specific git hashes rather than `develop` for the stacks-core dependencies -- otherwise a working build could fail when stacks-core gets updated (which seems to happen somewhat regularly?). I see an ~2 year old issue that is related to this (https://github.com/stx-labs/clarinet/issues/1346), so I'll comment there.

### DIFF
--- a/components/clarity-repl/src/repl/remote_data/mod.rs
+++ b/components/clarity-repl/src/repl/remote_data/mod.rs
@@ -37,8 +37,6 @@ pub const TESTNET_31_START_HEIGHT: u32 = 814;
 pub const TESTNET_32_START_HEIGHT: u32 = 3_140_887;
 pub const TESTNET_33_START_HEIGHT: u32 = 3_640_102;
 
-type InterpreterResult<T> = std::result::Result<T, VmExecutionError>;
-
 pub fn epoch_for_height(is_mainnet: bool, height: u32) -> StacksEpochId {
     if is_mainnet {
         epoch_for_mainnet_height(height)
--- a/components/clarity-repl/src/repl/datastore.rs
+++ b/components/clarity-repl/src/repl/datastore.rs
@@ -38,8 +38,6 @@ const CLARITY_STORAGE_BLOCK_TIME_KEY: &str = "_stx-data::clarity_storage::block_
 const SECONDS_BETWEEN_BURN_BLOCKS: u64 = 600;
 const SECONDS_BETWEEN_STACKS_BLOCKS: u64 = 10;
 
-type Result<T> = std::result::Result<T, VmExecutionError>;
-
 fn epoch_to_peer_version(epoch: StacksEpochId) -> u8 {
     use clarity::consts::*;
     match epoch {


components/clarity-repl/src/repl/remote_data/mod.rs
```diff
@@ -37,6 +37,8 @@ pub const TESTNET_31_START_HEIGHT: u32 = 814;
 pub const TESTNET_32_START_HEIGHT: u32 = 3_140_887;
 pub const TESTNET_33_START_HEIGHT: u32 = 3_640_102;
 
+type InterpreterResult<T> = std::result::Result<T, VmExecutionError>;
```
hugo-stacks: Why this change?
I understand we might need it when we upgrade `clarity`, but why do it in this PR that doesn't upgrade clarity?

aaronb-stacks: I might be building clarinet wrong (or at least, I might be building it differently than I should):

```
$ cargo install --path /home/dev/clarinet/components/clarinet-cli --bin clarinet
...
   Compiling clarity-repl v3.11.0 (/data/components/clarity-repl)
error[E0432]: unresolved import `clarity::vm::errors::InterpreterResult`
 --> components/clarity-repl/src/repl/remote_data/mod.rs:5:5
  |
5 | use clarity::vm::errors::InterpreterResult;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `InterpreterResult` in `vm::errors`

   Compiling stacks-rpc-client v3.11.0 (/data/components/stacks-rpc-client)
   Compiling chainhook-sdk v0.12.12 (/data/components/chainhook-sdk)
   Compiling stackslib v0.0.1 (https://github.com/stacks-network/stacks-core.git?branch=develop#cf8d6858)
For more information about this error, try `rustc --explain E0432`.
error: could not compile `clarity-repl` (lib) due to 1 previous error
warning: build failed, waiting for other jobs to finish...
error: failed to compile `clarinet-cli v3.11.0 (/data/components/clarinet-cli)`, intermediate artifacts can be found at `/home/dev/rust-target/`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

Basically, the build fetches the latest from `stacks-core:develop`, and then fails because `InterpreterResult` no longer exists in `develop`.

hugo-stacks: Ah, got it, you need `--locked`.

Or there's a cargo alias that does it: `cargo clarinet-install`.

But we'll switch back to setting the commit rev in the cargo.toml. Let's do it in #2099

hugo-stacks:
@aaronb-stacks #2099 was merged, so you now have conflicts, but I think they are in the two files that changed `InterpreterResult`.
So you should be good now after merging main

aaronb-stacks:
Okay, will resolve

components/clarity-repl/src/repl/remote_data/mod.rs
aaronb-stacks: Okay, removed

hugo-stacks (approved):
lgtm! Thanks @aaronb-stacks

hugo-stacks:
@aaronb-stacks Do you need it released or not necessarily? 
Just to know if we need to prioritize a release soon

aaronb-stacks:
Nope -- no need for a release, local builds worked perfectly for me.

PR #2103, opened 2025-12-09
Decrypt `encrypted_mnemonic` immediately when loading `NetworkManifest`

### LINKED ISSUE #2100: Generated default deployment has random expected-sender
**Describe the bug**
Generate deployment plan like this

`clarinet deployments generate --mainnet --low-cost`

results in random expected-sender when using encrypted mnemonic

**To Reproduce**
1. clarinet deployments encrypt
2. copy encrypted mnemonic to Mainnet.toml
3. clarinet deployments generate --mainnet --low-cost

**Expected behavior**
Ask for password and use corresponding account

- clarinet 3.11.0

xoloki-stacks:
### Description

In the initial `encrypted_mnemonic` work, we tried to preserve functional encapsulation by only accessing the terminal inside the `frontend` module.  But this lead to unexpected bugs due to the random mnemonics and stacks addresses that were generated with this approach.

So in this PR we simply decrypt the `encrypted_mnemonic` field while loading the `NetworkManifest`, and don't worry about accessing the terminal to read the password in a non-frontend module.

Fixes #2100 

#### Breaking change?

No

### Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -747,6 +747,7 @@ dependencies = [
  "dirs",
  "js-sys",
  "libsecp256k1",
+ "rpassword",
  "schemars 1.1.0",
  "serde",
  "serde-wasm-bindgen",


xoloki-stacks:
> Small quality of life improvement, could you change the `clarinet deployment encrypt` output so that it's a valid toml key/value and developers can directly copy/paste the whole line. 

https://github.com/stx-labs/clarinet/pull/2107

PR #2104, opened 2025-12-09
fix: list spacing consistency

### LINKED ISSUE #2097: Formatter: list has extra spaces
**Describe the bug**

When applying formatting to the function below it inserted a bunch of extra spaces.

**To Reproduce**

```clarity
(define-private (is-sorted-principals? (plist (list 20 principal)))
  (let ((n (len plist)))
    (if (> n u20)
      false
      (if (<= n u1)
        true ;; 0/1 trivial
        (get sorted
          (fold sorted-fold-step
            (list
              u0               u1               u2               u3
                            u4               u5               u6               u7
                            u8               u9
              u10               u11               u12               u13               u14
                            u15               u16               u17               u18
                            u19
            ) {
            sorted: true,
            plist: plist,
            n: n,
          })
        )
      )
    )
  )
)
```

**Expected behavior**

I expected the list to be formatted with spaces, e.g.

```clarity
        (fold sorted-fold-step
            (list u0 u1 u2 u3 u4 u5 u6 u7 u8 u9 u10 u11 u12 u13 u14 u15 u16 u17 u18 u19)
            {
              sorted: true,
              plist: plist,
              n: n,
            }
        )
```

 - Clarinet version [e.g. 1.7.0] - `clarinet 3.11.0`

brady-stacks:
closes #2097 

We had previously done some weird logic to preserve spacing in lists if it seemed intentional, which is nonsense in retrospect. This makes it consistent and adds some guard rails around line length as well as more accurately picking up whether a list is an expression `(list u1 u2)` or a type `(list 12 (buff 32))`

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -952,21 +952,17 @@ impl<'a> Aggregator<'a> {
             PreSymbolicExpressionType::AtomValue(ref value) => {
                 matches!(value, clarity::vm::types::Value::Int(_))
             }
-            PreSymbolicExpressionType::Atom(ref atom) => {
-                // Check if it's a plain signed number (not unsigned)
-                atom.as_str().parse::<i128>().is_ok()
-            }
             _ => false,
         }
     }
 
     /// Detect if this is a list type signature: (list <integer> <type>)
-    fn detect_list_type_signature(
-        &self,
-        exprs: &[PreSymbolicExpression],
-        is_list_cons: bool,
-    ) -> bool {
-        is_list_cons && exprs.len() == 3 && self.is_integer_expr(&exprs[1])
+    fn is_list_type_signature(&self, exprs: &[PreSymbolicExpression]) -> bool {
+        // the 1st item is a different type than the 2nd
+        exprs.len() >= 3
+            && exprs[0].match_atom() == Some(&clarity::vm::ClarityName::from("list"))
+            && self.is_integer_expr(&exprs[1])
+            && !self.is_integer_expr(&exprs[2])
     }
 
     /// Estimate the length of a list if formatted on a single line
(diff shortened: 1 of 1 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -955,39 +944,185 @@ impl<'a> Aggregator<'a> {
...
+    }
+
+    /// Detect if this is a list type signature: (list <integer> <type>)
+    fn detect_list_type_signature(
+        &self,
+        exprs: &[PreSymbolicExpression],
+        is_list_cons: bool,
+    ) -> bool {
+        is_list_cons && exprs.len() == 3 && self.is_integer_expr(&exprs[1])
+    }
```
hugo-stacks: The `is_integer_expr` check might not be robust enough. Since a `(list 10 int)` can be `(list 1 2)` and look like a type signature.
Don't know what are the consequences though. What is `is_list_cons`?

hugo-stacks (changes requested):
Looks pretty good, just had a question regarding a certain functions, but I don't know if that would really impact on the formatting

components/clarinet-format/src/formatter/mod.rs
brady-stacks: Yeah, you're right. We can just directly check it without that bool. I added a test and changed it to simply check if the 1st element is an int and the 2nd _isn't_

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -955,39 +944,184 @@ impl<'a> Aggregator<'a> {
...
+    /// Unsigned integers like u10 would be list elements, not type signatures
+    fn is_integer_expr(&self, expr: &PreSymbolicExpression) -> bool {
+        match &expr.pre_expr {
+            PreSymbolicExpressionType::AtomValue(ref value) => {
+                matches!(value, clarity::vm::types::Value::Int(_))
+            }
+            PreSymbolicExpressionType::Atom(ref atom) => {
+                // Check if it's a plain signed number (not unsigned)
+                atom.as_str().parse::<i128>().is_ok()
+            }
```
hugo-stacks: Can we really fall in to this case? Don't we expect an AtomValue?

Plus, I'm afraid that `atom.as_str().parse::<i128>().is_ok()` would match on a string like `"1"` (vs a number `1`)

brady-stacks: Yes! My fault. Removed, tests passing 👍

hugo-stacks (approved):
lgtm!

Issue #2106, opened 2025-12-11
Expose `generateDefaultDeploymentPlan()` in the JS SDK


hugo-stacks:
The JS SDK can already generate and save the default deployment plan, but they only way to call it is to actually start a new simnet session.
We should expose a function in the JS SDK that does just that, because other tools that rely on the SDK (clarigen, rendezvous) can extract valuable data from the deployment plan

PR #2109, opened 2025-12-12
feat: generate deployment plan in sdk

hugo-stacks:
fixes: #2106

Adds a new function `generate_deployment_plan()` that can be called in the JS SDK.

We should prioritise [this issue](https://github.com/stx-labs/clarinet/issues/1368) that would remove a lot of simnet custom deployment plan logic (the "merge deployment plan" logic) by making it the canonical behavior.

### DIFF
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -4,8 +4,8 @@ use std::path::PathBuf;
 
 use clarinet_deployments::diagnostic_digest::DiagnosticsDigest;
 use clarinet_deployments::types::{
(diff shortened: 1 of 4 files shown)


components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -393,74 +399,23 @@ impl SDK {
         Ok(())
     }
 
+    #[wasm_bindgen(js_name=clearCache)]
+    pub fn clear_cach(&mut self) {
```
brady-stacks, line 403: cache*

hugo-stacks: ah! I shouldn't have move this function 😓 
I'll change it in a follow up PR

PR #2110, opened 2025-12-13
fix: rework format-ignore

### LINKED ISSUE #2105: `@format-ignore` break formatting
The format-ignore directive seems to mess up the parent function:

input
```clar
(define-public (hello)
  (begin
    ;; @format-ignore
    (ok "world")
  )
)
```

`clarinet fmt --in-place` ouput:
```clar
(
  define-public   (hello)
  (
  begin
  ;; @format-ignore
  (ok "world")
)
)
```

brady-stacks:
closes #2105 

This is a somewhat heavy rework of format-ignore to cache ignored exprs within the main loop. I'm not sure it ever worked since we were only checking at the top level above function signatures.

### DIFF
--- a/components/clarinet-format/src/formatter/ignored.rs
+++ b/components/clarinet-format/src/formatter/ignored.rs
@@ -1,64 +1,85 @@
 use clarity::vm::representations::PreSymbolicExpression;
 
+fn u32_to_usize(val: u32) -> usize {
+    usize::try_from(val).unwrap()
+}
+
+/// spans use 1-based index, this const is largely only for this comment
+const COLUMN_TO_INDEX_OFFSET: usize = 1;
+
 /// Extract the original source text using its span
-/// This is useful for strings where we don't want to touch the formatting
 pub fn extract_expr_source(expr: &PreSymbolicExpression, source: &str) -> String {
     let span = &expr.span;
-    let start_line = usize::try_from(span.start_line).unwrap();
-    let start_column = usize::try_from(span.start_column).unwrap();
+    extract_source_range(
+        source,
+        span.start_line,
+        span.start_column,
+        span.end_line,
+        span.end_column,
+    )
+}
 
-    if let Some(line) = source.lines().nth(start_line - 1) {
-        let start_col = (start_column).saturating_sub(1);
-        let end_col = usize::try_from(span.end_column).unwrap();
+pub fn extract_source_range(
+    source: &str,
+    start_line: u32,
+    start_column: u32,
+    end_line: u32,
+    end_column: u32,
+) -> String {
+    let start_line = u32_to_usize(start_line);
+    let end_line = u32_to_usize(end_line);
+    let start_column = u32_to_usize(start_column);
+    let end_column = u32_to_usize(end_column);
 
-        if start_col < line.len() && end_col <= line.len() {
-            return line[start_col..end_col].to_string();
+    let lines: Vec<&str> = source.lines().collect();
+    let mut result = String::new();
+
+    for line_idx in start_line..=end_line {
+        if line_idx > lines.len() {
+            break;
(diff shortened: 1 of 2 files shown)


components/clarinet-format/src/formatter/ignored.rs
```diff
@@ -1,64 +1,85 @@
 use clarity::vm::representations::PreSymbolicExpression;
 
+fn u32_to_usize(val: u32) -> usize {
+    usize::try_from(val).unwrap()
+}
+
+/// spans use 1-based index, this const is largely only for this comment
```
xoloki-stacks, line 7: It's unclear what you mean by `this const is largely only for this comment`.  The plain reading is that the `const` is not used, but it is used in multiple places.  If you meant that the `const` is only used to convert a colums to an index (which the code seems to indicate), then maybe you should rewrite the comment to make it more clear?

components/clarinet-format/src/formatter/ignored.rs
```diff
@@ -1,64 +1,85 @@
...
+    for line_idx in start_line..=end_line {
+        if line_idx > lines.len() {
+            break;
         }
 
-        if idx == 0 {
-            // First line (the one with the opening parenthesis)
-            if let Some(paren_pos) = line.find('(') {
-                result.push_str(&line[paren_pos..]);
+        let line = lines[line_idx - COLUMN_TO_INDEX_OFFSET];
```
xoloki-stacks, line 42: Here again the usage of `COLUMN_TO_INDEX_OFFSET` is really unclear.  I would have expected it to be used when dealing with `start_column`/`end_column`, not `line_idx`.  

Maybe both lines and columns are 1-based?  If so, `COLUMN_TO_INDEX_OFFSET` needs to be renamed.

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -929,8 +1142,10 @@ impl<'a> Aggregator<'a> {
 
         // body expressions
         let mut prev_end_line = None;
-        for e in exprs.get(3..).unwrap_or_default() {
+        let body_exprs = exprs.get(3..).unwrap_or_default();
```
xoloki-stacks, line 1145: It's really not obvious to me where the magic number `3` comes from here.  It was in the previous code, so I don't think it's wrong, but it would be helpful for people unfamiliar with this code to explain why you need to pop the first few tokens to get`body_exprs`.

xoloki-stacks (approved):
I don't think there's anything wrong with this code, so I'm going to approve it in the service of getting it out the door tomorrow.  But it would be helpful in this or a follow-on PR to address my comments in regards to naming of `COLUMN_TO_INDEX_OFFSET` and the number of tokens to pop in `body_exprs`.

hugo-stacks:
Thanks for the thorough review @xoloki-stacks 
I need to mark comments as resolved i order to merge this PR.
So for the record, here are the links to the comments in case it's not obvious to find them once resolved:
- https://github.com/stx-labs/clarinet/pull/2110#discussion_r2629133720
- https://github.com/stx-labs/clarinet/pull/2110#discussion_r2629139558
- https://github.com/stx-labs/clarinet/pull/2110#discussion_r2629150417

@brady-stacks

components/clarinet-format/src/formatter/mod.rs
brady-stacks: above this we have the 1 and 2
```
        // asset-owner
        acc.push_str(&self.format_source_exprs(slice::from_ref(&exprs[1]), previous_indentation));

        // allowances
        if let Some(allowances_list) = exprs.get(2) {
```
The entirety of the PSE list should probably be parsed into a new structure in its own pass, but currently it's doing this indexing junk ☹️

PR #2117, opened 2025-12-17
fix: Enable `call_checker` and default lints in new project

jbencin-stacks:
### Description

- Add `call_checker` to `passes` in the default `Clarinet.toml`
- Rename lint group `recommended` to `default` (because it's easier to spell)
- Clean up strings for generating files using `indoc` macros

#### Breaking change?

It breaks the `recommended` lint group but I don't know if that's even made it into a release

### Checklist

- [x] I used `diff` to compare the results of `clarinet new` on this branch against `main`. It's basically the same, but the leading newlines in the config files have been removed (which is a good thing)

### DIFF
--- a/components/clarinet-utils/src/lib.rs
+++ b/components/clarinet-utils/src/lib.rs
@@ -239,8 +239,7 @@ mod tests {
         let _ = derive_key(password, &mut short)
             .expect_err("Should have failed with 1-byte output buffer");
 
-        derive_key(password, &mut right)
-            .expect("Should have succeeded with 32-byte output buffer");
+        derive_key(password, &mut right).expect("Should have succeeded with 32-byte output buffer");
     }
 
     #[test]


components/clarinet-cli/src/generate/project.rs
```diff
@@ -88,300 +89,309 @@ impl GetChangesForNewProject {
...
+            # [contracts.counter]
+            # path = "contracts/counter.clar"
+            # epoch = "latest"
+
+            [repl.analysis]
+            passes = ["check_checker", "call_checker"]
+            check_checker = {{ trusted_sender = false, trusted_caller = false, callee_filter = false }}
+
+            [repl.analysis.lint_groups]
+            default = true
```
hugo-stacks: Does this enables the new lints rules in new projects?

hugo-stacks (changes requested):
marking my review as "request changes", but it actually depends ont the answer to the question

components/clarinet-cli/src/generate/project.rs
jbencin-stacks: We can hold of on this for now

hugo-stacks (dismissed):

PR #2118, opened 2025-12-17
Suggest using encrypted mnemonics in testnet and mainnet TOML files

### LINKED ISSUE #2102: update generated Mainnet and Testnet setting files
`clarinet new` automatically generates setting files for mainnet and testnet

https://github.com/stx-labs/clarinet/blob/66679d4de2958e217db5d4b46cbece8cdcaf02f9/components/clarinet-cli/src/generate/project.rs#L193-L222

Instead of asking to set the `mnemonic`, they should encourage setting an encrypting mnemonic.
This could include a comment explaining that they need to use `clarinet deployments encrypt`

xoloki-stacks:
### Description

This PR updates the generated testnet and mainnet TOML files to encourage users to use the `encrypted_mnemonic` field instead of using plaintext mnemonics.

Fixes #2102 

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarinet-cli/src/generate/project.rs
+++ b/components/clarinet-cli/src/generate/project.rs
@@ -197,6 +197,15 @@ stacks_node_rpc_address = "https://api.testnet.hiro.so"
 deployment_fee_rate = 10
 
 [accounts.deployer]
+# It is strongly suggested to use encrypted mnemonics to avoid putting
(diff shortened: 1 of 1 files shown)


hugo-stacks (commented):
lgtm!
Thanks @xoloki-stacks

PR #2127, opened 2025-12-30
docs: Add console cheatsheet for debugging commands

jadonamite:
Added a cheatsheet for console commands to README.

### Example

---

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,13 @@ clarinet console
 (contract-call? .counter count-up)
 (contract-call? .counter get-count tx-sender)
 ```
+### Console Cheatsheet
+
+When debugging in `clarinet console`, use these commands to inspect the global chain state:
+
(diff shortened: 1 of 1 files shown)


hugo-stacks:
Hello @jadonamite 
Thanks for this contribution.
We try to avoid adding usage documentation to the READMEs, otherwise they become really big.
We prefer our product to be self documented as much as possible, for example, the console does print this on start.
```
Enter "::help" for usage hints.
```
👉 Maybe we can print it last (after the contracts list and assets maps) so that's it's easier to see.

Otherwise docs is added to docs.stacks.co

jadonamite:
Noted thanks for the feedback

hugo-stacks:
Let's do this:

>  print `::help` usage last (after the contracts list and assets maps) so that's it's easier to see

PR #2129, opened 2025-12-30
docs: add event assertion recipe

jadonamite:
Added a code snippet demonstrating how to filter and test contract events.


hugo-stacks:
Thanks for your contribution @jadonamite 
I'd rather it to https://github.com/stacks-network/docs/

What do you think?

jadonamite:
How would you want it @hugo-stacks

hugo-stacks:
Let's add it to the documentation repository instead 👍

PR #2132, opened 2026-01-05
feat(linter): Add `case_const` lint

jbencin-stacks:
### Description

> [!NOTE]
> I'm adding this lint to `LintGroup::Style`, which is not enabled by default
> Do we think style lints should be on by default?

Add `case_const` to linter (#2028)

This lint enforces `SCREAMING_SNAKE_CASE` on all constants. An identifier is considered `SCREAMING_SNAKE_CASE` if...
  - It contains only ASCII uppercase letters (`A`–`Z`), digits (`0`–`9`), and underscores (`_`)
  - Does **not** contain consecutive underscores (not allowed in Clarity anyways)

Note that leading or trailing underscores **are** allowed (but leading underscores aren't currently allowed in Clarity)

#### Breaking change?

No

### Example

Will generate a diagnostic message on a contract set like this:

```clarity
(define-constant this_is_regular_snake_case u100)
```

```clarity
;; #[allow(case_const)]
(define-constant this_is_regular_snake_case u100)
```

on the line above it

---

### Checklist

### DIFF
--- a/components/clarity-repl/src/analysis/lints/unused_map.rs
+++ b/components/clarity-repl/src/analysis/lints/unused_map.rs
@@ -1,6 +1,6 @@
 //! Lint to find unused map (`declare-map`) variables
 //!
-//! A dignostic is generated if:
+//! A diagnostic is generated if:
 //!  - The map is never referenced
 //!  - The map is used, but in a way that cannot affect contract execution
 
--- a/components/clarity-repl/src/analysis/lints/unused_data_var.rs
+++ b/components/clarity-repl/src/analysis/lints/unused_data_var.rs
@@ -1,6 +1,6 @@
 //! Lint to find unused global (`declare-data-var`) variables
 //!
-//! A dignostic is generated if:
+//! A diagnostic is generated if:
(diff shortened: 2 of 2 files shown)


brady-stacks:
> Do we think style lints should be on by default?

"Style lints" is somewhat broad but if it's not required by clarity or a danger to the contract in some way I'd say No

components/clarity-repl/src/analysis/lints/unused_map.rs
```diff
@@ -1,3 +1,9 @@
+//! Lint to find unused map (`declare-map`) variables
+//!
+//! A dignostic is generated if:
```
brady-stacks: diagnostic*

brady-stacks (approved):
LGTM
(style lints on by default 👍 )

components/clarity-repl/src/analysis/lints/unused_map.rs
jbencin-stacks: Fixed 1c54db3a

PR #2141, opened 2026-01-13
chore(linter): Enable lints by default

jbencin-stacks:
### Description

Enable most existing lints by default, at the `warning` level. This is what we are going to say in the release announcement, so the code needs to match. Users will have to explicitly disable lints if they don't want them, like this:

```toml
[repl.analysis.lint_groups]
all = false
```

Some additional notes:
- Not enabling the style lints yet by default. So far we only have one, and I want to have consensus on the conventions before finishing up and enabling this section. Will enable at the `notice` level when finished.
- Remove the `Default` lint group and replaced with a more flexible way to set defaults which can handle different lint levels
- Remove the `Misc` lint group. I don't think we'll use this, but I can always add it back
- Move the `Noop` lint to the `Safety` group. If you have one in your code, it's probably a mistake which will result in a bug

#### Breaking change?

Could possibly break a users CI if they run `clarinet check` and expect no errors

### Checklist

### DIFF
--- a/components/clarity-lsp/Cargo.toml
+++ b/components/clarity-lsp/Cargo.toml
@@ -5,7 +5,7 @@ edition = "2021"
 
 [dependencies]
 clarity = { workspace = true, default-features = false, features = ["developer-mode"] }
-clarity-types = { workspace = true, default-features = false, features = ["developer-mode"] }
+clarity-types = { workspace = true, features = ["developer-mode"] }
 lsp-types = "0.97.0"
 regex = { workspace = true }
 serde = { workspace = true, features = ["derive"] }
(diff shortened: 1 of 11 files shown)


jbencin-stacks:
The last test is failing because of how the `Clarinet.toml` is rewritten when  `clarinet contract new` runs. If I have this in `Clarinet.toml` initially:

```toml
[repl.analysis.lint_groups]
all = true
```

and I run `clarinet contract new <file>`, I end up with this:

```toml
[repl.analysis.lints]
case_const = 'Warning'
unused_token = 'Warning'
unused_trait = 'Warning'
unused_data_var = 'Warning'
unused_private_fn = 'Warning'
unused_const = 'Warning'
unused_binding = 'Warning'
noop = 'Warning'
unused_map = 'Warning'
```

Which is both invalid (should be `"warning"` instead of `'Warning'`), and not what the user configured. This bug is not new, and can be triggered in the existing `main` code by setting a lint group and running `clarinet contract new`, but adding these lints by default will trigger it on every `clarinet contract new` run, which is what made this unit test catch it

brady-stacks (dismissed):
👍

brady-stacks (approved):

PR #2149, opened 2026-01-25
perf: Improve Linux performance by using `jemalloc` instead of default allocator

jbencin-stacks:
### Description

Replace system allocator with `jemalloc` using the `tikv-jemallocator` crate. We have been doing this already in `stacks-core` for a couple years and it results in significant speed increases on supported platforms

Also, this PR fixes new Clippy error that occurs in Rust 1.93 in order to make CI pass 

#### Breaking change?

No

### Benchmarks

I used `hyperfine` to benchmark this PR against `main`. I ran `clarinet check` on `stacks-wormhole-ntt` for 20 iterations

I ran this for both the `glibc` and `musl` builds. The `musl` build shows a much larger performance boost, and I suspect that's what most Linux users are running, since it's easier to install

#### Against `glibc` build

The `jemalloc` version runs **16% faster** than the version using `glibc`s `malloc()`:

```console
❯ hyperfine -w 3 -r 20 "clarinet-std check" "clarinet-je check"
Benchmark 1: clarinet-std check
  Time (mean ± σ):     331.9 ms ±   4.8 ms    [User: 302.0 ms, System: 32.2 ms]
  Range (min … max):   326.8 ms … 343.0 ms    20 runs
 
Benchmark 2: clarinet-je check
  Time (mean ± σ):     285.3 ms ±   4.3 ms    [User: 264.2 ms, System: 24.0 ms]
  Range (min … max):   276.9 ms … 292.9 ms    20 runs
 
Summary
  clarinet-je check ran
    1.16 ± 0.02 times faster than clarinet-std check
``` 

#### Against `musl` build

The `jemalloc` version runs **54% faster** than the version using `musl`s `malloc()`:

```console
❯ hyperfine -w 3 -r 20 "clarinet-musl check" "clarinet-musl-je check"
Benchmark 1: clarinet-musl check
  Time (mean ± σ):     643.7 ms ±   5.1 ms    [User: 485.5 ms, System: 151.7 ms]
  Range (min … max):   637.6 ms … 663.0 ms    20 runs
 
Benchmark 2: clarinet-musl-je check
  Time (mean ± σ):     419.3 ms ±   6.9 ms    [User: 394.1 ms, System: 36.7 ms]
  Range (min … max):   410.5 ms … 434.1 ms    20 runs
 
Summary
  clarinet-musl-je check ran
    1.54 ± 0.03 times faster than clarinet-musl check
```

### Notes to Reviewers

I enable `jemalloc` on the following condition:

```rust
#[cfg(not(any(target_env = "msvc", target_os = "macos")))]
```

so basically it's only enabled in the Linux `clarinet` binary. I do not think this affects the WASM target at all.

`stacks-core` is somewhat more conservative and uses the following condition:

```rust
#[cfg(not(any(target_os = "macos", target_os = "windows", target_arch = "arm")))]
```

My choice for Clarinet is based on [this](https://chatgpt.com/c/69765cca-2770-8332-a180-5c247fa48820) chat with ChatGPT

### DIFF
--- a/components/clarity-repl/src/repl/hooks/perf.rs
+++ b/components/clarity-repl/src/repl/hooks/perf.rs
@@ -366,17 +366,11 @@ impl EvalHook for PerfHook {
                 }
 
                 for (call_stack, cost) in &self.collected_data {
-                    let final_call_stack =
-                        if overhead_runtime > 0 && self.contract_identifier.is_some() {
-                            format!(
-                                "{};{}",
-                                self.contract_identifier.as_ref().unwrap(),
-                                call_stack
-                            )
-                        } else {
-                            call_stack.clone()
-                        };
-                    writeln!(writer, "{} {}", final_call_stack, cost)
+                    let final_call_stack = match &self.contract_identifier {
+                        Some(id) if overhead_runtime > 0 => format!("{id};{call_stack}"),
+                        _ => call_stack.clone(),
+                    };
+                    writeln!(writer, "{final_call_stack} {cost}")
                         .expect("Failed to write to perf output");
                 }
             }


brady-stacks:
Looks like this should close #2152 also 💯

PR #2151, opened 2026-01-26
perf: Replace `std::collections::HashMap` with `hashbrown::HashMap`

jbencin-stacks:
### Description

Replace `std::collections::HashMap` with `hashbrown::HashMap`

Although `std::collections::HashMap` is based on `hashbrown::HashMap`, it uses a different hash algorithm by default (`foldhash::fast` instead of SipHash), which is responsible for the performance difference.

#### Breaking change?

No

### Benchmarks

It's about 2% faster running `clarinet check` on the `stacks-wormhole-ntt` repo

```console
❯ hyperfine -w 5 -r 100 "clarinet-std check" "clarinet-hashbrown check"
Benchmark 1: clarinet-std check
  Time (mean ± σ):     334.4 ms ±   5.7 ms    [User: 304.5 ms, System: 32.2 ms]
  Range (min … max):   327.5 ms … 353.2 ms    100 runs
 
Benchmark 2: clarinet-hashbrown check
  Time (mean ± σ):     329.3 ms ±   5.5 ms    [User: 296.6 ms, System: 34.9 ms]
  Range (min … max):   321.8 ms … 345.6 ms    100 runs
 
Summary
  clarinet-hashbrown check ran
    1.02 ± 0.02 times faster than clarinet-std check
```

> [!NOTE]
> I'd like to add VSCode extension (WASM) benchmarks also

### Checklist

- [x] `clarinet check` benchmarks
- [ ] VSCode extension benchmarks

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -785,6 +785,7 @@ dependencies = [
  "clarity-repl",
  "console_error_panic_hook",
  "gloo-utils",
+ "hashbrown 0.16.1",
  "js-sys",
  "serde",
  "serde-wasm-bindgen",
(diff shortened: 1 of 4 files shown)


jbencin-stacks:
Closing due to too many conflicts. It would be easier to re-do the entire thing than to resolve them

PR #2153, opened 2026-01-28
chore: remove generated completions files

brady-stacks:
The completions are generated with `clap` so there's no reason to keep them checked into git. Added documentation for an example of how to add a specific zsh completion


components/clarinet-cli/completions/README.md
```diff
@@ -1,14 +1,33 @@
...
+For example, if you want zsh completions
+
+```sh
+mkdir -p ~/.zsh/completions
+# Generate the completion file
+clarinet completions zsh
+mv _clarinet ~/.zsh/completions/_clarinet
+
+# Make sure the directory is in fpath
+echo 'fpath=(~/.zsh/completions $fpath)' >> ~/.zshrc
```
jbencin-stacks, line 28: Is there a standard place to put these with ZSH so you don't have to modify any config files?

components/clarinet-cli/completions/README.md
```diff
@@ -1,14 +1,33 @@
...
 
 ```sh
 clarinet completions (bash|elvish|fish|powershell|zsh)
-```
\ No newline at end of file
+```
+
+# Example
+
+For example, if you want zsh completions
```
jbencin-stacks, line 19: I think you should add an example for Bash also, but every location I've tried doesn't work, and I'd rather not suggest adding `source ...` to config files, but that might be the most reliable way to do it

components/clarinet-cli/completions/README.md
brady-stacks: I think it's `/usr/share/zsh/site-functions` on mac and `/usr/local/share/zsh/site-functions` on linux so it seems more confusing. This is probably why it simply said "Follow the instructions specific to your shell..." previously 😅

PR #2154, opened 2026-01-29
Remove code marked as #[env(simnet)] when deploying elsewhere

### LINKED ISSUE #2022: Conditionally deploy code in Simnet
## Description

Using annotations, Clarinet should let developers specify that a particular function or expression should be deployed only in Simnet.

By implementing this, we would not only enable developers to use it directly but also improve their experience with testing frameworks like `clarunit` and `rendezvous`.
Making it easier to **write tests in Clarity**.

Clarinet should check that the contract remains valid both _with_ and _without_ the conditionally deployed code.

### Comment annotation

The annotation could be `;; #[env(simnet)]`, similar to existing annotations that we have in the linter (eg, `;; #[allow(unchecked_data)]`) — and inspired by Rust conditional compilation flags.
```clar
;; #[env(simnet)]
(define-private (faucet (recipient principal))
  (ft-mint? SuperFT u1000 recipient)
)
```

### Validating code

One of the benefits of this feature, compared to the workaround that third-party tools had to find, is that the testing code can be considered as first-class Clarity code, and therefore benefits from all of the tools that exist in the lsp (check, linter, documentation, autocompletion, etc).

It means that, when a Clarinet project contains such annotations, `clarinet check` and the LSP diagnostics should make two check passes **with** and **without** the simnet-only code.

So that this contract would be invalid:

```clar
;; clarinet check should throw something like
;; `use of unresolved function 'another-func', only available on simnet`
(define-public (my-func)
  (another-func u"hello")
)

;; #[env(simnet)]
(define-public (another-func (msg (string-utf8 128)))
  (print u"hello")
)
```

### Testing function

Other tools had to implement this logic. See how [clarunit](http://github.com/stacks-network/clarunit?tab=readme-ov-file#adding-tests) did it, or how [rendezvous](https://stacks-network.github.io/rendezvous/chapter_6.html) did it.
It comes with major caveats:
- Not standardized across the ecosystem
- Hard to implement for test framework developers
  - In rendezvous, the testing code is injected into the regular smart contracts, and rendezvous handles a custom deployment with the new contracts
- Test code doesn't benefit from the LSP (auto completion, documentation, linter, etc)

This feature would overcome all of these caveats.
Here is an example from the `rendezvous` documentation. Instead of putting this code in a separate file `counter.test.clar`, developers could put it directly in `counter.clar` and have the code checked/linted.

```clar
;; counter.clar

(define-public (increment)
  (ok (var-set counter (+ (var-get counter) u1)))
)

;; #[env(simnet)]
(define-public (test-increment)
  (let ((counter-before (get-counter)))
    (unwrap-panic (increment))
    (asserts! (is-eq (get-counter) (+ counter-before u1)) (err u404))
    (ok true)
  )
)
```

### Admin function

Here, we want the `(faucet)` function to be deployed in simnet.
In this example, note that private function can be called like public functions in simnet, so the testing code _can_ be private functions
```clar
(define-fungible-token SuperFT)

;; #[env(simnet)]
(define-private (faucet (recipient principal))
  (ft-mint? SuperFT u1000 recipient)
)
```

### Other use cases

Any expression could be annotated to be deployed only in simnet. In the example below, add a print only in simnet (eg: for testing purpose).

```clar
  (begin
    (try! (is-admin))
    ;; #[env(simnet)]
    (print "updating ata admin")
    (ok (var-set admin new-admin))
  )
)
```

## What this is NOT

This annotation for conditional deployment will be super powerful, but also somehow dangerous for developers who misuse it (_with great powers [...]_).

So we want to:
- keep it simple
- easy to use
- hard to misuse

While Rust's conditional compilation slightly inspires it, we don't want something as complex (nor as powerful).
It should only work Simnet, and not for other env: eg, `[env(testnet)]`, which I think would be a big footgun. Similarly, we don't want to have complex conditions such as `[env(and(...))]` or `[env(or(...))]`.
Having this only for `simnet` is already super powerful and makes a lot of sense.

If developers need to implement different behaviors in their contracts for testnet/mainnet, they can already use features like `is-in-mainnet`.

xoloki-stacks:
### Description

This PR implements a conditional deployment annotation for clarity.  Code which is marked as `#[env(simnet)]` will be stripped out when deployed to other networks.  

Fixes #2022

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -826,7 +826,7 @@ pub fn main() {
                     (_, _) => unreachable!()
                 };
 
-                let mut deployment = match result {
+                let deployment = match result {
                     Ok(deployment) => deployment,
                     Err(e) => {
                         eprintln!("{e}");


components/clarity-repl/src/utils.rs
```diff
@@ -67,3 +73,122 @@ pub fn serialize_event(event: &StacksTransactionEvent) -> serde_json::Value {
...
+    settings.repl_settings.analysis.enable_all_passes();
+
+    let session = Session::new(settings.clone());
+    let epoch = DEFAULT_EPOCH;
+    let contract = ClarityContract {
+        code_source: ClarityCodeSource::ContractInMemory(source.clone()),
+        deployer: ContractDeployer::Address(id.issuer.to_string()),
+        name: id.name.to_string(),
+        clarity_version: ClarityVersion::default_for_epoch(epoch),
+        epoch: Epoch::Specific(epoch),
```
brady-stacks: I'm assuming this should get the epoch/version from somewhere? I'm just thinking of the scenario where you try to load an old clarity contract within the latest and it fails to load because of deprecated functions or something.

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -855,6 +856,17 @@ pub fn main() {
                     }
                 }
 
+                // if we aren't deploying to simnet, load AST for each contract so we can strip out anything marked as #[env(simnet)]
+                if network != StacksNetwork::Simnet {
+                    for (id, (ref mut source, loc)) in &mut deployment.contracts {
+                        let Ok(clean) = remove_env_simnet(id, source.to_string()) else {
+                            println!("Failed to remove code tagged as #[env(simnet)] from {loc}");
+                            std::process::exit(1);
```
brady-stacks: should this be an exit or a warning? I could see it going either way, just curious what you think

xoloki-stacks: We generally bail out in the CLI at the first sign of a problem, so I went with that here.  Thinking further, if people are going to put things into contracts that they absolutely don't want on mainnet, any problems in the stripping process seem like hard fails.

components/clarity-repl/src/utils.rs
xoloki-stacks: Good catch, looks like the clarity version and epoch are available in the deployment plan, so I can pass them into here.

xoloki-stacks: Ok, had to move the call to `remove_env_simnet` deep in the deployment parsing code, but now we are properly getting version and epoch

PR #2156, opened 2026-01-29
fix: missing Clarinet.toml warning now logs ONLY (no pop up)

### LINKED ISSUE #2155: Only raise Clarinet-specific IDE warnings if in a Clarinet workspace
**Is your feature request related to a problem? Please describe.**

I'm not sure if there are others, but for me specifically the "No Clarinet.toml associated with foo.clar" warning pops up often when working in a workspace which has `.clar` files but where Clarinet isn't used (e.g. in `stacks-core` & `sbtc` in VSCode). I ended up disabling all error/warning notifications from the extension, which feels unfortunate as I'd prefer to have them if I'm actually in a Clarinet workspace.

**Describe the solution you'd like**
1. Ideally the extension would detect if the user is in a Clarinet-enabled workspace or not prior to raising Clarinet-specific warnings/errors. 
2. Alternatively, comment-based shortcodes could be used (e.g. a top-of-contract `;; #clarinet(disable)` or so, similar to how `markdownlint` works) -- a paradigm which could be useful for controlling the new linter later as well.

**Describe alternatives you've considered**
As noted, I disabled all warnings from the extension, which is less than ideal.

brady-stacks:
- closes #2155 

Instead of a pop up with "No Clarinet.toml associated..." we log it to the console (defaulting the clarity version to latest, since we don't have that available without the manifest)

### DIFF
--- a/components/clarity-lsp/src/common/backend.rs
+++ b/components/clarity-lsp/src/common/backend.rs
@@ -193,7 +193,7 @@ pub async fn process_notification(
                                 // The contract can still be opened without a Clarinet.toml
                                 #[cfg(target_arch = "wasm32")]
                                 log!(
-                                    "No Clarinet.toml is associated to the contract {}",
+                                    "No Clarinet.toml is associated with the contract '{}'",
                                     contract_location_string
                                 );
                                 // Note: perhaps the version should alternatively allow setting within the contract source?


components/clarity-lsp/src/common/backend.rs
```diff
@@ -189,10 +189,15 @@ pub async fn process_notification(
                                     get_boot_contract_epoch_and_clarity_version(contract_name);
                                 version
                             } else {
-                                return Err(format!(
+                                // Log warning instead of returning error since this is non-fatal
+                                // The contract can still be opened without a Clarinet.toml
+                                #[cfg(target_arch = "wasm32")]
+                                log!(
                                     "No Clarinet.toml is associated to the contract {}",
```
cylewitruk: Minor suggestion:

```suggestion
                                    "No Clarinet.toml is associated with the contract '{}'",
```

components/clarity-lsp/src/common/backend.rs
```diff
@@ -189,10 +189,15 @@ pub async fn process_notification(
...
                                     contract_location_string
-                                ));
+                                );
+                                // Note: perhaps the version should alternatively allow setting within the contract source?
+                                clarity_repl::clarity::ClarityVersion::latest()
```
cylewitruk, line 200: Personally I think using `latest` if no explicit version configuration exists is a reasonable default; Most languages will use the latest [installed] version if not specifically specified, e.g. if no `rust-toolchain.toml` -- and in this case (no manifest) only linting should be affected, right?

Re: your comment on Slack, I think a pragma like `;; @clarity:version=4` or similar would be useful, particularly if you're working on a generational dApp where you maybe have already-deployed contracts on previous versions and you're adding a new contract at the current version. Ideally this would be a language feature though, imo (e.g., `@version=4`) -- but these probably warrant separate issues.

brady-stacks: > whether or not you want to keep the warning pop-up when a Clarinet manifest exists?

Currently we can't tell if a project is clarinet or not so I think we'd have to disable the pop up. 

The stacks-core pragma issue is [here](https://github.com/stacks-network/stacks-core/issues/6836)
And a follow up dependent on it for clarinet is [here](https://github.com/stx-labs/clarinet/issues/2157)

cylewitruk: @brady-stacks doesn't the following at the top of the match arm essentially reveal if you're in a Clarinet context or not?

https://github.com/stx-labs/clarinet/blob/382f9ebc9806c8e0bdfa551a4d378f810a7ee922/components/clarity-lsp/src/common/backend.rs#L127-L129

If so, then I _think_ it should be able to be refactored to return an error if it exists (and thus pop-up), and log-only if not (but the structure of the entire match would need to be altered). I have zero experience with VSCode/LSP dev though, so I could be making things up :)

hugo-stacks: I think defaulting to `latest()` might be more annoying than the popup itself.
In the use case described be Cyle (opening boot contracts in the stacks-core repo) it would mean that almost all of them would be checked with the wrong clarity version 🤔

cylewitruk: Well, that's a good point @hugo-stacks, which would also apply to all of the test contract stubs in [`clarity`](https://github.com/stacks-network/stacks-core/tree/master/clarity/src/vm/analysis/type_checker/v2_1/tests/contracts) as well (and a number of my small little test projects).

That's where pragmas would be useful (for linting), but I could also see people not wanting to modify potentially already deployed contract files to add a pragma.

Realistically, as long as irrelevant fields aren't required in a `Clarinet.toml` (like, that I wasn't forced to make up values for fields that have no meaning without Clarinet), then I personally wouldn't have any problem creating a little manifest file with the `[contracts.foo]` directives. Maybe it would need a `[project] clarinet = false` or something to disable it looking for additional files, I'm not familiar with what it requires 🤷

hugo-stacks: @cylewitruk 
Only the project name is required. This is a valid manifest:

```toml
[project]
name = 'project-template'

[contracts.counter]
clarity_version = 4
epoch = 'latest'
```

cylewitruk: @hugo-stacks and is that everything that the LSP needs to stop bugging me in VSCode? If that's the case then I'm totally fine with that!

Couple Q's:

1. Is there any wildcard/glob support for configuring multiple contracts with one entry?
2. Similarly, something like the following could be useful:

```toml
[project]
name = 'foo'
ignore = [
    'tests/contract_stubs/*',
    'poc/my_contract.clar',
]
```

_(I'm specifically thinking of e.g. the [clarity](https://github.com/stacks-network/stacks-core/tree/master/clarity/src/vm/analysis/type_checker/v2_1/tests/contracts) dir)_

brady-stacks:
closing in favor of #2177

PR #2161, opened 2026-02-02
chore: use most recent stacks-core/develop

brady-stacks:
pinned to commit [6c76d35624e2c08ff96f658d0d09b0bf9867ccaf](https://github.com/stacks-network/stacks-core/commit/6c76d35624e2c08ff96f658d0d09b0bf9867ccaf)


components/clarity-repl/src/analysis/ast_dependency_detector.rs
```diff
@@ -10,7 +10,7 @@ use clarity::vm::ast::ContractAST;
 use clarity::vm::representations::{SymbolicExpression, TraitDefinition};
 use clarity::vm::types::{FunctionSignature, TypeSignatureExt};
 use clarity::vm::{ClarityName, ClarityVersion, SymbolicExpressionType};
-use clarity_types::errors::analysis::CheckErrorKind;
+use clarity_types::errors::analysis::RuntimeCheckErrorKind as CheckErrorKind;
```
hugo-stacks: Can we avoid the `as` here and just use the real name?

Issue #2164, opened 2026-02-03
`clarinet contracts <new|rm>` rewrites `Clarinet.toml` file incorrectly


jbencin-stacks:
**Describe the bug**

When a user runs `clarinet contracts <new|rm>`  or `clarinet requirements add`, the project file file `Clarinet.toml` is modified and written. This is not done how it should be: Comments are stripped, and missing/optional fields are resolved to their default values and written to the file

**To Reproduce**

1. Create a new project with `clarinet new`
2. Copy the project file: `cp -a Clarinet.toml Clarinet.toml.old`
3. Add a file to the project with `clarinet contracts add <name>`
4. Diff the new project file: `diff Clarinet.toml.old Clarinet.toml`

**Expected behavior**

Newly written config file should match user's previous config file as closely as possible, and preserve comments and optionality of config items

**Additional context**

This was specifically creating issues for linter config before the 3.13.0 release, but still affects other items

PR #2167, opened 2026-02-05
fix: Preserve user settings, key ordering, and comments when editing `Clarinet.toml`

jbencin-stacks:
### Description

When running `clarinet contracts ...` or `clarinet requirements ...`, use the `toml_edit` crate to make changes to `Clarinet.toml`, rather than serializing a `ProjectManifest` struct to re-write the entire file. This will preserve the user's existing settings, key ordering, and comments

Fixes: #2164

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -2231,7 +2231,6 @@ mod tests {
 
         #[test]
         fn test_add_contract_preserves_comments() {
-            #[rustfmt::skip]
             let input = indoc! {r#"
                 [project]
                 name = "test-project"
@@ -2286,7 +2285,6 @@ mod tests {
 
         #[test]
         fn test_remove_contract_preserves_comments_and_other_contracts() {
-            #[rustfmt::skip]
             let input = indoc! {r#"
                 [project]
                 name = "test-project"
@@ -2332,7 +2330,6 @@ mod tests {
 
         #[test]
         fn test_add_requirement_preserves_comments() {
-            #[rustfmt::skip]
             let input = indoc! {r#"
                 [project]
                 name = "test-project"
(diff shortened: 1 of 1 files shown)


components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -2095,4 +2138,682 @@ mod tests {
...
+                    } else {
+                        return false;
+                    }
+                }
+            }
+            false
+        }
+
+        /// Helper to check if a contract exists in the TOML
+        fn has_contract(content: &str, contract_name: &str) -> bool {
```
brady-stacks, line 2206: has_toml, has_contract, and has_requirement are probably generally useful outside of the test. Not suggesting moving them now, but just saying "Nice!" for whoever needs them in the future

brady-stacks (approved):
Looks great 💯

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -2095,4 +2138,682 @@ mod tests {
...
+                # This is an important comment that should be preserved!
+                # Another comment line
+
+                [repl.analysis]
+                passes = [\"check_checker\"]
+
+                # Check-checker settings comment
+                [repl.analysis.check_checker]
+                strict = false
+            "};
```
hugo-stacks: can we automatically escape double quotes with `r#"..."#` here and in the other tests?

jbencin-stacks: I had to tell it to re-write all the strings using `indoc` also. Claude is great at architecture but it's default Rust style feels dated and ugly (or maybe it's just inferring that style from the existing code)

hugo-stacks: As we notice the "dated and ugly" patterns, we can add recommendations in the root Claude.md

jbencin-stacks: Done [be6f6fe](https://github.com/stx-labs/clarinet/pull/2167/commits/be6f6fe4135159010645cdc32ae7528fed7aa5b0)

PR #2169, opened 2026-02-06
feat: support epoch 3_4 and clarity 5

hugo-stacks:
### Description

- upgrade stacks-core to the latest commit
- add support for epoch 3.4
  - it supports using both a 3.3.x or a 3.4.x image in Devnet
- add support for clarity 5
- add helpers to make it slightly easier to add new epochs and clarity version in the future

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -31,7 +31,8 @@ use clarity_repl::frontend::Terminal;
 use clarity_repl::repl::diagnostic::output_diagnostic;
 use clarity_repl::repl::settings::{ApiUrl, RemoteDataSettings};
 use clarity_repl::repl::{
-    ClarityCodeSource, ClarityContract, ContractDeployer, Epoch, DEFAULT_EPOCH,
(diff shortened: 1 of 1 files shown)


components/clarinet-files/src/network_manifest.rs
```diff
@@ -175,6 +175,7 @@ pub struct DevnetConfigFile {
     pub epoch_3_1: Option<u64>,
     pub epoch_3_2: Option<u64>,
     pub epoch_3_3: Option<u64>,
+    pub epoch_3_4: Option<Option<u64>>,
```
brady-stacks: Is the nested value significant? Could flatten it if a value like `Some(None)` isn't needed.

brady-stacks (approved):
My comment above is non-blocking. LGTM (great refactor 💯 )

components/clarinet-files/src/network_manifest.rs
hugo-stacks: I think you right. I implemented it naively as if it was "more optional" than the others, but that doesn't make sense in a file..
I'll fix

hugo-stacks:
Fixed @brady-stacks

hugo-stacks:
Broke after a change on main, now fixed. 
Can I have a re-review @brady-stacks?

PR #2170, opened 2026-02-06
chore: upgrade cargo dependencies

hugo-stacks:
### Description

- upgrade many cargo dependencies
- upgrade time to fix new vuln
- remove obsolete `--ignore` in cargo audit (neon was removed a while ago #2005)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ ratatui = { version = "0.30.0", default-features = false, features = [
     "crossterm",
 ] }
 regex = "1.12.3"
-reqwest = { version = "0.12", default-features = false, features = [
+reqwest = { version = "0.12.28", default-features = false, features = [
     "json",
     "rustls-tls",
 ] }


Cargo.toml
```diff
@@ -36,44 +36,45 @@ base58 = "0.2.0"
...
 pretty_assertions = "1.4"
 rand = "0.8.5"
-ratatui = { version = "0.29.0", default-features = false, features = [
+ratatui = { version = "0.30.0", default-features = false, features = [
     "crossterm",
 ] }
-regex = "1.11.2"
-reqwest = { version = "0.12.22", default-features = false, features = [
+regex = "1.12.3"
+reqwest = { version = "0.12", default-features = false, features = [
```
xoloki-stacks: Why remove the version lock here and nowhere else?

hugo-stacks: I should've leave it. I've been playing with different versions of reqwest (0.13.x) but sticking with 0.12.x for now (it's a dependency of other packages). So the change wasn't voluntary
thanks

Cargo.toml
```diff
@@ -36,44 +36,45 @@ base58 = "0.2.0"
...
 pretty_assertions = "1.4"
 rand = "0.8.5"
-ratatui = { version = "0.29.0", default-features = false, features = [
+ratatui = { version = "0.30.0", default-features = false, features = [
     "crossterm",
 ] }
-regex = "1.11.2"
-reqwest = { version = "0.12.22", default-features = false, features = [
+regex = "1.12.3"
+reqwest = { version = "0.12.28", default-features = false, features = [
```
xoloki-stacks, line 54: Much better, thanks!

PR #2173, opened 2026-02-06
chore: migrate to ls-types instead of lsp-types

hugo-stacks:
### Description

- upgrade tower-lsp-types
- use ls-types: a maintained fork of lsp-types

context: https://github.com/tower-lsp-community/tower-lsp-server/blob/main/CHANGELOG.md#0230---2025-12-07

### DIFF
--- a/components/clarity-lsp/src/common/backend.rs
+++ b/components/clarity-lsp/src/common/backend.rs
@@ -4,16 +4,15 @@ use clarinet_files::{FileAccessor, FileLocation, ProjectManifest};
 use clarity_repl::clarity::diagnostic::Diagnostic;
 use clarity_repl::repl::boot::get_boot_contract_epoch_and_clarity_version;
 use clarity_repl::repl::ContractDeployer;
-use lsp_types::{
+use ls_types::{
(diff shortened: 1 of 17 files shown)


hugo-stacks:
@xoloki-stacks Follow-up of #2170 but I wanted to put it in its own PR

components/clarity-lsp/src/common/backend.rs
```diff
@@ -724,7 +723,7 @@ mod lsp_tests {
             panic!("Expected a Definition response, got: {response:?}");
         };
 
-        assert_eq!(location.uri.scheme().unwrap().as_str(), "file");
+        assert_eq!(location.uri.scheme().as_str(), "file");
```
xoloki-stacks, line 726: Why does the `unwrap` go away here?

hugo-stacks: I wasn't able to pinpoint the exact change in fluent-uri between 0.1 and 0.3, it look like the Uri type was fully removed in 0.2 and re-added in 0.3. But regardless, it doesn't return an Option anymore

PR #2175, opened 2026-02-06
feat: improve deployment plan transaction structure

hugo-stacks:
### Description

Replace the deprecated `serde_yaml` crate with the newer official `yaml_serde` and improve deployment plan structure.

- `TransactionSpecificationFile` enum serialization was incompatible with `yaml_serde`
- `serde_yaml` serialized externally-tagged enums as YAML map keys `variant: <fields>`, but `yaml_serde` expects !tag syntax, causing deserialization errors
- Use `transaction-type: <variant-name>` with a custom Deserialize impl that also accepts the old map-key format for backward compatibility

Overall, I think it results in a cleaner format for deployment plan

Screenshot 2026-02-06 at 18 20 25

There's no breaking change, the backward compat is handled


brady-stacks (approved):
👍

hugo-stacks:
I fixed the conflicts 
@brady-stacks Sorry, I pushed forced 😬 
The only conflict was on this file that was modified on both PRs https://github.com/stx-labs/clarinet/commit/c1df374c84ea558c523e93d874097ec226a9f827#diff-164c17e3a12dadcbd5040fb366e787cc29b529723339c4c7c947a87972dcd4b1R71

brady-stacks (dismissed):
LGTM

components/clarinet-sdk/node/tests/deployment-plan.test.ts
```diff
@@ -42,8 +42,7 @@ describe("generate deployment plans test", async () => {
   })
 });
 
-describe("initSiment deployment plans test", async () => {
-
+describe("initSimnet deployment plans test", async () => {
```
xoloki-stacks, line 45: Good catch!

Cargo.toml
```diff
@@ -60,8 +60,8 @@ schemars = "1.2.1"
 serde = "1.0.225"
 serde_json = "1.0.149"
 serde_derive = "1.0.225"
-serde_yaml = "0.8.23"
 serde-wasm-bindgen = "0.6.4"
+yaml_serde = "0.10"
```
xoloki-stacks: Why don't we specify a rev here like for every other dependency?  Looks like `0.10.3` is most recent.

hugo-stacks: Fixed

PR #2176, opened 2026-02-06
feat: add hex string support to toBeBuff matcher

hstove-stacks:
This PR adds support for passing a hex string to the `toBeBuff` matcher. I've also updated the README with a fixed link to `wasm-pack`, since the `rustwasm` org is no longer a thing.

### DIFF
--- a/components/clarinet-sdk/README.md
+++ b/components/clarinet-sdk/README.md
@@ -34,7 +34,7 @@ The SDK mainly relies on two components:
 - the TS component: `components/clarinet-sdk`
 
 To work with these two packages locally, the first one needs to be built with
-wasm-pack (install [wasm-pack](https://drager.github.io/wasm-pack/)).
+wasm-pack (install [wasm-pack](https://drager.github.io/wasm-pack/installer)).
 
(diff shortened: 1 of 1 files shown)


hugo-stacks:
Thanks @hstove-stacks!

Removing the trailing spaces in markdown changes how it's rendered; is this change intentional (as in, you think it's better)?

Screenshot 2026-02-06 at 20 40 53

hstove-stacks:
@hugo-stacks huh, I didn't even know that was a thing! The trailing whitespace being removed was just the result of my IDE auto-formatting.

IMO it does look odd, moreso depending on the breakpoint, ie:

image

I'll remove though, as it's unrelated to the PR

hugo-stacks:
Right maybe it's better without the line break.

Markdown has this weird trailing space handling so that you can add line breaks in the text (if you want to limit to 80 characters for instance) without impact the rendered document. Making it easier to version.

But yeah, unrelated to this PR 👍

hugo-stacks (approved):
LGTM! Thanks @hstove-stacks

PR #2178, opened 2026-02-09
Use `PreSymbolicExpressions` when parsing source for conditional deployment

xoloki-stacks:
### Description

The previous change to allow conditional deployment did a full `AST` parse into `SymbolicExpressions`, which unfortunately did not work well with annotations when the `AST` was reordered due to intra-contract dependencies.  This change instead uses only the first pass of the `AST` parser into `PresymbolicExpressions`, which means that nothing is reordered so the annotations will be tied to the correct expressions.

#### Breaking change?

None

### Example

The unit test has been updated to include intra-contract dependencies, and it still ties the annotation to the correct top level function.

---

### Checklist

### DIFF
--- a/components/clarity-repl/src/utils.rs
+++ b/components/clarity-repl/src/utils.rs
@@ -1,11 +1,10 @@
+use ::clarity::types::StacksEpochId;
+use ::clarity::vm::ast::parser;
 use ::clarity::vm::events::{FTEventType, NFTEventType, STXEventType, StacksTransactionEvent};
-use ::clarity::vm::types::QualifiedContractIdentifier;
-use ::clarity::vm::ClarityVersion;
+use ::clarity::vm::representations::PreSymbolicExpressionType::Comment;
 
 use crate::repl::clarity_values::value_to_string;
-use crate::repl::{
-    ClarityCodeSource, ClarityContract, ContractDeployer, Epoch, Session, SessionSettings,
-};
+use crate::repl::Epoch;
 
 pub fn serialize_event(event: &StacksTransactionEvent) -> serde_json::Value {
     match event {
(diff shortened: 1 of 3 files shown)


components/clarity-repl/src/utils.rs
```diff
@@ -73,65 +72,44 @@ pub fn serialize_event(event: &StacksTransactionEvent) -> serde_json::Value {
...
-        epoch,
+pub fn remove_env_simnet(epoch: Epoch, source: String) -> Result<String, String> {
+    let (pre_expressions, mut _diagnostics, success) = if epoch >= StacksEpochId::Epoch21 {
+        parser::v2::parse_collect_diagnostics(&source)
+    } else {
+        let parse_result = parser::v1::parse(&source);
+        match parse_result {
+            Ok(pre_expressions) => (pre_expressions, vec![], true),
+            Err(error) => (vec![], vec![error.diagnostic], false),
+        }
```
hugo-stacks, line 83: We've been having issues in the past with the parser V1 being too slow, especially in Wasm.
To the point where we had a hack in clarinet to force contract on epoch 2.1 in order to only use V2.

We recently got rid of this hack by using a flag in stacks-core instead: https://github.com/stacks-network/stacks-core/pull/6764, that we'll be able to remove once the core team eventually fully remove the parser v1.

But in Clarinet, we should only rely on parser V2

hugo-stacks (changes requested):

components/clarity-repl/src/utils.rs
xoloki-stacks: I merged this PR earlier when GitHub was acting wonky, so the branch did get squash/merged into main.  I'll open a new PR, remove the epoch parameter, and always use `parser::v2`.

xoloki-stacks: New PR is live: #2180

xoloki-stacks:
Closing this because GitHub merged the branch without doing whatever database magic is required to close the PR.  Note to self: never try to merge anything while GitHub is acting strangely.

Issue #2179, opened 2026-02-09
Replace `FileLocation` with `PathBuf`


hugo-stacks:
We have a  custom FileLocation to handle all of files paths/uri in Clarinet

https://github.com/stx-labs/clarinet/blob/8ff236c8f94f90a14aca9b87bf7e5c0e755d2955/components/clarinet-files/src/lib.rs#L58-L63

This abstraction adds unnecessary complexity in many places that have caused many issues and limitations in the past. `get_project_root_location` has been broken many times, especially in Wasm context

It should be possible to remove it in favor of only `PathBuf`, and eventually, converts Urls into `PathBuf` in the Wasm boundaries

PR #2182, opened 2026-02-10
refactor: remove FileLocation

hugo-stacks:
### Description

- Remove `FileLocation` in favor of `Path` or `PathBuf`. It mostly results in simplifications in the code, except in some place where the conversation of files URIs has to be handled, especially in wasm
- Refresh some part of the clarity-vscode integration tests
- Refactor `components/clarinet-cli/src/frontend/cli.rs` a bit further - it could still benefit from a lot of improvements, but I think it's in a good enough place (in the context of this PR)

Next, I'd like to refactor the `FileAccessor` trait, that could result in way less `to_string_lossy()` because the `FileAccessor` methods could likely accept `PathBuf` instead of `String`. I think it belongs in an other PR

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -42,10 +42,7 @@ use toml_edit::DocumentMut;
 #[cfg(feature = "telemetry")]
 use super::telemetry::{telemetry_report_event, DeveloperUsageDigest, DeveloperUsageEvent};
 use crate::deployments::types::DeploymentSynthesis;
-use crate::deployments::{
-    self, check_deployments, generate_default_deployment, get_absolute_deployment_path,
-    write_deployment,
-};
+use crate::deployments::{self, check_deployments, generate_default_deployment, write_deployment};
 use crate::devnet::package::{self as Package, ConfigurationPackage};
 use crate::devnet::start::{start, StartConfig};
 use crate::generate::changes::{Changes, TOMLEdition};
@@ -710,8 +707,6 @@ pub fn main() {
                     StacksNetwork::Simnet
                 };
 
-                let default_deployment_path =
-                    get_default_deployment_path(&manifest, &network).unwrap();
                 let (mut deployment, _) =
                     match generate_default_deployment(&manifest, &network, cmd.no_batch) {
                         Ok(deployment) => deployment,
@@ -747,19 +742,23 @@ pub fn main() {
 
                 let project_root =
                     paths::project_root_from_manifest_location(&manifest.location).unwrap();
+                let default_deployment_path =
+                    project_root.join(get_default_deployment_path(&network));
 
                 let write_plan = if default_deployment_path.exists() {
-                    let existing_deployment = load_deployment(&manifest, &default_deployment_path)
-                        .unwrap_or_else(|message| {
-                            eprintln!(
-                                "{}",
-                                format_err!(format!(
-                                    "unable to load {}\n{message}",
-                                    default_deployment_path.display(),
-                                ))
-                            );
-                            process::exit(1);
-                        });
+                    let existing_deployment =
+                        load_deployment(&project_root, &default_deployment_path).unwrap_or_else(
+                            |message| {
+                                eprintln!(
+                                    "{}",
+                                    format_err!(format!(
+                                        "unable to load {}\n{message}",
+                                        default_deployment_path.display(),
+                                    ))
+                                );
+                                process::exit(1);
+                            },
+                        );
                     should_existing_plan_be_replaced(
                         &existing_deployment,
                         &deployment,
@@ -819,7 +818,7 @@ pub fn main() {
                                 Ok(deployment)
                             }
                             None => {
-                                let default_deployment_path = get_default_deployment_path(&manifest, network).unwrap();
+                                let default_deployment_path = project_root.join(get_default_deployment_path(network));
                                 let (deployment, _) =
                                     generate_default_deployment(&manifest, network, false)
                                         .unwrap_or_else(|message| {
@@ -833,8 +832,8 @@ pub fn main() {
                         })
                     }
                     (None, Some(deployment_plan_path)) => {
-                        let deployment_path = get_absolute_deployment_path(&manifest, &deployment_plan_path).expect("unable to retrieve deployment");
-                        load_deployment(&manifest, &deployment_path)
+                        let deployment_path = project_root.join(&deployment_plan_path);
+                        load_deployment(&project_root, &deployment_path)
                     }
                     (_, _) => unreachable!()
                 };
@@ -974,7 +973,7 @@ pub fn main() {
                 let manifest = load_manifest_or_exit(cmd.manifest_path, true);
 
                 let changes = match generate::get_changes_for_new_contract(
-                    &manifest.location,
+                    manifest.location,
                     cmd.name,
                     None,
                     true,
@@ -997,7 +996,7 @@ pub fn main() {
                 let manifest = load_manifest_or_exit(cmd.manifest_path, true);
                 let contract_name = cmd.name.clone();
                 let changes =
-                    match generate::get_changes_for_rm_contract(&manifest.location, cmd.name) {
+                    match generate::get_changes_for_rm_contract(manifest.location, cmd.name) {
                         Ok(changes) => changes,
                         Err(message) => {
                             eprintln!("{}", format_err!(message));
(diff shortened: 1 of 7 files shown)


components/clarity-vscode/server/src/serverBrowser.ts
```diff
@@ -23,7 +23,7 @@ declare const __EXTENSION_URL__: string;
     new BrowserMessageWriter(self),
   );
 
-  initSync(await wasmModule);
+  initSync({ module: await wasmModule });
```
hugo-stacks, line 26: Fix warning (`initSync(module)` is deprecated)

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -507,16 +503,20 @@ impl SDK {
                 return Err(diags_digest.message);
             }
         }
-        let project_root = manifest.location.get_parent_location()?;
+        let project_root = manifest
```
jbencin-stacks: You are calculating `project_root` a second time here. If you need 2 owned copies you should `clone()` instead

hugo-stacks: Good catch, it wasn't needed at all

components/clarinet-files/src/project_manifest.rs
```diff
@@ -205,10 +203,7 @@ impl Serialize for ProjectConfig {
         map.serialize_entry("telemetry", &self.telemetry)?;
         map.serialize_entry(
             "cache_dir",
-            &self
-                .cache_location
-                .get_relative_location()
-                .unwrap_or(self.cache_location.to_string()),
+            &self.cache_location.to_string_lossy().to_string(),
```
jbencin-stacks, line 206: This looks like you are changing from a relative path to an absolute path. Was this intentional?

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1407,80 +1417,81 @@ pub fn load_deployment_and_artifacts_or_exit(
...
                 manifest,
                 &StacksNetwork::Simnet,
                 force_on_disk,
                 force_computed,
-            );
-            match res {
-                Some(Ok(deployment)) => {
+            )
+            .map_err(|e| {
+                format!("loading deployments/default.simnet-plan.yaml failed with error: {e}")
```
jbencin-stacks: I know it's just an error message, but probably better to use `get_default_deployment_path()` than hardcode the path. There is more than one instance in this function

components/clarinet-cli/src/frontend/dap.rs
```diff
@@ -35,9 +35,9 @@ pub fn run_dap() -> Result<(), String> {
 
             for (contract_id, (_, location)) in deployment.contracts.iter() {
                 dap.path_to_contract_id
-                    .insert(PathBuf::from(location.to_string()), contract_id.clone());
+                    .insert(location.clone(), contract_id.clone());
                 dap.contract_id_to_path
-                    .insert(contract_id.clone(), PathBuf::from(location.to_string()));
+                    .insert(contract_id.clone(), location.clone());
```
jbencin-stacks: Change `deployment.contracts.iter()` -> `deployment.contracts` above, and then you can remove the `clone()`s on this line

It's okay to consume `deployment`, it's local variable and you don't return it

components/clarinet-cli/src/generate/mod.rs
```diff
@@ -25,20 +26,20 @@ pub fn get_changes_for_new_project(
 }
 
 pub fn get_changes_for_new_contract(
-    manifest_location: &FileLocation,
+    manifest_location: &Path,
```
jbencin-stacks: As a rule, if you clone or take ownership of something in a function, it's better to accept the owned version as a parameter, so use `PathBuf` here

This let's the user or compiler optimize better, if they have an existing copy and can transfer ownership when they call the function, you won't need an extra allocation

components/clarinet-cli/src/generate/mod.rs
```diff
@@ -25,20 +26,20 @@ pub fn get_changes_for_new_project(
...
 ) -> Result<Vec<Changes>, String> {
     let mut command =
-        GetChangesForNewContract::new(manifest_location.clone(), contract_name, source);
+        GetChangesForNewContract::new(manifest_location.to_path_buf(), contract_name, source);
     command.run(include_test)
 }
 
 pub fn get_changes_for_rm_contract(
-    manifest_location: &FileLocation,
+    manifest_location: &Path,
```
jbencin-stacks: Same as above

components/clarinet-deployments/src/lib.rs
```diff
@@ -490,7 +489,7 @@ pub async fn generate_default_deployment(
 
     // Build the ASTs / DependencySet for requirements - step required for Simnet/Devnet/Testnet/Mainnet
     if let Some(ref requirements) = manifest.project.requirements {
-        let cache_location = &manifest.project.cache_location;
+        let cache_location = manifest.project.cache_location.as_path();
```
jbencin-stacks: This works also

```suggestion
        let cache_location = &manifest.project.cache_location;
```

components/clarinet-files/src/project_manifest.rs
```diff
@@ -313,7 +309,7 @@ impl ProjectManifest {
 
     pub fn from_project_manifest_file(
         project_manifest_file: ProjectManifestFile,
-        manifest_location: &FileLocation,
+        manifest_location: &Path,
```
jbencin-stacks, line 312: Pass `PathBuf` here because this function makes one anyway

components/clarinet-sdk/node/tests/simnet-usage.test.ts
```diff
@@ -643,7 +643,7 @@ describe("the simnet can execute commands", () => {
 //   it("initSimnet handles absolute path", async () => {
 //     const manifestPath = path.join(process.cwd(), "tests/fixtures/Clarinet.toml");
 //     const simnet = await initSimnet(manifestPath);
-//     expect(simnet.blockHeight).toBe(1);
+//     expect(simnet.blockHeight).toBe(3);
```
jbencin-stacks, line 646: This seems unrelated to the PR, is it intentional?

components/clarinet-files/src/project_manifest.rs
hugo-stacks: It triggers this clippy warning
https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#ptr_arg

```
writing `&PathBuf` instead of `&Path` involves a new object where a slice will do
for further information visit https://rust-lang.github.io/rust-clippy/rust-1.93.0/index.html#ptr_arg
`#[warn(clippy::ptr_arg)]` on by defaultclippy[Click for full compiler diagnostic]

project_manifest.rs(312, 28): change this to: `&Path`
```

components/clarinet-cli/src/generate/mod.rs
hugo-stacks: Thanks for the explanation 🙌

components/clarinet-cli/src/frontend/cli.rs
hugo-stacks: `get_default_deployment_path()` performs disk read to get the project root.. So I created `get_default_deployment_relative_path()`

hugo-stacks: Just to realize that whenever we call `get_default_deployment_path()`, we already know the project root. Leading to calling `project_root_from_manifest_location` twice many times.

Again, this is touching a lot of old code, calling for more refactoring, so I'm forcing myself to stop here

hugo-stacks: Ok so I didn't stop, the same thing was happening just below for
```
pub fn load_deployment(
    manifest: &ProjectManifest,
    deployment_plan_location: &Path,
)
```
that I changed into
```
pub fn load_deployment(
    project_root: &Path,
    deployment_plan_path: &Path,
)
```

And there are still many similar changes that could be done. I'll stop here for real, but opened this issue #2198

components/clarinet-sdk/node/tests/simnet-usage.test.ts
hugo-stacks: Not directly related, I tried running this commented test with the changes of this PR, but it still fails on windows.

`3` is the expected result though, so we might as well leave it like that.
I opened #2199

components/clarinet-files/src/project_manifest.rs
hugo-stacks: This keeps whatever the user put (absolute or relative). I think this is reasonable.
What do you think?

jbencin-stacks: Agree

hugo-stacks:
fixes #2179

PR #2183, opened 2026-02-10
refactor: remove forced epoch dead code

hugo-stacks:
### Description

- the `forced_clarity_version` in the `queue` is always set to `None`
- I think it was added a while ago to support the "subnet"
  - it should have been removed as part of #1908 https://github.com/stx-labs/clarinet/pull/1908/changes#diff-d82cc3e5231779103b51f967ee2535530f15937c509c7ceb36b1029506200fd4L423
  
  - also refactored a couple of `match` into `map_err()?`


brady-stacks (approved):
lgtm

PR #2186, opened 2026-02-10
feat(linter): Add `error_const` lint

jbencin-stacks:
### Description

Add `error_const` to linter (#2028)

This lint checks "error constants", and enforces two rules:

  1. All constants prefixed with `ERR_` should have an error type value, `(err ...)`
  2. No two `ERR_` constants should have the same value

#### Breaking change?

No

### Example

Will generate two diagnostic messages on a contract like this:

```clarity
(define-constant ERR_A (err u1))
(define-constant ERR_B u100)     ;; ERR_B: not an error type
(define-constant ERR_C (err u1)) ;; ERR_C: duplicate of ERR_A
```

You can ignore both checks on a specific constant by placing the comment:

```clarity
;; #[allow(error_const)]
```

on the line above it

---

### Checklist

### DIFF
--- a/components/clarity-repl/src/analysis/lints/error_const.rs
+++ b/components/clarity-repl/src/analysis/lints/error_const.rs
@@ -160,7 +160,6 @@ impl<'a, 'b> ErrorConst<'a, 'b> {
             }
         }
 
-        diagnostics.sort_by(|a, b| a.spans[0].cmp(&b.spans[0]));
         diagnostics
     }
 }
--- a/components/clarity-repl/src/analysis/lints/case_const.rs
+++ b/components/clarity-repl/src/analysis/lints/case_const.rs
@@ -106,8 +106,6 @@ impl<'a, 'b> CaseConst<'a, 'b> {
             diagnostics.push(diagnostic);
         }
 
-        // Order the sets by the span of the error (the first diagnostic)
-        diagnostics.sort_by(|a, b| a.spans[0].cmp(&b.spans[0]));
         diagnostics
     }
 }
--- a/components/clarity-repl/src/analysis/lints/unused_const.rs
+++ b/components/clarity-repl/src/analysis/lints/unused_const.rs
@@ -91,8 +91,6 @@ impl<'a, 'b> UnusedConst<'a, 'b> {
             diagnostics.push(diagnostic);
         }
 
-        // Order the sets by the span of the error (the first diagnostic)
-        diagnostics.sort_by(|a, b| a.spans[0].cmp(&b.spans[0]));
         diagnostics
     }
 }


components/clarity-repl/src/analysis/lints/error_const.rs
```diff
@@ -0,0 +1,369 @@
...
+            .and_then(|atom| NativeFunctions::lookup_by_name_at_version(atom, clarity_version))
+            .is_some_and(|func| func == NativeFunctions::ConsError)
+    }
+
+    /// Get the inner value of an `(err X)` expression (returns `X`)
+    fn get_err_inner_value(value: &SymbolicExpression) -> Option<&SymbolicExpression> {
+        value.match_list().and_then(|list| list.get(1))
+    }
+
+    fn make_not_err_message(name: &ClarityName) -> String {
```
brady-stacks, line 89: (feel free to ignore) These error message functions could probably be an enum (naming would be hard, e.g. ErrorConstError 😆), like collecting all the error messages in one place and stringifying them in the Diagnostic message

components/clarity-repl/src/analysis/lints/error_const.rs
```diff
@@ -0,0 +1,369 @@
...
+const ERR_PREFIX: &str = "ERR_";
+
+use crate::analysis::annotation::{Annotation, AnnotationKind, WarningKind};
+use crate::analysis::cache::constants::ConstantData;
+use crate::analysis::cache::AnalysisCache;
+use crate::analysis::linter::Lint;
+use crate::analysis::{self, AnalysisPass, AnalysisResult, LintName};
+
+struct ErrorConstSettings {
+    // TODO
```
hugo-stacks, line 24: Is this a "TODO" to do before merge, or later?

components/clarity-repl/src/analysis/lints/error_const.rs
```diff
@@ -0,0 +1,369 @@
...
+                    message: Self::make_not_err_message(name),
+                    spans: vec![const_data.expr.span.clone()],
+                    suggestion: Some(format!(
+                        "Use `(err ...)` as the value or remove the `{ERR_PREFIX}` prefix"
+                    )),
+                });
+                continue;
+            }
+
+            // Check 2: No duplicate error values
```
hugo-stacks, line 136: Should we also recommend to only use uint as error code since it seems to be the norm?
Or would that be an other rule?

components/clarity-repl/src/analysis/lints/error_const.rs
jbencin-stacks: There's an empty settings struct in all the lints. Was going to use to implement per-lint settings but never had  any to implement

components/clarity-repl/src/analysis/lints/error_const.rs
jbencin-stacks: If we do, it should go into this lint for sure. Yeah I've only seen `uint`, but I could see someone using `int` or even strings for readability. Honestly, strings seem like a good idea

components/clarity-repl/src/analysis/lints/error_const.rs
jbencin-stacks: Not a bad idea, and would be useful if I was passing the errors around, but as it is I'd basically just construct the enum variant and immediately `.to_string()` it to stuff it into a `Diagnostic`

components/clarity-repl/src/analysis/lints/error_const.rs
jbencin-stacks: I'll ask on Slack to see what others think

jbencin-stacks: So some disagreement on this, I'm going to merge what I have in this PR and may do another PR to change the rule later

PR #2189, opened 2026-02-11
Run diagnostics with and without `env(simnet)` code during CLI `clarinet check` and LSP

### LINKED ISSUE #2022: Conditionally deploy code in Simnet

xoloki-stacks:
### Description

This PR closes #2022 by running diagnostics twice, with and without the `env(simnet)` code.  This will show if the code can be deployed outside of `simnet`.

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -1568,8 +1568,7 @@ pub fn load_deployment_and_artifacts_or_exit(
                 if force_remove_env_simnet {
                     found_env_simnet |= deployment.remove_env_simnet();
                 }
-                let artifacts =
-                    setup_session_with_deployment(manifest, &mut deployment, None);
+                let artifacts = setup_session_with_deployment(manifest, &mut deployment, None);
                 (
                     (
                         deployment,
@@ -1580,7 +1579,10 @@ pub fn load_deployment_and_artifacts_or_exit(
                 )
             }
             Err(e) => {
-                eprintln!("{}", format_err!(format!("loading {path} failed with error: {e}")));
+                eprintln!(
+                    "{}",
+                    format_err!(format!("loading {path} failed with error: {e}"))
+                );
                 process::exit(1);
             }
         }
@@ -1612,11 +1614,8 @@ pub fn load_deployment_and_artifacts_or_exit(
             };
 
         if ast_artifacts.success {
-            let mut artifacts = setup_session_with_deployment(
-                manifest,
-                &mut deployment,
-                Some(&ast_artifacts.asts),
-            );
+            let mut artifacts =
+                setup_session_with_deployment(manifest, &mut deployment, Some(&ast_artifacts.asts));
             for (contract_id, mut parser_diags) in ast_artifacts.diags.into_iter() {
                 if let Some(ref mut diags) = artifacts.diags.remove(&contract_id) {
                     parser_diags.append(diags);


brady-stacks (dismissed):
I wonder if it would be better to expose a flag (something like clarinet check --simnet ) ? Do we care about running check multiple times implicitly? 

(fine with merging this regardless, just curious how you're thinking about it)

xoloki-stacks:
> Do we care about running check multiple times implicitly?

Code updated to only run the second time if the annotation was found the first time, and code was removed.

xoloki-stacks:
Converting to draft since the LSP tests are in heavy flux

xoloki-stacks:
Closing since we did this is a cleaner way in #2252

PR #2197, opened 2026-02-12
chore(clarity-repl): Clean up multi-line strings with `indoc`

jbencin-stacks:
### Description

Clean up all the multi-line strings in `clarity-repl` with `indoc` so that the string indentation matches the code indentation

### DIFF
--- a/components/clarity-repl/src/analysis/check_checker/mod.rs
+++ b/components/clarity-repl/src/analysis/check_checker/mod.rs
@@ -920,8 +920,7 @@ mod tests {
                     (var-set p1 p)
                     (var-set b1 b)
                 )
-            )
-        ").to_string();
+            )        ").to_string();
         match session.formatted_interpretation(snippet, Some("checker".to_string()), false, None) {
             Ok((output, result)) => {
                 assert_eq!(result.diagnostics.len(), 2);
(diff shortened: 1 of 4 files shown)


components/clarity-repl/src/analysis/check_checker/mod.rs
```diff
@@ -908,25 +910,25 @@ mod tests {
...
-;; #[allow(unchecked_params)]
-(define-private (my-func-p (p principal) (b bool))
-    (begin
-        (var-set p1 p)
-        (var-set b1 b)
-    )
-)"
-        .to_string();
+        #[rustfmt::skip]
+        let snippet = indoc!("
```
hugo-stacks, line 914: @jbencin-stacks When is the `indoc!()` notation preferred over `indoc! {}`?

jbencin-stacks: They do the same thing, it's just how `rustfmt` handles them

- `indoc!()`: Need to add `#[rustfmt::skip]` on the line before or `rustfmt` will mangle it
- `indoc!{}`: Don't need to add `#[rustfmt::skip]` on the line before, but then `rustfmt` will move the `.to_string()` after it to a new line

So it's kind of annoying either way but way better than the unindented strings

hugo-stacks: Got it

PR #2201, opened 2026-02-13
test: Add more unit tests for Clarity linter

jbencin-stacks:
### Description

I asked Claude to identify gaps in unit test coverage for the linter, and it generated several test cases which are useful and valid tests

PR also contains a small refactor in `unused_map.rs`

### DIFF
--- a/components/clarity-repl/src/analysis/lints/unused_map.rs
+++ b/components/clarity-repl/src/analysis/lints/unused_map.rs
@@ -38,7 +38,7 @@ impl From<&MapData<'_>> for Usage {
         } else if !map_data.map_get && !map_data.map_set {
             Self::Unused
         } else {
-            unreachable!("Unhandled usage pattern: {:?}", map_data)
+            unreachable!("Unhandled usage pattern: {map_data:?}")
         }
     }
 }


components/clarity-repl/src/analysis/lints/unused_map.rs
```diff
@@ -27,6 +27,22 @@ enum Usage {
...
+        if map_data.is_used() {
+            Self::Used
+        } else if map_data.map_get && !map_data.map_set {
+            Self::Unset
+        } else if !map_data.map_get && map_data.map_set {
+            Self::Unread
+        } else if !map_data.map_get && !map_data.map_set {
+            Self::Unused
+        } else {
+            unreachable!("Unhandled usage pattern: {map_data:?}")
```
brady-stacks, line 41: I got sniped by this again 😆 
Tried to re-arrange it so there's no unreachable but... This is probably good enough. My idea was to use match
```
match (map_data.map_get, map_data.mep_set) {
  (true, false) => Self::Unset,
  (false, true) => Self::Unread,
  (false, false) => Self::Unused,
  (true, true) => if map_data.is_used() { Self::Used } else { unreachable }
}
```

It doesn't feel much better tbh

jbencin-stacks: I think I originally had a `match` statement like that and we changed it. IDK if there's any nice, clean way to do it with these particular conditions, but we could try asking Claude

PR #2206, opened 2026-02-16
feat(linter): Find more noops in `noop` lint

jbencin-stacks:
### Description

Claude and I identified some additional noops that were missing from the current lint. Here is the full list that should be detected now:

- Operations that always return `true` or `false`
   - Comparison ops
     - `(is-eq x)` -> `true`
   - Logical ops
     - `(and ... false)` -> `false`
     - `(or ... true)` -> `true`
   - Arithmetic ops
     - `(* ... 0)` -> `0`
     - `(/ 0 ...)` -> `0`
     - `(/ ... 0)` -> Runtime error
- Operations that always return one of the operands
  - Comparison ops
    - `(is-eq x true)` -> `x`
    - `(is-eq x false)` -> `(not x)`
  - Logical ops
    - `(not (not x))` -> `x`
    - `(and x)`, `(and x true)`, etc. -> `x`
    - `(or x)`, `(or x false)`, etc. -> `x`
  - Arithmetic ops
    - `(+ x)`, `(+ x u0)`, `(+ x 1 -1)`, etc. -> `x`
    - `(- x)`, `(- x u0)`, `(- x 1 -1)`, etc. -> `x`
    - `(* x)`, `(* x u1)`, `(* x u1 u1)`, etc. -> `x`
    - `(/ x)`, `(/ x u1)`, `(/ x u1 u1)`, etc. -> `x`

> [!NOTE]
> I also want to add a lint to check for constant expressions like `(+ u1 u2)`. I'd like to make it a separate lint, `const_expr`, but this lint basically detects these cases already so might add it here.
>
> I might refactor this into one or more lints since it's kind of big, but only if I can split stuff up in a way that makes logical sense and doesn't require recomputing the same stuff in both lints

### Checklist

### DIFF
--- a/components/clarity-repl/src/analysis/lints/noop.rs
+++ b/components/clarity-repl/src/analysis/lints/noop.rs
@@ -59,9 +59,9 @@ enum ClaritySum {
     Signed(i128),
     /// Clarity `uint`
     Unsigned(u128),
-    /// Tried to mix `uint` and `int` or used non-integer type
+    /// Type error from mixing `uint` and `int` or using non-integer type
     /// Should not be possible, such cases should fail to parse
-    Mixed,
+    Error,
 }
 
 /// Result of separating operands into constants and non-constants,
@@ -86,16 +86,16 @@ impl<'a> SplitOperands<'a> {
                     acc = match acc {
                         None => Some(ClaritySum::Unsigned(*n)),
                         Some(ClaritySum::Unsigned(a)) => uint_op(a, *n).map(ClaritySum::Unsigned),
-                        Some(ClaritySum::Signed(_)) => Some(ClaritySum::Mixed),
-                        Some(ClaritySum::Mixed) => acc,
+                        Some(ClaritySum::Signed(_)) => Some(ClaritySum::Error),
+                        Some(ClaritySum::Error) => acc,
                     };
                 }
                 Some(Value::Int(n)) => {
                     acc = match acc {
                         None => Some(ClaritySum::Signed(*n)),
                         Some(ClaritySum::Signed(a)) => int_op(a, *n).map(ClaritySum::Signed),
-                        Some(ClaritySum::Unsigned(_)) => Some(ClaritySum::Mixed),
-                        Some(ClaritySum::Mixed) => acc,
+                        Some(ClaritySum::Unsigned(_)) => Some(ClaritySum::Error),
+                        Some(ClaritySum::Error) => acc,
                     };
                 }
                 _ => non_constants.push(op),
@@ -136,7 +136,7 @@ impl<'a> SplitOperands<'a> {
         self.acc.as_ref().and_then(|val| match val {
             ClaritySum::Unsigned(0) => Some("u0"),
             ClaritySum::Signed(0) => Some("0"),
-            ClaritySum::Unsigned(_) | ClaritySum::Signed(_) | ClaritySum::Mixed => None,
+            ClaritySum::Unsigned(_) | ClaritySum::Signed(_) | ClaritySum::Error => None,
         })
     }
 }


components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -77,43 +172,370 @@ impl<'a> NoopChecker<'a> {
...
+    /// Format a SymbolicExpression back to Clarity source for use in suggestion text
+    fn format_source(expr: &SymbolicExpression) -> String {
+        match &expr.expr {
+            SymbolicExpressionType::Atom(name) => name.to_string(),
+            SymbolicExpressionType::LiteralValue(v) | SymbolicExpressionType::AtomValue(v) => {
+                match v {
+                    Value::Int(n) => format!("{n}"),
+                    Value::UInt(n) => format!("u{n}"),
+                    Value::Bool(b) => format!("{b}"),
+                    _ => "...".to_string(),
```
hugo-stacks: nit: can call `v.to_string()` here

components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -77,43 +172,370 @@ impl<'a> NoopChecker<'a> {
...
+        &mut self,
+        expr: &'a SymbolicExpression,
+        operands: &'a [SymbolicExpression],
+    ) {
         if operands.len() < 2 {
-            self.add_noop_diagnostic(
+            return;
+        }
+
+        // TODO: What if first arg is a constant?
```
hugo-stacks: Is it a TODO for this PR or for later? Can you open a matching an issue if this is meant to be merged?

components/clarity-repl/src/repl/interpreter.rs
```diff
@@ -1377,11 +1377,13 @@ mod tests {
         assert!(result.is_err());
 
         let diagnostics = result.unwrap_err();
-        assert_eq!(diagnostics.len(), 1);
+        // Expect error message from linter
+        // TODO: Disable linter for all unit tests in this file
```
hugo-stacks, line 1381: Is it a TODO for this PR or for later? Can you open a matching an issue if this is meant to be merged?

components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -77,43 +172,370 @@ impl<'a> NoopChecker<'a> {
...
+        let (bool_idx, other_idx) = if Self::as_bool_literal(&operands[0]).is_some() {
+            (0, 1)
+        } else if Self::as_bool_literal(&operands[1]).is_some() {
+            (1, 0)
+        } else {
+            return;
+        };
+
+        let b = Self::as_bool_literal(&operands[bool_idx]).unwrap();
+        let other = Self::format_source(&operands[other_idx]);
```
hugo-stacks: `other` can be move to line 309 (after the early return line 307)

components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -77,43 +172,370 @@ impl<'a> NoopChecker<'a> {
...
+    fn check_bool_comparison(
+        &mut self,
+        expr: &'a SymbolicExpression,
+        operands: &'a [SymbolicExpression],
+    ) {
+        if operands.len() != 2 {
+            return;
+        }
+
+        let (bool_idx, other_idx) = if Self::as_bool_literal(&operands[0]).is_some() {
```
hugo-stacks: nit: when dealing with such expressions, I like calling expressions `left` and `right`. I find `other` and `bool` or `b` a bit opaque

```suggestion
        let (left_idx, right_idx) = if Self::as_bool_literal(&operands[0]).is_some() {
```

components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -77,43 +172,370 @@ impl<'a> NoopChecker<'a> {
...
+        } else {
+            return;
+        };
+
+        let b = Self::as_bool_literal(&operands[bool_idx]).unwrap();
+        let other = Self::format_source(&operands[other_idx]);
+
+        // Check if both operands are bool literals
+        if Self::as_bool_literal(&operands[other_idx]).is_some() {
+            let result = Self::as_bool_literal(&operands[other_idx]).unwrap() == b;
```
hugo-stacks: Can be a `if let` to avoid the redundant cal `Self::as_bool_literal(&operands[other_idx])`

components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -138,8 +592,32 @@ impl<'a> ASTVisitor<'a> for NoopChecker<'a> {
...
+            let message = format!("`{func}` with fewer than 2 operands has no effect",);
+            let suggestion = format!("Replace with `{}`", Self::format_source(&operands[0]));
+            self.add_diagnostic(expr, message, suggestion);
+            return true;
+        } else {
+            // Check for identity elements in arithmetic ops
+            self.check_arithmetic_identity(expr, func, operands);
+        }
+
+        true
```
hugo-stacks: I think the `else` isn' necessary after the early return

```suggestion
        // Check for identity elements in arithmetic ops
        self.check_arithmetic_identity(expr, func, operands);
        true
```

components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -138,8 +592,32 @@ impl<'a> ASTVisitor<'a> for NoopChecker<'a> {
...
     ) -> bool {
-        if !Self::is_single_op_arithmetic(func) {
-            self.check_bin_op(expr, func, operands);
+        if Self::is_single_op_arithmetic(func) || self.check_allowed(&expr.span) {
+            return true;
+        }
+
+        if operands.len() < 2 {
+            // Variable length arithmetic ops with single operand have no effect
+            let message = format!("`{func}` with fewer than 2 operands has no effect",);
```
hugo-stacks:
```suggestion
            let message = format!("`{func}` with fewer than 2 operands has no effect");
```

components/clarity-repl/src/analysis/lints/noop.rs
```diff
@@ -77,43 +172,370 @@ impl<'a> NoopChecker<'a> {
...
+        }
+    }
+
+    /// Check for double negation: `(not (not x))`
+    fn check_double_negation(
+        &mut self,
+        expr: &'a SymbolicExpression,
+        operands: &'a [SymbolicExpression],
+    ) {
+        if operands.len() == 1 && Self::is_not_call(&operands[0]) {
```
hugo-stacks: I think that, because `is_not_call` is only called by `check_double_negation`, it could be refactored to return the inner expression, avoiding to  destructure the expression twice:

```rs
    fn is_not_call(expr: &SymbolicExpression) -> Option<&SymbolicExpression> {
        let SymbolicExpressionType::List(exprs) = &expr.expr else {
            return None;
        };
        let func = exprs.first()?.match_atom()?;
        matches!(
            NativeFunctions::lookup_by_name(func),
            Some(NativeFunctions::Not)
        )
        .then(|| exprs.get(1))
    }
```

Could be renamed as `get_not_inner_expr()`, `as_not_inner()`, or something like that. 

And in `check_double_negation`
```rs
        if operands.len() != 1 {
            return;
        }
        if let Some(inner) = Self::is_not_call(&operands[0]) {
            let suggestion = format!("Replace with `{}`", Self::format_source(inner));
            ...
```

components/clarity-repl/src/repl/interpreter.rs
jbencin-stacks: Sometimes adding these lints causes new errors to be found in these code snippets, which messes up the `assert!()` statements

The linter should be disabled for ALL unit tests, made issue #2213 for this

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: [31c15f7](https://github.com/stx-labs/clarinet/pull/2206/commits/31c15f722f46e3df233785d9682ebf4aedbcfa4f)

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: [31c15f7](https://github.com/stx-labs/clarinet/pull/2206/commits/31c15f722f46e3df233785d9682ebf4aedbcfa4f)

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: [31c15f7](https://github.com/stx-labs/clarinet/pull/2206/commits/31c15f722f46e3df233785d9682ebf4aedbcfa4f)

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: [31c15f7](https://github.com/stx-labs/clarinet/pull/2206/commits/31c15f722f46e3df233785d9682ebf4aedbcfa4f)

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: [31c15f7](https://github.com/stx-labs/clarinet/pull/2206/commits/31c15f722f46e3df233785d9682ebf4aedbcfa4f)

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: [31c15f7](https://github.com/stx-labs/clarinet/pull/2206/commits/31c15f722f46e3df233785d9682ebf4aedbcfa4f)

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: Sure. I think also the `operands.len() != 1` isn't necessary as something like `(not x y)` will fail to parse

jbencin-stacks: [913ee1a](https://github.com/stx-labs/clarinet/pull/2206/commits/913ee1a898110225715c47d0d536a3a3af5eefba)

hugo-stacks (dismissed):
lgtm!

hugo-stacks:
@jbencin-stacks 
A note on splitting into multiple lints:

re:
> I might refactor this into one or more lints since it's kind of big, but only if I can split stuff up in a way that makes logical sense and doesn't require recomputing the same stuff in both lints

I it makes sense, I think the `is-eq` checks (eg: `(is-eq false <expr>)`-> `(not <expr>)` ) could be in their own lints. Mostly because `(is-eq false <expr>)` is _not_ a noop. Replacing it with `not` is more of a best practice imo

components/clarity-repl/src/analysis/lints/noop.rs
jbencin-stacks: I don't need to address this case, it wouldn't be a noop unless *all* the args were constant (and then it would trigger the still-to-do `const_expr` lint)

PR #2207, opened 2026-02-16
chore: update stacks-core/develop

brady-stacks:
gives us access to the new min_size functions in core for use in the static cost analysis

adds --epoch flag to cli and the correct contract epoch in LSP for the new depth limit in clarity parser

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -27,13 +27,13 @@ use clarity_repl::analysis::call_checker::ContractAnalysis;
 use clarity_repl::clarity::vm::analysis::AnalysisDatabase;
 use clarity_repl::clarity::vm::costs::LimitedCostTracker;
 use clarity_repl::clarity::vm::types::QualifiedContractIdentifier;
-use clarity_repl::clarity::ClarityVersion;
+use clarity_repl::clarity::{ClarityVersion, StacksEpochId};
 use clarity_repl::frontend::Terminal;
 use clarity_repl::repl::diagnostic::output_diagnostic;
 use clarity_repl::repl::settings::{ApiUrl, RemoteDataSettings};
 use clarity_repl::repl::{
-    clarity_version_to_u8, epoch_from_str, ClarityCodeSource, ClarityContract, ContractDeployer,
-    Epoch, DEFAULT_EPOCH,
+    clarity_version_to_u8, ClarityCodeSource, ClarityContract, ContractDeployer, Epoch,
+    DEFAULT_EPOCH,
 };
 use clarity_repl::{analysis, repl};
 use stacks_network::{self, DevnetOrchestrator};
@@ -106,8 +106,6 @@ struct Formatter {
     pub file: Option<String>,
     #[clap(long = "max-line-length", short = 'l')]
     pub max_line_length: Option<usize>,
-    #[clap(long = "epoch")]
-    pub epoch: Option<String>,
     #[clap(long = "indent", short = 'i', conflicts_with = "use_tabs")]
     /// indentation size, e.g. 2
     pub indentation: Option<usize>,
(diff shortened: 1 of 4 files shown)


brady-stacks:
@hugo-stacks for formatter related parsing I'm not particularly worried. You think it's worth building in some epoch passing? formatter doesn't care about it currently, but we could certainly add it

brady-stacks:
@hugo-stacks Alright, added the epoch passing 👍

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1281,6 +1280,12 @@ pub fn main() {
             if cmd.use_tabs {
                 settings.indentation = formatter::Indentation::Tab;
             }
+            if let Some(epoch) = cmd.epoch {
```
hugo-stacks: If the sources is loaded from the manifest, the epoch from each contract should be the one set in the manifest.
So in way, you only want `cmd.epoch` along with `cmd.file`.

Maybe it could me somehow handled in the same way as clarinet check that has
```rust
Command::Check(cmd) if cmd.file.is_some() => {}
Command::Check(cmd) => {}
```

--- 

I would also be ok to just not accept formatting src a file if it's not associated to the manifest - We could discuss it

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -40,13 +42,19 @@ const FORMAT_IGNORE_SYNTAX: &str = "@format-ignore";
 pub struct Settings {
     pub indentation: Indentation,
     pub max_line_length: usize,
+    pub epoch: Option<StacksEpochId>,
```
hugo-stacks: The `epoch` should be passed as an argument in `format()`, it can't be set in the settings. The same formatter instance can be used to format file with different epochs

hugo-stacks (changes requested):
@brady-stacks I left two comments that are basically related to the same thing

components/clarinet-cli/src/frontend/cli.rs
brady-stacks: 👌 That's what I ended up doing. Works the same as Check

components/clarinet-format/src/formatter/mod.rs
```diff
@@ -83,12 +89,20 @@ impl ClarityFormatter {
...
-    pub fn format_section(&self, source: &str) -> Result<String, String> {
-        let pse = clarity::vm::ast::parser::v2::parse(source).map_err(|e| e.to_string())?;
+    pub fn format_section(
+        &self,
+        source: &str,
+        epoch: Option<StacksEpochId>,
+    ) -> Result<String, String> {
+        let pse = clarity::vm::ast::parser::v2::parse(
+            source,
+            StackDepthLimits::for_epoch(epoch.unwrap_or(StacksEpochId::latest())),
```
hugo-stacks, line 103: Nit: we usually prefer using Clarinet `DEFAULT_EPOCH` to have control over the epoch. This is especially useful in period of transitions (like the one we are experiencing right now), where Clarinet should behave as if the current/latest epoch _is_ 3.3, because this is the epoch on Mainnet. But because we have early support for 3.4 and Clarity 5, `StacksEpochId::latest()` is 3.4 right now, which is not really what we want.

I'm marking this as a nit because it's likely not gonna cause any issue, and it would be annoying for `clarinet-format` to depend on `clarity_repl::repl::DEFAULT_EPOCH`. But if you can find an a way to id it easily/cleanly that would be ideal. Otherwise, let's just merge 👍

brady-stacks: I think we would have to define DEFAULT_EPOCH elsewhere. I can open an issue. It would probably be beneficial in general to have these constants defined in a shared location that every module can access without loading the whole repl module...

Issue #2198, opened 2026-02-13
Improve `project_root` handling


hugo-stacks:
The work on #2182 highlighted that the `project_root` is computed many times.
There might be room for more improvements here (possibly simply in `ProjectConfig` that already holds the manifest location)

PR #2209, opened 2026-02-17
refactor: project root and manifest path handling

hugo-stacks:
### Description
Fixes: #2198

Get the `project_root` directory only once and save it in the manifest. Use it later instead of calling 
```rust
project_root_from_manifest_location()
```

### Initial and remaining issues

- `project_root_from_manifest_location()` was doing a lot of disk access for nothing - this PR fixes that

- Clarinet let's users specify a custom manifest path in, but then tries to find a `Clarinet.toml` in `project_root_from_manifest_location()`, so certain commands were actually broken when a custom manifest path was given (`clarinet contract add -m CustomManifest.toml` for instance). In which case, just calling `manifest.parent()` is actually more robust.

We might still need to improve custom manifests handling, but I think the PR fixes a few issues already

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -1165,7 +1165,13 @@ pub fn main() {
         }
         Command::Check(cmd) if cmd.file.is_some() => {
             let file = cmd.file.unwrap();
-            let mut settings = repl::SessionSettings::default();
+            let mut settings = repl::SessionSettings {
+                repl_settings: repl::Settings {
+                    analysis: analysis::Settings::with_default_lints(),
+                    ..Default::default()
+                },
+                ..Default::default()
+            };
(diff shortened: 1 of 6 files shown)


components/clarinet-cli/src/deployments/mod.rs
```diff
@@ -17,45 +17,33 @@ pub fn generate_default_deployment(
...
+    let deployment_files_paths: Vec<PathBuf> = entries
+        .filter_map(Result::ok)
+        .map(|entry| entry.path())
+        .filter(|p| {
+            matches!(
+                p.extension()
+                    .and_then(|e| e.to_str())
+                    .map(str::to_ascii_lowercase)
+                    .as_deref(),
+                Some("yml") | Some("yaml")
```
xoloki-stacks, line 33: I didn't realize this was possible in a `matches!` statement

PR #2210, opened 2026-02-17
feat: static cost analysis

brady-stacks:
- [x] gives min/max execution cost without executing the contracts
- [x] gives trait call-counts to determine if excessive cost is incurred by trait calls

migrated from [here](https://github.com/stacks-network/stacks-core/pull/6704)

depends on #2207


components/clarity-static-cost/src/static_cost/cost_functions.rs
```diff
@@ -0,0 +1,167 @@
...
+            | StacksEpochId::Epoch22
+            | StacksEpochId::Epoch23
+            | StacksEpochId::Epoch24
+            | StacksEpochId::Epoch25
+            | StacksEpochId::Epoch30
+            | StacksEpochId::Epoch31
+            | StacksEpochId::Epoch32 => self.eval::<Costs3>(n),
+            StacksEpochId::Epoch33 => self.eval::<Costs4>(n),
+            StacksEpochId::Epoch34 => self.eval::<Costs4>(n),
+            StacksEpochId::Epoch10 => self.eval::<Costs1>(n),
```
hugo-stacks: Can you put it at the top? Also, it can also be unreachable as it is here: https://github.com/jbencin-stacks/clarinet/blob/8f7322899fa456df3707a0c99703051c701de413/components/clarinet-deployments/src/types.rs#L65-L66

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1618 @@
...
+        assert!(matches!(right_val.expr, CostExprNode::AtomValue(_)));
+
+        let ok_true_node = &cost_tree.children[2];
+        assert!(matches!(
+            ok_true_node.expr,
+            CostExprNode::NativeFunction(NativeFunctions::ConsOkay)
+        ));
+        assert_eq!(ok_true_node.children.len(), 1);
+
+        let ok_false_node = &cost_tree.children[2];
```
jbencin-stacks, line 1516: Copy/paste error? `ok_false_node` is the same value as `ok_true_node`

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1618 @@
...
+
+    let epoch = env.global_context.epoch_id;
+    let ast = make_ast(&contract_source, epoch, clarity_version)?;
+
+    static_cost_tree_from_ast(&ast, clarity_version, epoch, env)
+}
+
+/// same idea as `static_cost` but returns the root of the cost analysis tree for each function
+/// Useful if you need to analyze specific nodes in the cost tree
+pub fn static_cost_tree(
```
jbencin-stacks: This function is identical to `static_cost`, is that intentional?

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1618 @@
...
+            let mut static_cost: StaticCost = summing_cost.into();
+
+            let overhead = compute_function_overhead_costs(
+                contract_size,
+                name,
+                &contract_ast.expressions,
+                epoch,
+            );
+            static_cost.min.add(&overhead.0).ok()?;
+            static_cost.max.add(&overhead.1).ok()?;
```
jbencin-stacks: This fails silently on overflow. You should saturate to `u64::MAX` instead

This occurs in other places as well

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1618 @@
...
+                            // Fallback: infer type from literal value if type_map is not available
+                            // This handles the case where type_map wasn't persisted to the database
+                            infer_type_from_expression(&binding_pair[1], epoch).ok()
+                        });
+
+                    if let Some(binding_type) = binding_type {
+                        extended_user_args.add_argument(var_name.clone(), binding_type);
+                    }
+                    // If we can't determine the type, skip adding it to the context
+                    // This will result in a conservative cost estimate (zero cost for lookups)
```
jbencin-stacks: In this context, "conservative" should mean worst case scenario, rather than optimistic (zero cost), right?

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,849 @@
...
+        .match_atom_value()
+        .or_else(|| tuple_expr.match_literal_value())
+    {
+        if let Ok(size) = literal_value.size() {
+            let s = size as u64;
+            return (s, s);
+        }
+    }
+
+    (0, 0)
```
jbencin-stacks, line 213: Silently return `0` on failure?

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -0,0 +1,344 @@
+// Library API - items will be used when integrated with LSP, CLI, etc.
+#![allow(dead_code)]
```
jbencin-stacks: Probably shouldn't silence warnings for entire module, just on the public functions/types

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1618 @@
...
+///    - Charged when applying a user-defined function (not native functions)
+///
+/// 3. **Inner Type Check Cost** (`cost_inner_type_check_cost`):
+///    - Charged when type checking the arguments of a user-defined function
+fn compute_function_overhead_costs(
+    contract_size: Option<u64>,
+    function_name: &str,
+    ast_expressions: &[SymbolicExpression],
+    epoch: StacksEpochId,
+) -> (ExecutionCost, ExecutionCost) {
```
jbencin-stacks: Should this return `StaticCost` instead?

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,849 @@
...
+use super::cost_analysis::{StaticCost, UserArgumentsContext};
+use super::cost_functions::{from_native_function, ClarityCostFunctionExt};
+
+// Constants for tuple serialization overhead
+const TUPLE_LENGTH_ENCODING_BYTES: u64 = 4;
+const TUPLE_FIELD_OVERHEAD_BYTES: u64 = 2;
+
+/// Get the serialized size of a reserved variable based on its type
+fn get_reserved_variable_size(native_var: NativeVariables) -> Option<(u64, u64)> {
+    use clarity::vm::variables::NativeVariables;
```
jbencin-stacks: Already imported at the file level

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,849 @@
...
+}
+
+/// Infer the type of a field value from a tuple binding
+/// Checks reserved variables, user arguments, and falls back to type inference
+fn infer_field_type_from_binding(
+    binding_pair: &[SymbolicExpression],
+    epoch: StacksEpochId,
+    user_args: Option<&UserArgumentsContext>,
+) -> Option<TypeSignature> {
+    use clarity::vm::variables::NativeVariables;
```
jbencin-stacks: Already imported at the file level

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1618 @@
...
+    }
+
+    /// minimum cost across all paths
+    pub fn min(&self) -> ExecutionCost {
+        if self.costs.is_empty() {
+            ExecutionCost::ZERO
+        } else {
+            self.costs
+                .iter()
+                .fold(self.costs[0].clone(), |acc, cost| ExecutionCost {
```
jbencin-stacks: This adds `self.costs[0]` twice. Need to add `.skip(1)`, or even better, could simplify to `.reduce()`, or implement `Sum` trait and use `.sum()`

Same for `max()`

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -0,0 +1,344 @@
...
+use stacks_common::types::StacksEpochId;
+pub use trait_counter::{
+    TraitCount, TraitCountCollector, TraitCountContext, TraitCountPropagator, TraitCountVisitor,
+};
+
+const STRING_COST_BASE: u64 = 36;
+const STRING_COST_MULTIPLIER: u64 = 3;
+
+pub(crate) fn calculate_function_cost(
+    function_name: String,
```
jbencin-stacks: Could be an `&str`

```suggestion
    function_name: &str,
```

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1618 @@
...
+                    epoch,
+                    env,
+                    let_depth,
+                )?;
+                children.push(child_tree);
+            }
+            // Add the nested tree as a child (its cost will be included when summing children)
+            children.insert(0, nested_tree);
+            // The root cost is zero - the actual cost comes from the nested expression
+            let expr_node = CostExprNode::Atom(ClarityName::from("nested-expression"));
```
jbencin-stacks: How about a `CostExprNode::NestedExpression` variant instead of a sentinel value

components/clarity-static-cost/src/static_cost/cost_analysis.rs
brady-stacks: `static_cost` should be returning the `StaticCost` instead of the root node 👍 changed

components/clarity-static-cost/src/static_cost/cost_analysis.rs
brady-stacks: Yes, if it can't determine the binding it means it's probably something that is impossible to let-bind. Updated the comment
```
                    // If we can't determine the type (e.g. the binding value is a function call
                    // or variable reference without a type sig), the variable is
                    // omitted from the context. Any lookup of that variable will fall back to
                    // ZERO but let bindings should not be able to use function
                    // calls or variables with no binding type.
```

components/clarity-static-cost/src/static_cost/special_costs.rs
brady-stacks: This probably requires more digging. I think it will miss let-bound variables, contract data variables, and stuff that uses a function call directly as the map key/value arg. I updated the comment as a TODO and I'll try to drill down on this..

components/clarity-static-cost/src/static_cost/cost_analysis.rs
jbencin-stacks: If we have what we think is illegal Clarity code, I'd recommend doing

```rust
panic!("Illegal expression: {expr}");
```

instead

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,861 @@
...
+                return (expr_min, expr_max);
+            }
+        }
+    }
+
+    // Try atom (variable reference) via user_args
+    if let Some(atom_name) = tuple_expr.match_atom() {
+        if let Some(type_sig) = user_args.and_then(|ua| ua.get_argument_type(atom_name)) {
+            let min = type_sig.min_size().unwrap_or(0) as u64;
+            let max = type_sig.size().unwrap_or(0) as u64;
```
jbencin-stacks: Try to avoid `as` since it can silently truncate. Not a problem here specifically, but just better not to have it in the codebase

```suggestion
            let min = u64::from(type_sig.min_size().unwrap_or(0));
            let max = u64::from(type_sig.size().unwrap_or(0));
```

jbencin-stacks (commented):
I looked over the recent changes today and they look great! I will do a full re-review later but looking good so far

components/clarity-static-cost/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+edition = "2021"
+description = "Static cost analysis for Clarity smart contracts"
+
+[dependencies]
+thiserror = "2"
+clarity = { workspace = true, default-features = false, features = [
+    "developer-mode",
+    "devtools",
+    "rusqlite",
+    "testing",
```
hugo-stacks, line 13: Do we need `testing` here?
Clarity uses an old version of `rstest`

components/clarity-static-cost/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+] }
+clarity-types = { workspace = true }
+stacks-common = { workspace = true, default-features = false }
+
+[dev-dependencies]
+stackslib = { workspace = true, default-features = false }
+rstest = "0.21"
```
hugo-stacks: Latest version of rstest is 0.26, any reason to use 0.21?

brady-stacks: I think both this and the testing dep are copy pastes from the stacks-core. I'll update these 👍

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -0,0 +1,341 @@
...
+pub(crate) fn calculate_function_cost(
+    function_name: &str,
+    cost_map: &HashMap<String, Option<StaticCost>>,
+    _clarity_version: &ClarityVersion,
+) -> StaticCost {
+    match cost_map.get(function_name) {
+        Some(Some(cost)) => cost.clone(),
+        // Function exists but cost not yet computed (circular dependency?) or not found:
+        // return zero to avoid infinite recursion.
+        Some(None) | None => StaticCost::ZERO,
```
jbencin-stacks: This seems like another place where we are returning `StaticCost::ZERO` when returning an error or panicking would be better. Like if we detect a circular dependency, which shouldn't be possible in Clarity, we should panic. Also, should we handle `Some(None)` and `None` differently?

components/clarity-static-cost/Cargo.toml
brady-stacks: This is actually pretty annoying to remove as it stands. I can possibly remove it later, but so much stuff breaks without it. It would take a decent amount to untangle it.

components/clarity-static-cost/src/static_cost/special_costs.rs
brady-stacks: I'm changing all the usages of `as u64` in this module 👌

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+};
+pub use cost_functions::ClarityCostFunctionExt;
+pub use error::StaticCostError;
+pub use special_costs::get_cost_for_special_function;
+use stacks_common::types::StacksEpochId;
+pub use trait_counter::{
+    TraitCount, TraitCountCollector, TraitCountContext, TraitCountPropagator, TraitCountVisitor,
+};
+
+pub(super) fn saturating_add_cost(a: &mut ExecutionCost, b: &ExecutionCost) {
```
jbencin-stacks: Minor, but could be
```suggestion
pub(crate) fn saturating_add_cost(a: &mut ExecutionCost, b: &ExecutionCost) {
```

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+    a.read_count = a.read_count.saturating_add(b.read_count);
+}
+
+const STRING_COST_BASE: u64 = 36;
+const STRING_COST_MULTIPLIER: u64 = 3;
+
+pub(crate) fn calculate_function_cost(
+    function_name: &str,
+    cost_map: &HashMap<String, Option<StaticCost>>,
+    _clarity_version: &ClarityVersion,
```
jbencin-stacks: Could remove if not going to use

components/clarity-static-cost/tests/mod.rs
```diff
@@ -0,0 +1,1069 @@
...
+    }
+}
+
+#[test]
+fn test_trait_counts_simplified() {
+    // Simplified test case to debug trait counting
+    // This focuses on the key scenarios:
+    // 1. CONTEXT function with trait parameters and struct parameter with trait field
+    // 2. mint function that calls CONTEXT and uses trait parameters
+    let contract_src = r#"
```
jbencin-stacks: Have Claude convert the multi-line strings to use `indoc!()`

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1721 @@
...
+pub struct SummingExecutionCost {
+    pub costs: Vec<ExecutionCost>,
+}
+
+impl SummingExecutionCost {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn from_single(cost: ExecutionCost) -> Self {
```
jbencin-stacks: Implement `From<ExecutionCost>` instead?

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+        // Always add the succeed path (total_cost after all children have been processed).
+        // When there are no asserts!/branching children this is the only path.
+        // When there are asserts! children, total_cost represents the path where all
+        // assertions passed and all siblings were evaluated.
+        // When there are If/Match children, total_cost == their min/max, which is
+        // already represented in summing_cost — adding it again is harmless.
+        summing_cost.add_cost(total_cost_min);
+        summing_cost.add_cost(total_cost_max);
```
jbencin-stacks, line 318: Claude says this can double count in some cases. I'm not sure I understand though

> When there's a branching child, the code adds all branch paths to `summing_cost` (lines 294-299), but also adds the child cost to `total_cost_min`/`total_cost_max` (lines 304-305). Then at the end of the function (lines 327-328), `total_cost_min` and `total_cost_max` are added to `summing_cost` again. This means the branching child's cost is counted twice for the "all-passed" path. The comment on line 325-326 acknowledges this ("adding it again is harmless") but this only holds if `min()` and `max()` are taken over the result — the intermediate `summing_cost` has more paths than necessary, which may mask the issue.

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,897 @@
...
+        if let Some(min) = min_size {
+            total_min_size = total_min_size.saturating_add(min);
+        }
+        if let Some(max) = max_size {
+            total_max_size = total_max_size.saturating_add(max);
+        }
+    }
+
+    // Fallback to args.len() if we couldn't determine sizes
+    if total_min_size == 0 && total_max_size == 0 {
```
jbencin-stacks: What happens if only one arg is `0`? Let's say `total_min_size` is `5` and `total_max_size` is `0`, you could have `total_max_size` < total_min_size`, right?

components/clarity-static-cost/src/static_cost/cost_functions.rs
```diff
@@ -0,0 +1,167 @@
...
+        NativeFunctions::TransferToken => ClarityCostFunction::FtTransfer,
+        NativeFunctions::TransferAsset => ClarityCostFunction::NftTransfer,
+        NativeFunctions::MintAsset => ClarityCostFunction::NftMint,
+        NativeFunctions::MintToken => ClarityCostFunction::FtMint,
+        NativeFunctions::GetTokenSupply => ClarityCostFunction::FtSupply,
+        NativeFunctions::BurnToken => ClarityCostFunction::FtBurn,
+        NativeFunctions::BurnAsset => ClarityCostFunction::NftBurn,
+        NativeFunctions::GetStxBalance => ClarityCostFunction::StxBalance,
+        NativeFunctions::StxTransfer => ClarityCostFunction::StxTransfer,
+        NativeFunctions::StxBurn => ClarityCostFunction::Unimplemented,
```
jbencin-stacks: Leave this unimplemented? Do we give any sort of notice to users if functions in their contracts are unimplemented?

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,897 @@
...
+    StaticCost {
+        min: min_cost,
+        max: max_cost,
+    }
+}
+
+// MintAsset cost is based on asset_size
+pub fn cost_mint_asset(args: &[SymbolicExpression], epoch: StacksEpochId) -> ExecutionCost {
+    let size = args
+        .get(2)
```
jbencin-stacks: Should be:

```suggestion
        .get(1)
```

`nft-mint?` has the signature `(nft-mint? asset-class asset-identifier recipient)`

`args[0]` (`nft-mint?`) was stripped when we called `calculate_function_cost_from_native_function()`, so *here*, `args[0]` is the asset class name, `args[1]` is the asset identifier (whose size drives cost), and `args[2]` is the recipient principal, so `get(2)` is giving you the recipient principal instead of the asset identifier

jbencin-stacks (commented):
Some additional comments. It's coming along nicely, `StaticCostError` is a nice addition!

components/clarity-static-cost/src/static_cost/mod.rs
brady-stacks: 👍  split it up into 2 different CostCalculation errors, one for "function not found" and one for circular dep

components/clarity-static-cost/src/static_cost/special_costs.rs
brady-stacks: checks now to see if either resolved so we avoid the && case where max < min

components/clarity-static-cost/src/static_cost/special_costs.rs
brady-stacks: great catch. Added a dynamic-vs-static test to make sure this was working 👌

components/clarity-static-cost/src/static_cost/cost_functions.rs
brady-stacks: This one specifically is actually implemented, so I updated that. The allowance ones are however unimplemented. It now prints a warning if those are encountered. Specifically if no cost_fn is found for any native function

components/clarity-static-cost/src/static_cost/mod.rs
brady-stacks: Yea, this summing was over complicated. It's now simplified. I think the reason it wasn't showing up was because the dynamic-vs-static tests where still within range even if min and/or max was inflated. The problem would have arisen with larger functions

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -0,0 +1,322 @@
...
+
+    for child in &node.children {
+        let child_summing = calculate_total_cost_with_summing(child);
+        // Combine each existing path with each child path (cartesian product)
+        let current_paths = summing_cost.costs.clone();
+        summing_cost = SummingExecutionCost::new();
+        for current_path in current_paths {
+            for child_path in &child_summing.costs {
+                let mut combined_path = current_path.clone();
+                let _ = combined_path.add(child_path);
```
jbencin-stacks: There are still 12 instances of `let _ = cost.add(...)` where an overflow error is silently discarded. I don't think it's likely that we'll ever overflow `u64`s, but in case we do it's better to saturate or panic than silently fail

This one, and lines 222, 230, 242, 270, 273, 278, 279, 283, 284, 293, and 294

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,922 @@
...
+                .map(u64::from)
+        })
+        .unwrap_or(0);
+    ClarityCostFunction::FetchVar
+        .eval_for_epoch(size, epoch)
+        .unwrap_or(ExecutionCost::ZERO)
+}
+
+// SetVar cost is epoch-dependent and uses result size
+// TODO
```
jbencin-stacks: Do you plan to address this `TODO` and the one in the next function?

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -0,0 +1,1729 @@
...
+                        .eval_for_epoch(u64::from(sig.size().unwrap_or(0)), epoch)
+                        .unwrap_or(ExecutionCost::ZERO);
+                    let type_check_min_cost = ClarityCostFunction::InnerTypeCheckCost
+                        .eval_for_epoch(u64::from(sig.min_size().unwrap_or(0)), epoch)
+                        .unwrap_or(ExecutionCost::ZERO);
+                    saturating_add_cost(&mut overhead.min, &type_check_min_cost);
+                    saturating_add_cost(&mut overhead.max, &type_check_cost);
+                }
+            }
+            Err(_e) => {
```
jbencin-stacks: Do something here, at least print to stderr

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,922 @@
...
+        | NativeFunctions::CmpGreater
+        | NativeFunctions::CmpLess => cost_comparison(native_function, args, epoch, user_args),
+        NativeFunctions::Equals => cost_equals(args, epoch, user_args),
+        NativeFunctions::MintAsset => cost_mint_asset(args, epoch, user_args),
+        native_function => {
+            let cost_fn = from_native_function(native_function);
+            if matches!(cost_fn, ClarityCostFunction::Unimplemented) {
+                eprintln!(
+                    "warning: no cost function implemented for {:?}, using zero cost",
+                    native_function
```
jbencin-stacks:
```suggestion
                    "warning: no cost function implemented for {native_function:?}, using zero cost",
```

components/clarity-static-cost/src/static_cost/trait_counter.rs
```diff
@@ -0,0 +1,782 @@
...
+
+/// Second pass visitor: propagates trait counts through function calls
+pub struct TraitCountPropagator<'a> {
+    trait_counts: &'a mut TraitCount,
+    trait_names: &'a HashMap<ClarityName, String>,
+    /// Set of all function names from the contract (used to distinguish function calls from let-bound variables)
+    visited_functions: &'a HashSet<String>,
+    /// Map from let-bound variable names to their trait counts
+    #[allow(dead_code)]
+    let_binding_trait_counts: HashMap<String, MinMaxTraitCount>,
```
jbencin-stacks: This is unused, can it be removed?

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,922 @@
...
+                    // Use serialized sizes: key stored as-is, value stored as Optional(value)
+                    let key_min = u64::from(key_type.min_size().unwrap_or(0));
+                    let key_max = u64::from(key_type.max_serialized_size().unwrap_or(0));
+                    let value_min = u64::from(value_type.min_size().unwrap_or(0));
+                    let value_max = TypeSignature::new_option(value_type.clone())
+                        .ok()
+                        .and_then(|t| t.max_serialized_size().ok())
+                        .map(u64::from)
+                        .unwrap_or(0);
+                    min_size = key_min + value_min;
```
jbencin-stacks: Should use saturating adds in this function

components/clarity-static-cost/src/static_cost/special_costs.rs
brady-stacks: Yes 👌 added

components/clarity-static-cost/src/static_cost/mod.rs
brady-stacks: Fixed ✅

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,1030 @@
...
+    _env: Option<&clarity::vm::contexts::Environment>,
+    user_args: Option<&UserArgumentsContext>,
+) -> ExecutionCost {
+    let size = args
+        .first()
+        .and_then(|arg| arg.match_atom())
+        .and_then(|var_name| {
+            user_args?
+                .get_data_var_type(var_name)
+                .and_then(|type_sig| type_sig.max_serialized_size().ok())
```
jbencin-stacks: We have `max_serialized_size()` so we can do like in `cost_set_var()` and use it to compute a max cost return `StaticCost` instead of `ExecutionCost`

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,1030 @@
...
+                    if let Some(value) =
+                        arg.match_atom_value().or_else(|| arg.match_literal_value())
+                    {
+                        if let Ok(size) = value.serialized_size() {
+                            let s = u64::from(size);
+                            return (Some(s), Some(s));
+                        }
+                    }
+                    // For variables, fall back to get_argument_sizes
+                    get_argument_sizes(arg, epoch, user_args)
```
jbencin-stacks: `get_argument_sizes()` uses `TypeSignature::size()` / `TypeSignature::min_size()`, but for literals we use `serializad_size()`. Not sure which one is better but should be consistent

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,1030 @@
...
+    ClarityCostFunction::ReplaceAt
+        .eval_for_epoch(size, epoch)
+        .unwrap_or(ExecutionCost::ZERO)
+}
+
+// FetchVar cost uses the serialized size of the stored variable's value type
+pub fn cost_fetch_var(
+    args: &[SymbolicExpression],
+    epoch: StacksEpochId,
+    _env: Option<&clarity::vm::contexts::Environment>,
```
jbencin-stacks: Can remove unused var

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,1030 @@
...
+    // SetEntry args: [map-name, key, value]
+    // For epoch 3.3+, cost is based on serialized entry size
+    // We calculate min/max from TypeSignature sizes
+
+    let mut min_size = 0u64;
+    let mut max_size = 0u64;
+
+    if args.len() >= 3 {
+        // Try to get map types from contract context (doesn't require mutable access)
+        if let Some(environment) = env {
```
jbencin-stacks: You have pretty much the same code in `cost_fetch_entry()`. You could factor it out into:

```rust
fn resolve_map_types(map_name, env, user_args) -> Option<(&TypeSignature, &TypeSignature)>
{
    // ...
}
```

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -0,0 +1,1030 @@
...
+                min_size = u64::from(key_type.min_size().unwrap_or(0));
+                // max: key found case, key + Optional(value) sizes charged
+                // Values are stored as Optional(value_type) in the DB
+                let key_max = u64::from(key_type.max_serialized_size().unwrap_or(0));
+                let value_max = TypeSignature::new_option(value_type.clone())
+                    .ok()
+                    .and_then(|t| t.max_serialized_size().ok())
+                    .map(u64::from)
+                    .unwrap_or(0);
+                max_size = key_max + value_max;
```
jbencin-stacks: This line, and line 736 still use `+` instead of saturating add

brady-stacks: It would be nice if `+` meant `saturating_add` and there was a separate `unsafe_add` that did what the current addition does...

components/clarity-static-cost/src/static_cost/special_costs.rs
brady-stacks: I looked into this a bit further and narrowed it down to this: Basically any cost that is computed with `cost_input_sized_varg` is currently using the serialized size. This is Equals, TupleMerge, IndexOf, ToConsensusBuff, and all the has functions. You can grep the function name in stacks-core to see all the uses.

The fix was to pass an argument to `get_argument_sizes` that specifies if we should use serialized or normal.

components/clarity-static-cost/src/static_cost/special_costs.rs
```diff
@@ -213,27 +213,42 @@ fn infer_tuple_size_from_expression(
...
     arg: &SymbolicExpression,
     epoch: StacksEpochId,
     user_args: Option<&UserArgumentsContext>,
+    serialized: bool,
 ) -> (Option<u64>, Option<u64>) {
     // Try literal value first
     if let Some(value) = arg.match_atom_value().or_else(|| arg.match_literal_value()) {
-        if let Ok(size) = value.size() {
+        let size = if serialized {
+            value.serialized_size().map_err(|_| ())
```
jbencin-stacks: Discard error? We should at least print to stderr, we can do something better in a followup PR

jbencin-stacks (dismissed):

PR #2216, opened 2026-02-18
fix: api should not load snapshot db if `--from-genesis`

brady-stacks:
devnet api container wasn't respecting the `--from-genesis` flag

### DIFF
--- a/components/stacks-network/src/orchestrator.rs
+++ b/components/stacks-network/src/orchestrator.rs
@@ -460,7 +460,7 @@ impl DevnetOrchestrator {
                 &format!("http://localhost:{stacks_api_port}/doc"),
             );
 
-            match self.boot_stacks_api_container(ctx).await {
+            match self.boot_stacks_api_container(ctx, no_snapshot).await {
                 Ok(_) => {}
                 Err(message) => {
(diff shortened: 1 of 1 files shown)


components/stacks-network/src/orchestrator.rs
```diff
@@ -1565,49 +1569,53 @@ db_path = "stacks-signer-{signer_id}.sqlite"
...
+                        String::from_utf8_lossy(&output.stderr)
+                    ));
+                }
 
-            // Run the import command
-            let import_command = format!(
+                // Run the import command
+                let import_command = format!(
                 "docker exec {container_name} node /app/lib/index.js import-events --file /tmp/events_cache.tsv --wipe-db"
             );
```
brice-stacks: indentation?

brice-stacks (approved):
Looks good, just noticed one indentation weirdness.

components/stacks-network/src/orchestrator.rs
brady-stacks: weird. fmt missed that? 🤔

brice-stacks (approved):

PR #2217, opened 2026-02-18
chore: new clarinet-defaults

brady-stacks:
follow up from [this comment](https://github.com/stx-labs/clarinet/pull/2207/changes#r2823463733)

Adds a shareable module for defaults to allow using DEFAULT_* without importing all of clarinet_repl


hugo-stacks:
Sorry for the delays on this one @brady-stacks 
I've been trying to find it would make sense to add more stuff here. I think their will be, but let's start with that.
It just needs a small conflict solving

brady-stacks:
@hugo-stacks updated 👍

PR #2224, opened 2026-02-20
perf(clarinet-format): Reduce unnecessary `String` allocations

jbencin-stacks:
### Description

I have noticed previously that there were a lot of unnecessary string allocations with `format!()` in `clarinet-formatter`

Wasn't worth the effort for a human to clean it up, but now that I've got the Claude subscription I wanted to see what it could do

#### Change Made

  1. Replaced ~20 `push_str(&format!(...))` with `push`/`push_str` sequences - Avoids allocating a temporary String for each `format!()` call. Instead, characters and string slices are pushed directly to the accumulator.
  2. Cached `indentation.to_string()` once in Aggregator - Added `indentation_str: String` field to Aggregator, computed once in `new()`. Previously, every formatting function called `self.settings.indentation.to_string()` independently (15 separate allocations per format call).
  3. Hoisted `double_indent` computation out of loops - In `format_let`, `format_restrict_assets`, and `format_key_value_sugar`, the double_indent string was being recomputed on every loop iteration. Now computed once before the loop.
  4. Eliminated unnecessary `.to_string()` clones - Removed `formatted.to_string()` where formatted was already a `String`, `value.to_string()` on already-owned strings, and `"".to_owned()` replaced with `String::new()`.
  5. Optimized `NativeFunctions`/`DefineFunctions` match arms - Replaced format! concatenation with direct `push`/`push_str` on owned strings. Eliminated `"\n".to_string()`, `" ".to_string()` allocations.
  6. Optimized `to_inner_content` closing - Replaced two-step format! with pre-allocated `String::with_capacity` and direct pushes.
  7. Optimized `comment_piece` - Avoided allocating a `String` from `""` or `" "` in the common case where `space_count == 0`.

#### Breaking change?

No

### Example

This PR results in about a **15%** performance improvement

#### Benchmark Results

```
  ┌────────────────────────┬─────────────────┬────────────────┬───────────────┐
  │       Benchmark       │ Before (median) │ After (median)│  Improvement  │
  ├────────────────────────┼─────────────────┼────────────────┼───────────────┤
  │ format_all_contracts  │ 39.74 ms        │ 34.06 ms      │ ~14.3% faster │
  ├────────────────────────┼─────────────────┼────────────────┼───────────────┤
  │ format_clarity_bitcoin│ 6.806 ms        │ 5.777 ms      │ ~15.1% faster │
  └────────────────────────┴─────────────────┴────────────────┴───────────────┘
```

### DIFF
--- a/components/clarinet-format/src/formatter/mod.rs
+++ b/components/clarinet-format/src/formatter/mod.rs
@@ -666,8 +666,7 @@ impl<'a> Aggregator<'a> {
                             if is_comment(next_item) {
                                 let count =
                                     next_item.span().start_column - arg.span().end_column - 1;
-                                let spaces = " ".repeat(count as usize);
-                                acc.push_str(&spaces);
+                                push_spaces(&mut acc, count);
                                 acc.push_str(&self.display_pse(next_item, previous_indentation));
                                 items_iter.next();
                             }
@@ -683,8 +682,7 @@ impl<'a> Aggregator<'a> {
                 if let Some(comment) = trailing {
                     if let Some(last_item) = method_list.last() {
                         let count = comment.span().start_column - last_item.span().end_column - 1;
-                        let spaces = " ".repeat(count as usize);
-                        acc.push_str(&spaces);
+                        push_spaces(&mut acc, count);
                         acc.push_str(&self.display_pse(comment, previous_indentation));
                     }
                 }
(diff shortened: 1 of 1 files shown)


components/clarinet-format/src/formatter/mod.rs
```diff
@@ -1813,8 +1828,15 @@ impl<'a> Aggregator<'a> {
             let trimmed = last_line.trim();
             !(trimmed == ")" || trimmed == "}")
         };
-        let newlined = format!("\n{previous_indentation})");
-        format!("({}{}", result, if break_lines { &newlined } else { ")" })
+        let mut out = String::with_capacity(result.len() + previous_indentation.len() + 4);
```
brady-stacks, line 1831: +4 ? `(` `\n` and `)` should be 3?

jbencin-stacks: I'm guessing it's leaving room for null termination, for compatibility with C-strings over FFI

PR #2225, opened 2026-02-20
refactor: devnet orchestrator improvements

hugo-stacks:
### Description

fixes: #2228

- relying more on tokio
  - for sleep()s without blocking the thread
  - booting containers in parallel
- code hygiene (dryness)
- indoc for configs
- simplify bitcoin rpc client, with call_retry

### DIFF
--- a/components/stacks-network/src/orchestrator.rs
+++ b/components/stacks-network/src/orchestrator.rs
@@ -1906,36 +1906,14 @@ impl DevnetOrchestrator {
             },
         )?;
 
-        // Start all containers in parallel
-        // let _ = tokio::join!(
-        //     docker.start_container::<String>(&bitcoin_node_c_id, None),
-        //     docker.start_container::<String>(bitcoin_explorer_c_id, None),
-        //     docker.start_container::<String>(postgres_c_id, None),
-        //     docker.start_container::<String>(stacks_api_c_id, None),
-        //     docker.start_container::<String>(stacks_explorer_c_id, None),
-        //     docker.start_container::<String>(&stacks_node_c_id, None),
-        // );
-        let _ = docker
-            .start_container::<String>(&bitcoin_node_c_id, None)
-            .await;
-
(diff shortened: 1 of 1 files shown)


components/stacks-network/src/orchestrator.rs
```diff
@@ -195,8 +234,8 @@ impl DevnetOrchestrator {
             stacks_node_host: format!("stacks-blockchain.{namespace}.svc.cluster.local:20443"),
             postgres_host: format!("stacks-blockchain-api.{namespace}.svc.cluster.local:5432"),
             stacks_api_host: format!("stacks-blockchain-api.{namespace}.svc.cluster.local:3999"),
-            stacks_explorer_host: "localhost".into(), // todo (micaiah)
-            bitcoin_explorer_host: "localhost".into(), // todo (micaiah)
```
hugo-stacks, line 199: The hosted devnet (on the platform) and the whole k8s related stuff, will likely be removed soon, and is already more or less unmaintained. Let's forget about it

components/stacks-network/src/orchestrator.rs
```diff
@@ -682,62 +633,44 @@ impl DevnetOrchestrator {
...
+                port_binding(devnet_config.bitcoin_node_rpc_port),
+            ),
+            (
+                format!("{}/tcp", devnet_config.bitcoin_node_p2p_port),
+                port_binding(devnet_config.bitcoin_node_p2p_port),
+            ),
+            // ZMQ block notifications
+            ("28332/tcp".to_string(), port_binding(28332)),
+            // ZMQ transaction notifications
+            ("28333/tcp".to_string(), port_binding(28333)),
```
brady-stacks, line 648: will we be able to delete these once ZMQ is gone? Or are they used by the stacks observer too?

hugo-stacks: I think we will yes 👍

Issue #2214, opened 2026-02-18
`unused` lint warning displayed on the inner expression


hugo-stacks:
It can be confusing that the `unused` warning is on the inner expression, instead of being on the variable name (or the whole binding expression)

<img width="664" height="183" alt="Image" src="https://github.com/user-attachments/assets/1a1ea99d-0e9f-49a8-a207-acd05e955f79" />

jbencin-stacks:
This is due to how `ASTVisitor` works. I process the bindings in this function:

```rust
fn traverse_let(
    &mut self,
    expr: &'a SymbolicExpression,
    bindings: &HashMap<&'a ClarityName, &'a SymbolicExpression>,
    body: &'a [SymbolicExpression],
) -> bool {
    // ...
}
```

Inside of `travers_let()`, I have the `ClarityName` of the bindings and the `SymbolicExpression` of the value. I don't have any `SymbolicExpression` or `Span` that includes the name of the binding. I may have to change `ASTVisitor` to pass this info

PR #2230, opened 2026-02-20
fix: Highlight binding name instead of expression in `unused_binding` lint

jbencin-stacks:
### Description

Fixes: #2214

Highlight binding name instead of expression in `unused_binding` lint. This was caused by the desired information not being available in `ASTVisitor::traverse_let()`, so that has been fixed to include a `Span` for the binding name also

#### Breaking change?

No

### Example

Sample output from `clarinet check`:

```
warning: `let` binding `check1` is never used
--> /home/jbencin/git/stx-labs/stacks-wormhole-ntt/contracts/ntt-manager-v1.clar:238:10
  (let ((check1 (try! (check-enabled)))
         ^~~~~~
Remove this expression or suffix binding with '_' if this is intentional
```


brady-stacks (approved):
lgtm

Issue #2215, opened 2026-02-18
Requirements analysis inconsistencies


hugo-stacks:
# Context

The contract analysis can be run via both the CLI and the LSP.
But they don't return the same output for requirements:

- `clarinet check` does return the diagnostics (especially lints) for requirements
- the lsp doesn't return  those diagnostic

After a quick investigation, I think it's because we only collect the requirements for the contract ids that are specified in the project contract in the manifest (not in the requirements) [here](https://github.com/jbencin-stacks/clarinet/blob/8f7322899fa456df3707a0c99703051c701de413/components/clarity-lsp/src/common/state.rs#L649)

# Expected behavior

Ideas:
- The LSP and CLI should return consistent diagnostics
- We have a choice between collecting those diagnostics or not
  - I think we should **not** get the requirements diagnostics by default, it's noise for developers (including for AI agents). Other tools don't show diagnostics for project dependencies (think cargo crates, npm packages)
  - We could have an options to also run checks/lints on the requirements, but I think that's **an other feature**
  - We should still show clarity-vm error if a requirement failed to be parsed or deployed, that would be a critical bug

tl;dr - i don't think we should run the linter on requirements (by default)

PR #2231, opened 2026-02-21
fix: Don't run linter on external requirements

jbencin-stacks:
### Description

Fixes: #2215

`clarinet check` should not run linter/analysis passes on project requirements. It **should** continue to run the interpreter and report any errors from that

#### Breaking change?

No

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -296,16 +296,16 @@ pub async fn generate_default_deployment(
     file_accessor: Option<&dyn FileAccessor>,
 ) -> Result<(DeploymentSpecification, DeploymentGenerationArtifacts), String> {
     let network_manifest = match file_accessor {
-        None => NetworkManifest::from_project_manifest_location(
-            &manifest.location,
+        None => NetworkManifest::from_project_root(
+            &manifest.root_dir,
             &network.get_networks(),
             manifest.use_mainnet_wallets(),
             Some(&manifest.project.cache_location),
             None,
         )?,
         Some(file_accessor) => {
-            NetworkManifest::from_project_manifest_location_using_file_accessor(
-                &manifest.location,
+            NetworkManifest::from_project_root_using_file_accessor(
+                &manifest.root_dir,
                 &network.get_networks(),
                 manifest.use_mainnet_wallets(),
                 file_accessor,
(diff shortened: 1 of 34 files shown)


hugo-stacks:
I'll test it now

components/clarinet-deployments/src/lib.rs
```diff
@@ -198,7 +198,14 @@ pub fn update_session_with_deployment_plan(
                         should_mint_sbtc = true;
                     }
                     let contract_ast = contracts_asts.as_ref().and_then(|m| m.get(&contract_id));
-                    let result = handle_emulated_contract_publish(session, tx, contract_ast, epoch);
+                    let is_project_contract = deployment.contracts.contains_key(&contract_id);
```
hugo-stacks: Requirements _are_ in the deployment, as regular contracts.

I guess there are multiple options, I guess the most straightforward one would be to pass the manifest to `update_session_with_deployment_plan()`

components/clarinet-deployments/src/lib.rs
```diff
@@ -244,7 +252,7 @@ fn handle_emulated_contract_publish(
         epoch: clarity_repl::repl::Epoch::Specific(epoch),
     };
 
-    let result = session.deploy_contract(&contract, false, contract_ast);
+    let result = session.deploy_contract(&contract, false, contract_ast, run_repl_analysis);
```
hugo-stacks: Idea: we could also consider adding a `is_requirement` field to `ClarityContract`.
Not sure if it would actually be better. Just something to consider

hugo-stacks (changes requested):

components/clarinet-deployments/src/lib.rs
jbencin-stacks: Could work, but I think if we can solve this without changing the file structure of the TOML or YAML files that would be ideal

hugo-stacks: @jbencin-stacks 
Oooh right, I didn't realize this was the structure used in the manifest

hugo-stacks: We could still skip  serialization / deserialization though

hugo-stacks: (edited - _skip_ serialization)

PR #2236, opened 2026-02-24
chore: remove chainhooks v1

### LINKED ISSUE #2227: Remove `chainhook` v1
## Context

"Chainhook v1" was originally a part of Clarinet that was used to monitor the Devnet Bitcoin and Stacks blockchain activities and orchestrate the Devnet accordingly.
**It is _still_ being used for that purpose** (around [here](https://github.com/stx-labs/clarinet/blob/b82533a947c03379b4a1e82a9df621f5df8a89b0/components/stacks-network/src/chains_coordinator.rs#L375) and in other places)

It's also used for Devnet users to register custom chainhooks predicates (implemented [here](https://github.com/stx-labs/clarinet/blob/b82533a947c03379b4a1e82a9df621f5df8a89b0/components/stacks-network/src/chainhooks.rs#L23-L51)), and test those chainhook predicates locally.
See documentation: https://docs.stacks.co/clarinet/integrations/chainhook.

Chainhook v1 will be shutdown completely in March.
See blog post: https://www.hiro.so/blog/chainhooks-v2-is-now-generally-available
We can stop supporting it in Clarinet. But we'll still need it, or something similar, to orchestrate the Devnet.

Chainhook v2 is quite different. It's implemented in Node.js. With no clear path on how to use it locally, but that would be different to the v1 implementation.

## Technical details

_Here, chainhook means chainhook v1_

The chainhook code was initially in Clarinet, then pulled out in its own repository and changed to handle way more uses cases (more L1 predicate, ordinals and things like that). It had a strong dependency on stacks-core and Clarinet, creating very annoying circular dependencies between the clarinet <-> chainhook repositories.

The code was eventually put back into the clarinet repository, making it easier to manage (the Hiro repository was more or less unmaintained in favor of the new version).

The thing is that it contains tons of things that are not needed in Clarinet, especially as "custom predicates" support is being removed.
There is an opportunity to remove tons of dead code and simplify the code base here. The chainhook unit tests, because they support so many use cases, also take a lot of time to run.

## Tasks

- Remove support for custom chainhook in Devnet
- Simplify "chainhook" into something way simpler that only handle the needs of Devnet orchestration

brady-stacks:
- closes #2227 
- [x] remove old chainhooks (all predicate related code paths) 
- [x] ensure working with devnet

What was preserved from previous impl
  - Core observer event loop                                                                                                                             
  - `PoxConfig`, `StacksOpcodes`, `get_stacks_canonical_magic_bytes`, `get_canonical_pox_config` (used by indexer)
  - `try_decode_clarity_value` (used by indexer)
  - All indexer code (unchanged)

**Note: Please make sure to test `devnet start` and `devnet start --from-genesis`**


hugo-stacks:
@brady-stacks 
- It might make sense to rename chainhook into something more generic (like "observer") to avoid future confusion (especially as we might re-integrate totally chainhook v2 concepts in the future)
- Would it make sense, architecturally, to leave it into it's own crate as a project components instead of putting it into `stacks-network/src/`?

brady-stacks:
@hugo-stacks we can merge this now or later, but the observer renaming is [here](https://github.com/brady-stacks/clarinet/pull/2)

hugo-stacks:
@brady-stacks 

I think it would be simpler to have just one PR for both.

For instance, this PR adds a lot of dependencies to `components/stacks-network/Cargo.toml`, while the `observer` PR puts those dependencies in `components/observer/Cargo.toml` but doesn't remove them from stacks-core

components/stacks-network/Cargo.toml
```diff
@@ -6,12 +6,19 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+axum = { version = "0.8.6", features = ["json"] }
+tower = "0.5.2"
```
hugo-stacks: I'm assuming _most_ of the changes in this Cargo.toml are not needed @brady-stacks right? 
Since they belong to `observer/Cargo.toml`

brady-stacks: Removed 👍

components/observer/src/event_handler/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+            config.get_stacks_node_config().ingestion_port
+        )
+    });
+
+    ctx.try_log(|logger| slog::info!(logger, "Observing Bitcoin chain events via Stacks node"));
+
+    Ok(())
+}
+
+pub async fn start_stacks_event_observer(
```
hugo-stacks: Doesn't need to be `pub`

components/observer/src/event_handler/mod.rs
```diff
@@ -0,0 +1,496 @@
...
+    observer_events_tx: Option<crossbeam_channel::Sender<ObserverEvent>>,
+    stacks_startup_context: StacksObserverStartupContext,
+    ctx: Context,
+) -> Result<(), Box<dyn Error>> {
+    let indexer_config = IndexerConfig {
+        bitcoind_rpc_url: config.bitcoind_rpc_url.clone(),
+        bitcoind_rpc_username: config.bitcoind_rpc_username.clone(),
+        bitcoind_rpc_password: config.bitcoind_rpc_password.clone(),
+        stacks_network: StacksNetwork::Devnet,
+        bitcoin_network: BitcoinNetwork::Regtest,
```
hugo-stacks: From a logical point of view, it should be passed-in from the `config: EventObserverConfig`.

In this context where the chainhook / observer is only use in Devnet, this will also be hardcoded to Devnet and Regtest.

So either:
- simply take it from the config (setup [here](https://github.com/brady-stacks/clarinet/blob/74828fe01c1dfb14f59cd423489888b2a5af9b28/components/stacks-network/src/chains_coordinator.rs#L147-L148))
- or fully remove the configs

brady-stacks: took it from the config 👍

PR #2244, opened 2026-02-27
ci: Fix flaky CI tests by using mock HTTP server instead of API

jbencin-stacks:
### Problem

I am occasionally seeing errors like this in GitHub CI: https://github.com/stx-labs/clarinet/actions/runs/22461148736/job/65056169375?pr=2231

image

Rerunning the failed jobs usually fixes it, but it's still a waste of time and resources to run the CI repeatedly. The issue is that fetching the required sBTC contracts from the API is failing sometimes.

### Changes

- Commits the required contracts as fixtures under `components/clarinet-sdk/node/tests/fixtures/api-responses`
- In the Rust code, introduces unit tests using mock HTTP server `mockito` to test network component of `fetch_contract()` 
- In the Typscript Code, start server using `node.http` to serve contracts from fixtures

#### Breaking change?

No

### DIFF
--- a/components/clarinet-deployments/src/requirements.rs
+++ b/components/clarinet-deployments/src/requirements.rs
@@ -28,7 +28,7 @@ pub async fn retrieve_contract(
     contract_id: &QualifiedContractIdentifier,
     cache_location: &Path,
     file_accessor: &Option<&dyn FileAccessor>,
-    api_url: Option<&str>,
+    api_base_url: Option<&str>,
 ) -> Result<(String, StacksEpochId, ClarityVersion, PathBuf), String> {
     let contract_deployer = contract_id.issuer.to_address();
     let contract_name = contract_id.name.to_string();
@@ -70,8 +70,8 @@ pub async fn retrieve_contract(
         .unwrap()
         .is_mainnet();
 
-    let base_url = api_url.unwrap_or_else(|| default_api_url(is_mainnet));
-    let contract = fetch_contract(base_url, &contract_deployer, &contract_name).await?;
+    let api_base_url = api_base_url.unwrap_or_else(|| default_api_base_url(is_mainnet));
+    let contract = fetch_contract(api_base_url, &contract_deployer, &contract_name).await?;
 
     let epoch = epoch_for_height(is_mainnet, contract.block_height);
     let clarity_version = match contract.clarity_version {
@@ -131,7 +131,7 @@ struct Contract {
     clarity_version: Option<u8>,
 }
 
-fn default_api_url(is_mainnet: bool) -> &'static str {
+fn default_api_base_url(is_mainnet: bool) -> &'static str {
     if is_mainnet {
         "https://api.hiro.so"
     } else {
@@ -139,8 +139,12 @@ fn default_api_url(is_mainnet: bool) -> &'static str {
     }
 }
 
-async fn fetch_contract(base_url: &str, deployer: &str, name: &str) -> Result<Contract, String> {
-    let url = format!("{base_url}/extended/v1/contract/{deployer}.{name}");
+async fn fetch_contract(
+    api_base_url: &str,
+    deployer: &str,
+    name: &str,
+) -> Result<Contract, String> {
+    let url = format!("{api_base_url}/extended/v1/contract/{deployer}.{name}");
     let response = reqwest::get(&url)
         .await
         .map_err(|e| format!("Unable to retrieve contract {url}: {e}"))?;
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -278,7 +278,7 @@ pub async fn generate_default_deployment(
     network: &StacksNetwork,
     no_batch: bool,
     file_accessor: Option<&dyn FileAccessor>,
-    api_url: Option<&str>,
+    api_base_url: Option<&str>,
 ) -> Result<(DeploymentSpecification, DeploymentGenerationArtifacts), String> {
     let network_manifest = match file_accessor {
         None => NetworkManifest::from_project_root(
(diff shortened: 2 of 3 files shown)


components/clarinet-sdk/node/tests/simnet-with-sbtc.test.ts
```diff
@@ -34,6 +34,10 @@ afterEach(() => {
   deleteExistingDeploymentPlan();
 });
 
+afterAll(() => {
+  simnet.free();
```
hugo-stacks: @jbencin-stacks What's the specific purpose of calling it?
I think we don't need to manage it ourselves in the context of unit tests where the garbage collector will take care of it (or the process will just end). And should do a good enough job so that we don't have to call it

jbencin-stacks: Claude wanted to put that in. I couldn't quite understand the reasoning it gave, but it's not going to hurt anything. Now it's telling me that it wasn't really necessary. I can take it out

hugo-stacks: Yeah LLMs tends to do that, especially for lesser known tools (eg: it's good at React but less good at wasm-bindgen...) 
We always have to by skeptical about what it does and don't let this kind of thing endup on main

jbencin-stacks:
@hugo-stacks I think we should have some test of network downloads in the test suite, but it would be better to spin up a local HTTP server which just handles the endpoints we need to fetch the contracts. Do you agree?

components/clarinet-deployments/src/requirements.rs
```diff
@@ -28,6 +28,7 @@ pub async fn retrieve_contract(
     contract_id: &QualifiedContractIdentifier,
     cache_location: &Path,
     file_accessor: &Option<&dyn FileAccessor>,
+    api_url: Option<&str>,
```
xoloki-stacks: Maybe call this `api_base_url`?  It's a little confusing how this is sometimes called `api_url` or `base_url`.

components/clarinet-deployments/src/requirements.rs
```diff
@@ -129,13 +131,15 @@ struct Contract {
     clarity_version: Option<u8>,
 }
 
-async fn fetch_contract(is_mainnet: bool, deployer: &str, name: &str) -> Result<Contract, String> {
-    let base_url = if is_mainnet {
+fn default_api_url(is_mainnet: bool) -> &'static str {
```
xoloki-stacks: Same here, consider `default_api_base_url`.

components/clarinet-deployments/src/requirements.rs
```diff
@@ -129,13 +131,15 @@ struct Contract {
...
+    if is_mainnet {
         "https://api.hiro.so"
     } else {
         "https://api.testnet.hiro.so"
-    };
+    }
+}
 
+async fn fetch_contract(base_url: &str, deployer: &str, name: &str) -> Result<Contract, String> {
```
xoloki-stacks: Same

components/clarinet-deployments/src/requirements.rs
```diff
@@ -151,3 +155,126 @@ async fn fetch_contract(is_mainnet: bool, deployer: &str, name: &str) -> Result<
...
+    #[tokio::test]
+    async fn test_fetch_contract_returns_error_on_404() {
+        let mut server = Server::new_async().await;
+
+        let mock = server
+            .mock(
+                "GET",
+                format!("/extended/v1/contract/{TEST_DEPLOYER}.{TEST_CONTRACT_NAME}").as_str(),
+            )
+            .with_status(404)
```
xoloki-stacks, line 214: Here you are explicitly marking a path to return `404`.  Wouldn't it be better to have this path return `200`, and then ask the server for a different path?  Isn't that a more natural test than this?

jbencin-stacks: I'm not sure I follow. There are two other tests here which return `200` and test cases in which the path was found. The purpose of this one is to test the `404` case, where the path isn't found

components/clarinet-deployments/src/requirements.rs
jbencin-stacks: Yeah that makes sense

components/clarinet-deployments/src/requirements.rs
xoloki-stacks: Sorry, I'll be more explicit.  Here, you're setting the mock server up to explicitly return a 404 for a particular path. But that's not a normal configuration for a web server; normally you setup a web server to return 200 for known paths, and it by default returns 404 for paths which are not setup.

So instead, you could configure the server to return 200 for a particular path, but in the test request a different path.  That way we're testing that the mock HTTP server behaves as expected; it *is* configured to return 200 for a correct path, but you ask it for a path that it is not configured for.

The result in the same for this test; you get a 404 from the mock server, and test that `fetch_contract` returns it correctly.  But you also validate that the mock server behaves as expected, i.e. it returns 404 for a path for which it is not configured.

Does that make sense?

xoloki-stacks: Feel free to tell me I'm overthinking this, and keep the test as is.

jbencin-stacks: It's not a typical server configuration, but we are trying to test a 404 error code. I could also just remove the path and it should get a 404 anyway. Probably makes more sense that way, and always better to have less code

jbencin-stacks: I tried removing the path and mockito returns a 501 instead of 404 if not found, so it has to stay like this

xoloki-stacks: Sounds good, thanks for looking into this.

PR #2251, opened 2026-03-02
feat(linter): Add lint `case_binding`

jbencin-stacks:
### Description

Add `case_binding` to linter (#2028)

This lint enforces `kebab-case` on bindings (`let` bindings and function arg names)

#### Definition of `kebab-case` used:
- We allow variables to start or end with `-` (Clarity allows this at the end, but not the start)
- We do not allow multiple consecutive `-`s, although Clarity does allow this

#### Breaking change?

No

### Example

Will generate a diagnostic message on code like this:

```clarity
(let ((camelCaseVar (u0))))
```

You can ignore both checks on a specific constant by placing the comment:

```clarity
;; #[allow(case_binding)]
```

### Checklist

### DIFF
--- a/components/clarity-repl/src/analysis/util/mod.rs
+++ b/components/clarity-repl/src/analysis/util/mod.rs
@@ -1,14 +1,16 @@
 //! Utility functions used by analysis passes/lints
 
-#[derive(Debug, PartialEq)]
+use thiserror::Error;
+
+#[derive(Debug, Error, PartialEq)]
 pub enum CaseError {
-    /// Empty string
+    #[error("identifier is empty")]
     Empty,
-    /// Identifier contains character not allowed in this case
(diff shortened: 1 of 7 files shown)


components/clarity-repl/src/analysis/util/mod.rs
```diff
@@ -149,6 +201,88 @@ mod tests {
...
+            "-l",
+            // Trailing hyphen
+            "trailing-hyphen-",
+            "54321-",
+            "t-",
+            // Leading and trailing hyphens
+            "-leading-and-trailing-hyphens-",
+            "-12321-",
+            "-lt-",
+            "-",
```
brady-stacks, line 234: This is valid?

jbencin-stacks: It's valid kebab-case but not valid Clarity

PR #2252, opened 2026-03-04
Run diagnostics with and without env(simnet) code during CLI clarinet check and LSP

xoloki-stacks:
### Description

This PR closes https://github.com/stx-labs/clarinet/issues/2022 by running diagnostics twice, with and without the env(simnet) code. This will show if the code can be deployed outside of simnet.

This is an alternative to #2189 which is significantly smaller, at the cost of not trying to unify the code paths between CLI and LSP.  However, that PR failed to do so, and resorted to manually inlining and touching many more places.

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -37,7 +37,7 @@ use clarity_repl::repl::settings::{ApiUrl, RemoteDataSettings};
 use clarity_repl::repl::{
     clarity_version_to_u8, ClarityCodeSource, ClarityContract, ContractDeployer, Epoch,
 };
-use clarity_repl::utils::REMOVE_ENV_SIMNET_PASSES;
+use clarity_repl::utils::CHECK_ENVIRONMENTS;
 use clarity_repl::{analysis, repl};
 use serde::Serialize;
 use stacks_network::{self, DevnetOrchestrator};
@@ -750,14 +750,18 @@ pub fn main() {
                     StacksNetwork::Simnet
                 };
 
-                let (mut deployment, _, _) =
-                    match generate_default_deployment(&manifest, &network, cmd.no_batch, false) {
-                        Ok(result) => result,
-                        Err(message) => {
-                            eprintln!("{}", format_err!(message));
-                            std::process::exit(1);
-                        }
-                    };
+                let (mut deployment, _, _) = match generate_default_deployment(
+                    &manifest,
+                    &network,
+                    cmd.no_batch,
+                    Environment::Simnet,
+                ) {
+                    Ok(result) => result,
+                    Err(message) => {
+                        eprintln!("{}", format_err!(message));
+                        std::process::exit(1);
+                    }
+                };
 
                 if !cmd.manual_cost
                     && matches!(network, StacksNetwork::Testnet | StacksNetwork::Mainnet)
@@ -861,7 +865,7 @@ pub fn main() {
                             None => {
                                 let default_deployment_path = project_root.join(get_default_deployment_path(network));
                                 let (deployment, _, _) =
-                                    generate_default_deployment(&manifest, network, false, false)
+                                    generate_default_deployment(&manifest, network, false, Environment::Simnet)
                                         .unwrap_or_else(|message| {
                                             eprintln!("{}", red!(message));
                                             std::process::exit(1);
@@ -1093,7 +1097,7 @@ pub fn main() {
                             &cmd.deployment_plan_path,
                             cmd.use_on_disk_deployment_plan,
                             cmd.use_computed_deployment_plan,
-                            false,
+                            Environment::Simnet,
                         );
 
                         if !artifacts.success {
@@ -1259,14 +1263,14 @@ pub fn main() {
             let manifest = load_manifest_or_exit(cmd.manifest_path, false);
             let mut exit_codes = Vec::new();
             let mut global_found_env_simnet = false;
-            for force_remove_env_simnet in REMOVE_ENV_SIMNET_PASSES {
+            for environment in CHECK_ENVIRONMENTS {
                 let ((deployment, _, artifacts), found_env_simnet) =
                     load_deployment_and_artifacts_or_exit(
                         &manifest,
                         &cmd.deployment_plan_path,
                         cmd.use_on_disk_deployment_plan,
                         cmd.use_computed_deployment_plan,
-                        force_remove_env_simnet,
+                        environment,
                     );
                 global_found_env_simnet |= found_env_simnet;
 
@@ -1284,12 +1288,7 @@ pub fn main() {
                                 Some((path.to_string_lossy().to_string(), diags))
                             })
                             .collect();
-                        let environment = if force_remove_env_simnet {
-                            Environment::OnChain
-                        } else {
-                            Environment::Simnet
-                        }
-                        .to_string();
+                        let environment = environment.to_string();
                         let output = JsonCheckOutput {
                             success: artifacts.success,
                             diagnostics,
@@ -1304,10 +1303,13 @@ pub fn main() {
                     }
                     OutputFormat::Standard => {
                         if global_found_env_simnet {
-                            if !force_remove_env_simnet {
-                                println!("Checking contracts with #[env(simnet)] code");
-                            } else {
-                                println!("Checking contracts without #[env(simnet)] code");
+                            match environment {
+                                Environment::Simnet => {
+                                    println!("Checking contracts with #[env(simnet)] code");
+                                }
+                                Environment::OnChain => {
+                                    println!("Checking contracts without #[env(simnet)] code");
+                                }
                             }
                         }
                         let diags_digest = DiagnosticsDigest::new(&artifacts.diags, &deployment);
@@ -1521,7 +1523,7 @@ pub fn load_deployment_and_artifacts_or_exit(
     deployment_plan_path: &Option<String>,
     force_on_disk: bool,
     force_computed: bool,
-    force_remove_env_simnet: bool,
+    environment: Environment,
 ) -> (
     (
         DeploymentSpecification,
@@ -1544,7 +1546,7 @@ pub fn load_deployment_and_artifacts_or_exit(
             .and_then(|opt| match opt {
                 Some(mut deployment) => {
                     eprintln!("{} using {default_deployment_file}", yellow!("note:"));
-                    if force_remove_env_simnet {
+                    if environment == Environment::OnChain {
                         found_env_simnet |= deployment.remove_env_simnet().unwrap_or(false);
                     }
                     let artifacts = setup_session_with_deployment(manifest, &mut deployment, None);
@@ -1556,7 +1558,7 @@ pub fn load_deployment_and_artifacts_or_exit(
                             manifest,
                             &StacksNetwork::Simnet,
                             false,
-                            force_remove_env_simnet,
+                            environment,
                         )?;
                     found_env_simnet |= gen_found_env_simnet;
                     if ast_artifacts.success {
@@ -1584,7 +1586,7 @@ pub fn load_deployment_and_artifacts_or_exit(
             let deployment_location = project_root.join(path);
             load_deployment(project_root, &deployment_location)
                 .map(|mut deployment| {
-                    if force_remove_env_simnet {
+                    if environment == Environment::OnChain {
                         found_env_simnet |= deployment.remove_env_simnet().unwrap_or(false);
                     }
                     let artifacts = setup_session_with_deployment(manifest, &mut deployment, None);
@@ -1665,7 +1667,7 @@ fn load_deployment_if_exists(
     }
 
     if !force_on_disk {
-        match generate_default_deployment(manifest, network, true, false) {
+        match generate_default_deployment(manifest, network, true, Environment::Simnet) {
             Ok((deployment, _, _)) => {
                 use similar::{ChangeTag, TextDiff};
 
@@ -2130,7 +2132,7 @@ fn devnet_start(cmd: DevnetStart, clarinetrc: ClarinetRC) {
                         &manifest,
                         &StacksNetwork::Devnet,
                         false,
-                        false,
+                        Environment::Simnet,
                     )
                     .unwrap_or_else(|message| {
                         eprintln!("{}", red!(message));
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -19,7 +19,7 @@ use clarity_repl::clarity::vm::types::{QualifiedContractIdentifier, StandardPrin
 use clarity_repl::clarity::vm::EvaluationResult;
 use clarity_repl::clarity::{ClarityName, ClarityVersion, StacksEpochId, SymbolicExpression};
 use clarity_repl::repl::ContractDeployer;
-use clarity_repl::utils::REMOVE_ENV_SIMNET_PASSES;
+use clarity_repl::utils::CHECK_ENVIRONMENTS;
 use clarity_static_cost::static_cost::StaticCost;
 use ls_types::{
     CompletionItem, DocumentSymbol, Hover, Location, MessageType, Position, Range, SignatureHelp,
@@ -730,16 +730,14 @@ impl ProtocolState {
 
 pub use clarity_repl::utils::Environment;
 
-fn tag_chain(remove_env_simnet: bool, found_env_simnet: bool, diagnostics: &mut Vec<Diagnostic>) {
+fn tag_diagnostics(
+    environment: Environment,
+    found_env_simnet: bool,
+    diagnostics: &mut Vec<Diagnostic>,
+) {
     if found_env_simnet {
-        let env = if remove_env_simnet {
-            Environment::OnChain
-        } else {
-            Environment::Simnet
-        };
-
         for ref mut diag in diagnostics {
-            let _ = write!(diag.message, " ({env})");
+            let _ = write!(diag.message, " ({environment})");
         }
     }
 }
@@ -771,14 +769,14 @@ pub async fn build_state(
 
     let mut global_found_env_simnet = false;
     let mut session = initiate_session_from_manifest(&manifest);
-    for force_remove_env_simnet in REMOVE_ENV_SIMNET_PASSES {
+    for environment in CHECK_ENVIRONMENTS {
         let (deployment, mut artifacts, found_env_simnet) = generate_default_deployment(
             &manifest,
             &StacksNetwork::Simnet,
             false,
             file_accessor,
             None,
-            force_remove_env_simnet,
+            environment,
         )
         .await?;
         global_found_env_simnet |= found_env_simnet;
@@ -816,8 +814,8 @@ pub async fn build_state(
             match result {
                 Ok(mut execution_result) => {
                     if let Some(entry) = artifacts.diags.get_mut(&contract_id) {
-                        tag_chain(
-                            force_remove_env_simnet,
+                        tag_diagnostics(
+                            environment,
                             global_found_env_simnet,
                             &mut execution_result.diagnostics,
                         );
@@ -837,11 +835,7 @@ pub async fn build_state(
                 }
                 Err(ref mut diags) => {
                     if let Some(entry) = artifacts.diags.get_mut(&contract_id) {
-                        tag_chain(
-                            force_remove_env_simnet,
-                            global_found_env_simnet,
-                            &mut *diags,
-                        );
+                        tag_diagnostics(environment, global_found_env_simnet, &mut *diags);
                         entry.append(diags);
                     }
                     continue;
--- a/components/clarity-repl/src/utils.rs
+++ b/components/clarity-repl/src/utils.rs
@@ -12,9 +12,9 @@ use strum::{Display, EnumString};
 use crate::analysis::annotation::AnnotationKind;
 use crate::repl::clarity_values::value_to_string;
 
-pub const REMOVE_ENV_SIMNET_PASSES: [bool; 2] = [true, false];
+pub const CHECK_ENVIRONMENTS: [Environment; 2] = [Environment::OnChain, Environment::Simnet];
 
-#[derive(Debug, EnumString, Display, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, Copy, EnumString, Display, PartialEq, Eq, Hash)]
 #[strum(serialize_all = "lowercase")]
 pub enum Environment {
     OnChain,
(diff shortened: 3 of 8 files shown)


xoloki-stacks:
The current implementation of `strip_env_simnet_exprs` is recursive.  I made a linearized version of it to see if that was causing the LSP freezing issue, but as we found it was unrelated.  I think the recursive version is cleaner, but its not tail-recursive so it's possible it could cause problems.  Lmk if you'd prefer the linearized version @hugo-stacks @brady-stacks

hugo-stacks:
I follow your lead @xoloki-stacks. I think the recursive approach is good.
I think there's a hard limit on how deeply nested clarity can be (not sure of the exact value), so we could eventually add an exit condition on this number of recursions for safety

components/clarity-lsp/src/common/backend.rs
```diff
@@ -732,6 +732,191 @@ mod lsp_tests {
...
 
+    struct TestFileAccessor {
+        project_manifest: String,
+        network_manifest: String,
+        contract: String,
+    }
+
+    impl TestFileAccessor {
+        fn new(contract: String) -> Self {
+            let project_manifest = r#"
```
jbencin-stacks: We've been trying to use `indoc!` for multiline strings to keep the code readable. Please do that here

components/clarity-repl/src/utils.rs
```diff
@@ -84,35 +85,60 @@ pub fn remove_env_simnet(source: String) -> Result<String, String> {
...
+
+    if !found {
+        Ok((source, false))
+    } else {
+        let mut source = String::new();
+        for line in lines {
+            if let Some(line) = line {
+                source.push_str(line);
+            }
+            source.push('\n');
```
jbencin-stacks, line 111: If I understand this we are replacing the stripped code with blank lines. Is this what we want?

components/clarity-lsp/src/common/state.rs
```diff
@@ -640,69 +643,117 @@ pub async fn build_state(
...
+                    };
+                }
+                Err(ref mut diags) => {
+                    if let Some(entry) = artifacts.diags.get_mut(&contract_id) {
+                        if global_found_env_simnet {
+                            let suffix = if force_remove_env_simnet {
+                                " (onchain)"
+                            } else {
+                                " (simnet)"
+                            };
```
jbencin-stacks: This code appears multiple times. Could be a helper function, or could use an `enum` instead of `bool` and implement `Display`

components/clarity-lsp/src/common/backend.rs
xoloki-stacks: Done

components/clarity-repl/src/utils.rs
xoloki-stacks: Yes, that way we keep the original line numbers intact, which means the diagnostics' spans are correct even after stripping.  This is necessary because in the user's editor, the code will not be removed.

components/clarity-lsp/src/common/state.rs
xoloki-stacks: Done

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1252,68 +1255,92 @@ pub fn main() {
...
-                    };
-                    println!("{json}");
-                }
-                OutputFormat::Standard => {
-                    let diags_digest = DiagnosticsDigest::new(&artifacts.diags, &deployment);
-                    if diags_digest.has_feedbacks() {
-                        println!("{}", diags_digest.message);
+            let mut exit_codes = Vec::new();
+            let mut global_found_env_simnet = false;
+            for force_remove_env_simnet in [true, false] {
```
jbencin-stacks: It seems the order is important here, we want to do `false` second so we end up with the full AST with all the code. We also use the same order in state.rs. Can we make this a constant with a comment about why the order is the way it is:

```rust
/// This ensures the AST has ALL the code at the end of the second pass
const REMOVE_ENV_SIMNET_PASSES: [bool; 2] = [true, false];
```

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1252,68 +1255,92 @@ pub fn main() {
...
+                        let diagnostics: HashMap<String, Vec<Diagnostic>> = artifacts
+                            .diags
+                            .into_iter()
+                            .filter(|(_, diags)| !diags.is_empty())
+                            .filter_map(|(contract_id, diags)| {
+                                let (_, path) = deployment.contracts.get(&contract_id)?;
+                                Some((path.to_string_lossy().to_string(), diags))
+                            })
+                            .collect();
+                        let environment = if force_remove_env_simnet {
```
jbencin-stacks: Can this share code with `tag_chain()` so we're not hardcoding the same string multiple times?

components/clarity-lsp/src/common/backend.rs
```diff
@@ -732,37 +733,374 @@ mod lsp_tests {
...
+
+        let contract_location = PathBuf::from("test.clar".to_string());
+        let notification = LspNotification::ContractSaved(contract_location);
+        let response =
+            process_notification(notification, &mut editor_state_input, Some(&file_accessor))
+                .await
+                .expect("Failed to process notification");
+
+        let response_json = json!(response);
+        println!("full respeonse: {response_json}");
```
jbencin-stacks: Same issue on line 1005:

```suggestion
        println!("full response: {response_json}");
```

components/clarity-lsp/src/common/backend.rs
```diff
@@ -732,37 +733,374 @@ mod lsp_tests {
...
+        assert_eq!(message, "use of unresolved function 'increment' (onchain)");
+
+        let spans = diagnostic
+            .get("spans")
+            .expect("Failed to find \"spans\": in diagnostic");
+        let Some(spans_array) = spans.as_array() else {
+            panic!("spans should be an array");
+        };
+        assert_eq!(
+            diagnostics_array.len(),
```
jbencin-stacks: Should this be `spans_array` to match the message?

components/clarity-lsp/src/common/backend.rs
```diff
@@ -732,37 +733,374 @@ mod lsp_tests {
...
+        );
+
+        let spans = diagnostic
+            .get("spans")
+            .expect("Failed to find \"spans\": in diagnostic");
+        let Some(spans_array) = spans.as_array() else {
+            panic!("spans should be an array");
+        };
+        assert_eq!(
+            diagnostics_array.len(),
```
jbencin-stacks: Should this be `spans_array` to match the message?

components/clarity-lsp/src/common/backend.rs
```diff
@@ -732,37 +733,374 @@ mod lsp_tests {
...
+            .get("end_line")
+            .expect("span didn't have a \"end_line\" field");
+
+        let end_column1 = span1
+            .get("end_column")
+            .expect("span didn't have a \"end_column\" field");
+
+        assert_eq!(
+            start_line1,
+            &Value::Number(Number::from_string_unchecked("21".to_string()))
```
jbencin-stacks: How about this instead:

```suggestion
            &Value::Number(Number::from("21"))
```

Several of these `from_string_unchecked()` calls can be replaced

components/clarinet-deployments/src/types.rs
```diff
@@ -1255,6 +1255,33 @@ impl DeploymentSpecification {
         }
         self.sort_batches_by_epoch();
     }
+
+    pub fn remove_env_simnet(&mut self) -> bool {
+        let mut global_found_env_simnet = false;
+        for ref mut batch in self.plan.batches.iter_mut() {
+            for ref mut transaction in batch.transactions.iter_mut() {
```
jbencin-stacks: Can remove the `ref mut`:

```suggestion
        for batch in self.plan.batches.iter_mut() {
            for transaction in batch.transactions.iter_mut() {
```

components/clarity-repl/src/utils.rs
```diff
@@ -84,35 +85,60 @@ pub fn remove_env_simnet(source: String) -> Result<String, String> {
...
                 lines[(i - 1) as usize] = None;
             }
             if !matches!(expr.pre_expr, Comment(_)) {
                 found_env_simnet = false;
             }
+            continue;
         }
 
         if let Comment(comment) = &expr.pre_expr {
             if comment.contains("#[env(simnet)]") {
```
jbencin-stacks: Is there a way to do this using the existing annotation parsing code in annotations.rs?

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1322,7 +1349,11 @@ pub fn main() {
                     DeveloperUsageDigest::new(&manifest.project.name, &manifest.project.authors),
                 ));
             }
-            std::process::exit(exit_code);
+            if exit_codes.contains(&1) {
+                std::process::exit(1);
+            } else {
+                std::process::exit(0);
+            }
```
jbencin-stacks: If `exit_codes` is `[0, 2]` this will return 0. How about using the first non-zero exit code:


```suggestion
            let exit_code = exit_codes.iter().cloned().find(|c| *c != 0).unwrap_or(0);
            std::process::exit(exit_code);
```

components/clarinet-deployments/src/types.rs
```diff
@@ -1255,6 +1255,33 @@ impl DeploymentSpecification {
...
+                if let TransactionSpecification::EmulatedContractPublish(ref mut spec) = transaction
+                {
+                    let Ok((clean, found_env_simnet)) = remove_env_simnet(spec.source.to_string())
+                    else {
+                        return false;
```
jbencin-stacks: Consider propagating the error instead of returning `false`

components/clarity-lsp/src/common/state.rs
```diff
@@ -618,12 +619,29 @@ impl ProtocolState {
...
+fn tag_chain(remove_env_simnet: bool, found_env_simnet: bool, diagnostics: &mut Vec<Diagnostic>) {
+    if found_env_simnet {
+        let suffix = if remove_env_simnet {
+            " (onchain)"
+        } else {
+            " (simnet)"
+        };
+
+        for ref mut diag in diagnostics {
+            diag.message = diag.message.clone() + suffix;
```
jbencin-stacks: Avoid the extra allocations:

```suggestion
            diag.message.push_str(suffix);
```

components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -34,6 +35,16 @@ impl std::str::FromStr for AnnotationKind {
                         Ok(AnnotationKind::Allow(params))
                     }
                 }
+                "env" => {
+                    let network = value;
+                    if network.is_empty() {
```
jbencin-stacks: Would make more sense for this field to be an `enum` and use `strum` (already a dependency in `clarity-repl`) to derive `TryFrom<&str>`

components/clarity-repl/src/utils.rs
jbencin-stacks: Is it possible to remove the blank lines and update the spans?

I wouldn't say this is a blocker, but if so, you could make an issue for it and it can be done when we have time

components/clarity-repl/src/utils.rs
```diff
@@ -84,35 +85,60 @@ pub fn remove_env_simnet(source: String) -> Result<String, String> {
...
-        source.push_str(line);
-        source.push('\n');
+        // recurse into nested lists and tuples
+        match &expr.pre_expr {
+            ::clarity::vm::representations::PreSymbolicExpressionType::List(children)
+            | ::clarity::vm::representations::PreSymbolicExpressionType::Tuple(children) => {
+                global_found |= strip_env_simnet_exprs(children, lines);
+            }
+            _ => {}
+        }
```
jbencin-stacks: Default case is empty so should be `if`/`let`. Also, a `use` statement would clean this up:

```suggestion
        use ::clarity::vm::representations::PreSymbolicExpressionType as PSEType;
        if let PSEType::List(children) | PSEType::Tuple(children) = &expr.pre_expr {
            global_found |= strip_env_simnet_exprs(children, lines);
        };
```

components/clarity-repl/src/utils.rs
xoloki-stacks: I explored doing this before settling on leaving the lines blank.  I've added an issue https://github.com/stx-labs/clarinet/issues/2274 so this doesn't fall off the radar.

components/clarinet-cli/src/frontend/cli.rs
xoloki-stacks: Done

components/clarinet-cli/src/frontend/cli.rs
xoloki-stacks: Done

components/clarity-lsp/src/common/backend.rs
xoloki-stacks: `Number` doesn't `impl From` either `&str` or `String`, hence the admittedly ugly construction here.

components/clarity-lsp/src/common/backend.rs
xoloki-stacks: Fixed

components/clarity-lsp/src/common/backend.rs
xoloki-stacks: Fixed

components/clarity-lsp/src/common/backend.rs
xoloki-stacks: Fixed

components/clarinet-deployments/src/types.rs
xoloki-stacks: Good catch, I thought Rust was more pedantic and required the `ref mut`.

Fixed.

xoloki-stacks: (not committing suggestions because they end up with unverified commits)

components/clarity-repl/src/utils.rs
xoloki-stacks: I could conceivably `impl Display for AnnotationKind`, but that would require implementing all the variants.  Seems a bit overkill for this.

components/clarinet-cli/src/frontend/cli.rs
xoloki-stacks: Done, much cleaner

components/clarity-lsp/src/common/state.rs
xoloki-stacks: Since we're using a enum now, I still had to encode the space here, so we still have to allocate a `format!` but it is cleaner.

components/clarity-repl/src/utils.rs
xoloki-stacks: Done

components/clarinet-deployments/src/types.rs
xoloki-stacks: I originally did propagate but it caused a `panic!` in LSP code.  I'll try propagating again and fixing it at a higher level.

xoloki-stacks: I must have been thinking of a different function, I was able to propagate it and `unwrap_or` at the only two callsites.

components/clarity-repl/src/utils.rs
xoloki-stacks: Done

components/clarity-lsp/src/common/backend.rs
xoloki-stacks: Oh it looks like `Value` itself has `impl From<&str>`

xoloki-stacks: Done

components/clarity-repl/src/analysis/annotation.rs
xoloki-stacks: Done

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1252,68 +1257,93 @@ pub fn main() {
...
+                        let diagnostics: HashMap<String, Vec<Diagnostic>> = artifacts
+                            .diags
+                            .into_iter()
+                            .filter(|(_, diags)| !diags.is_empty())
+                            .filter_map(|(contract_id, diags)| {
+                                let (_, path) = deployment.contracts.get(&contract_id)?;
+                                Some((path.to_string_lossy().to_string(), diags))
+                            })
+                            .collect();
+                        let environment = if force_remove_env_simnet {
```
jbencin-stacks: Does it make sense to propagate the enum back further instead of using the boolean `force_remove_env_simnet`?

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -1322,6 +1352,7 @@ pub fn main() {
                     DeveloperUsageDigest::new(&manifest.project.name, &manifest.project.authors),
                 ));
             }
+            let exit_code = exit_codes.iter().cloned().find(|c| *c != 0).unwrap_or(0);
```
jbencin-stacks, line 1358: Could add a comment here:

```rust
// Returns first non-zero exit code, if any
```

components/clarity-lsp/src/common/state.rs
```diff
@@ -618,12 +620,43 @@ impl ProtocolState {
...
+fn tag_chain(remove_env_simnet: bool, found_env_simnet: bool, diagnostics: &mut Vec<Diagnostic>) {
+    if found_env_simnet {
+        let env = if remove_env_simnet {
+            Environment::OnChain
+        } else {
+            Environment::Simnet
+        };
+
+        for ref mut diag in diagnostics {
+            diag.message.push_str(&format!(" ({env})"));
```
jbencin-stacks: Could do:

```suggestion
            write!(diag.message, " ({env})").unwrap();
```

To get rid of the allocation

components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -58,6 +65,12 @@ impl std::str::FromStr for AnnotationKind {
     }
 }
 
+#[derive(Debug, EnumString, PartialEq, Eq, Hash)]
+#[strum(serialize_all = "snake_case")]
+pub enum NetworkKind {
```
jbencin-stacks: How about combining this and `Environment` enum into a single type, and using `strum` to implement `Display`:

```rust
#[derive(Display, Debug, EnumString, PartialEq, Eq, Hash)]
pub enum Environment {
    #[strum(serialize = "onchain")]
    OnChain,
    #[strum(serialize = "simnet")]
    Simnet,
}
```

components/clarinet-cli/src/frontend/cli.rs
xoloki-stacks: Do you mean propagate it down, as in pass the enum into `load_deployment_and_artifacts_or_exit`/`generate_default_deployment` etc?  That might make some things places cleaner, but in others we'd need to encode the logic about what to do based on the enum.

I could go either way.

components/clarinet-cli/src/frontend/cli.rs
xoloki-stacks: Done

components/clarity-lsp/src/common/state.rs
xoloki-stacks: Ok this is just wild; I asked `claude` if this would really get rid of the allocation, but didn't say anything about which callsite:
```
does the write! macro, if passed a format string, do an allocation?
```
Claude realized immediately that this was the callsite I was asking about :O That's craxy levels of inference.

For the record, this is the response:
```
⏺ Yes. write!(f, " ({env})") with a format argument allocates a temporary String internally via format_args!. But write!(f, " (onchain)") with a     
  plain string literal does not — the compiler optimizes it to a direct write.                                                                       
                                                                                                                                                     
  However, for Display impls writing to a Formatter, the allocation is in the formatting machinery, not an extra heap allocation — it writes directly
   to the underlying buffer. So write!(f, "onchain") is zero-alloc.                                                                                  
                  
  The place where allocation matters more is the call site in tag_chain:

  diag.message.push_str(&format!(" ({env})"));  // allocates a temporary String

  vs.

  write!(diag.message, " ({env})").unwrap();  // writes directly into the existing String

  String implements std::fmt::Write, so write! appends directly without a temporary allocation.
```

xoloki-stacks: I don't want to actually make this change though, because it's adding an `unwrap` inside LSP library code, and we're trying to get rid of all of those.

components/clarity-repl/src/analysis/annotation.rs
xoloki-stacks: I'm not sure I want to do that, since it implies the possibility of `#[env(onchain)]`.  I think the code is cleaner as is.

components/clarinet-cli/src/frontend/cli.rs
jbencin-stacks: I mean just use an enum in place of a `bool` everywhere. You're call, not really necessary

components/clarity-lsp/src/common/state.rs
jbencin-stacks: There's an `unwrap()`, but `write!()` can't actually return an error in this context. That could happen if writing to a file or network socket,which is why this macro returns a `Result`, but not a `String`

`write!()` can fail in the same way `format!()` can, when OOM, but then it just panics

Also can do:

```rust
_ = write!(diag.message, " ({env})");
```

To just ignore the result

components/clarity-repl/src/analysis/annotation.rs
jbencin-stacks: You could just fail on  `#[env(onchain)]` with an message that it's not supported

components/clarity-lsp/src/common/state.rs
xoloki-stacks: Done, went with the `let _ = ` approach so it doesn't trigger anything when we do the `unwrap` fix.

components/clarity-repl/src/analysis/annotation.rs
xoloki-stacks: Yeah I was more worried about the implications to a reader of the code, not how to do it in the code.  But I tried it out and it doesn't look so bad.

Done

components/clarinet-cli/src/frontend/cli.rs
xoloki-stacks: Yeah that's what I was trying to say :P I'll see how it looks when claude does it and decide if I want to keep it.

xoloki-stacks: Agreed it does look better this way.

Done

jbencin-stacks (approved):

PR #2258, opened 2026-03-05
Never lint boot contracts, and don't run linter on commands like `clarinet console` and `npm test`

### LINKED ISSUE #2253: Do not run analysis on boot contracts
The linter currently runs on boot contracts.

It's invisible to the users because we don't collect the diagnostics. And it runs fairly quickly.
But it mean that it's run all the time for every simnet session for the same boot contracts.

We should avoid running it (both the linter and check_checker fwiw)

## Reproduce / Observe

:one: Add a log around here
https://github.com/stx-labs/clarinet/blob/6a45e07d863fc8593061679511b6bc5e99712624/components/clarity-repl/src/analysis/lints/unused_data_var.rs#L152
```rs
            println!("unused: {}", message);
```

:two: Run `clarinet check` in any clarinet project (even an empty one)

3️⃣ Observe the some lints triggered in boot contracts
```
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-pox-4-reward-cycle` never read
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-pox-4-reward-cycle` never read
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-pox-4-reward-cycle` never read
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-2-1-reward-cycle` never read
unused: data variable `first-pox-4-reward-cycle` never read
unused: data variable `cost` never modified
unused: data variable `contract-owner` never modified
```

### LINKED ISSUE #2254: Only run analysis when need (check and lsp)
The contract analysis (lint and checker) run on every simnet session.
Meaning that even `clarinet console` (or `npm test`) will spend some time running the linter on every contract before running.
Even though it won't do anything with the diagnostics.

Our code base tries to minimize the code paths and every simnet session sort of follow the same path, but it leads to this kind of scenarios where un-needed functions get called

## Reproduce / Observe

It can be reproduced in the same way as https://github.com/stx-labs/clarinet/issues/2253.
Except, run `$ clarinet console` instead of `$ clarinet check`

jbencin-stacks:
### Description

We don't want to waste CPU cycles running the linter and analysis passes on the boot contracts, or when we are going to ignore the output. This includes running `clarinet console` and `npm test`

This PR does two things:
  - Changes the `is-requirement` flag in `ClarityContract` to `skip_analysis`, and sets it to `true` for boot contracts
  - Add an `enable_analysis` flag to `setup_session_with_deployment()`, to skip all analysis for `clarinet console` and `npm test`

#### Breaking change?

No

### Related Issues

- Fixes #2253 
- Fixes #2254 

### Checklist


components/clarinet-deployments/src/lib.rs
```diff
@@ -1159,6 +1164,210 @@ mod tests {
...
+        session.interpreter.repl_settings.analysis.disable_all();
+
+        let spec = EmulatedContractPublishSpecification {
+            contract_name: ContractName::from("lint-test"),
+            emulated_sender: PrincipalData::parse_standard_principal(DEPLOYER).unwrap(),
+            source: LINT_TRIGGERING_SOURCE.to_string(),
+            clarity_version: ClarityVersion::Clarity2,
+            location: PathBuf::from("/contracts/lint-test.clar"),
+            skip_analysis: false, // Even with skip_analysis=false, disable_all() prevents lints
+        };
```
brady-stacks, line 1278: Could probably make this into a helper function since it's used in many of these tests

components/clarinet-deployments/src/lib.rs
```diff
@@ -41,23 +41,28 @@ pub fn setup_session_with_deployment(
     manifest: &ProjectManifest,
     deployment: &mut DeploymentSpecification,
     contracts_asts: Option<&BTreeMap<QualifiedContractIdentifier, ContractAST>>,
+    enable_analysis: bool,
```
xoloki-stacks, line 46: So you call this field `enable_analysis` here and `skip_analysis` elsewhere, because here it's a hint and there it's enforced?

jbencin-stacks: It's `enable_analysis` on the project level, and `skip_analysis` on the contract level. Really just because the logic for `skip_analysis` matched the logic of the `is_requirement` flag which it replaced

jbencin-stacks: And yeah, it's also true that `enable_anaylsis` won't necessarily run the analysis on all files, but `skip_analysis` will always skip it

PR #2273, opened 2026-03-10
feat: codelens config enhancements

### LINKED ISSUE #2271: Costs CodeLens follow up
- New configs:
  - [ ] Ability to show exact number or percentage of budget used
  - [ ] Only show the code lens if a function exceeds a certain percentage, default to something like 10%, but configurable
- [ ] Proper traits handling

brady-stacks:
- partially closes #2271 

Adds the additional configs for percentage view and percentage gating

Screenshot 2026-03-10 at 11 58 31 AM

Note: defaulted the gating to 0 because I assume it would be confusing to enable the codelens and then see nothing for most functions (if you didn't see the percentage gating option)

### DIFF
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -808,6 +808,7 @@ pub async fn build_state(
     manifest_location: &Path,
     protocol_state: &mut ProtocolState,
     file_accessor: Option<&dyn FileAccessor>,
+    static_cost_analysis: bool,
 ) -> Result<(), String> {
     let mut locations = HashMap::new();
     let mut asts = BTreeMap::new();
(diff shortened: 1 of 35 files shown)


components/clarity-lsp/src/common/state.rs
```diff
@@ -525,17 +563,41 @@ impl EditorState {
...
+                    "runtime: {:.1}%, read count: {:.1}%, read length: {:.1}%, write count: {:.1}%, write length: {:.1}%",
+                    pct.runtime, pct.read_count, pct.read_length, pct.write_count, pct.write_length,
+                )
+            }
+        }
+    }
+
+    fn exceeds_threshold(cost_node: &StaticCost, threshold: f64) -> bool {
+        let pct = cost_percentages(&cost_node.max);
+        [
```
xoloki-stacks: I wonder if this might be better as `pub fn exceeds_threshold` on `CostPercentages`

brady-stacks: Done. Good call

PR #2278, opened 2026-03-11
feat: latest version check for cli

### LINKED ISSUE #2276: Inform user when a new version of Clarinet is available
When a developer uses the Clarinet CLI, it should periodically check if it's using the latest Clarinet version.

It can be done by fetching the latest release on github
```
curl -s https://api.github.com/repos/stx-labs/clarinet/releases/latest | jq -r '.tag_name'
```

I think there is one edge case: when a release is created on github, it can take a few minutes (or an hour) to be available everywhere

brady-stacks:
- closes #2276 

- Check if the user's cli version is up-to-date
- will only check at most every 72 hours

```
A new release of clarinet is available: 3.14.1 -> 3.14.2                                                                                                      
```

Screenshot 2026-03-12 at 12 15 44 PM

### DIFF
--- a/components/clarinet-cli/src/frontend/cli.rs
+++ b/components/clarinet-cli/src/frontend/cli.rs
@@ -32,6 +32,7 @@ use clarity::vm::types::QualifiedContractIdentifier;
 use clarity::vm::ClarityVersion;
 use clarity_lsp::state::Environment;
 use clarity_repl::analysis::call_checker::ContractAnalysis;
+use clarity_repl::analysis::linter::{LintGroup, LintName};
 use clarity_repl::frontend::Terminal;
 use clarity_repl::repl::diagnostic::output_diagnostic;
 use clarity_repl::repl::settings::{ApiUrl, RemoteDataSettings};
@@ -42,6 +43,7 @@ use clarity_repl::utils::CHECK_ENVIRONMENTS;
 use clarity_repl::{analysis, repl};
 use serde::Serialize;
 use stacks_network::{self, DevnetOrchestrator};
+use strum::{EnumMessage, VariantArray};
 use toml_edit::DocumentMut;
 
 #[cfg(feature = "telemetry")]
@@ -560,6 +562,9 @@ struct Check {
         conflicts_with = "use_on_disk_deployment_plan"
     )]
     pub use_computed_deployment_plan: bool,
+    /// List available lints and lint groups
+    #[clap(long = "show-lints")]
+    pub show_lints: bool,
     /// Set the output format
     #[clap(long = "output", default_value = "standard")]
     pub output_format: OutputFormat,
@@ -1193,6 +1198,9 @@ pub fn main() {
                 display_contract_new_hint(None);
             }
         }
+        Command::Check(cmd) if cmd.show_lints => {
+            print_available_lints();
+        }
         Command::Check(cmd) if cmd.file.is_some() => {
             let file = cmd.file.unwrap();
             let mut settings = repl::SessionSettings {
@@ -1479,6 +1487,46 @@ pub fn main() {
     };
 
     update_check::print_update_message(update_handle);
+}
+
+fn print_available_lints() {
+    println!("Lint groups:");
+    println!();
+    for group in LintGroup::VARIANTS {
+        println!(
+            "  {group:<12} {}",
+            group.get_documentation().unwrap_or_default()
+        );
+    }
+
+    println!();
+    println!("Lints:");
+    println!();
+    for lint in LintName::VARIANTS {
+        let groups: Vec<_> = LintGroup::of(lint)
+            .into_iter()
+            .map(ToString::to_string)
+            .collect();
+        let groups = format!("[{}]", groups.join(", "));
+        println!(
+            "  {lint:<28} {groups:<12} {}",
+            lint.get_documentation().unwrap_or_default()
+        );
+    }
+
+    indoc::printdoc! {r#"
+
+        Configure lints in Clarinet.toml:
+
+          [repl.analysis.lint_groups]
+          style = "warning"
+
+          [repl.analysis.lints]
+          unused_const = "error"
+          at_block = false
+
+        Suppress a lint in source code with: ;; #[allow(lint_name)]
+    "#};
 }
 
 fn overwrite_formatted(file_path: &str, output: &str) -> io::Result<()> {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -664,6 +664,7 @@ dependencies = [
  "serde_json",
  "similar",
  "stacks-network",
+ "strum",
  "tempfile",
  "tikv-jemallocator",
  "tokio",
--- a/components/clarinet-cli/Cargo.toml
+++ b/components/clarinet-cli/Cargo.toml
@@ -28,6 +28,7 @@ serde_json = { workspace = true, features = ["preserve_order"] }
 similar = "2.1.0"
 reqwest = { workspace = true, features = ["blocking"] }
 semver = { workspace = true }
+strum = { workspace = true }
 tokio = { workspace = true }
 toml_edit = { workspace = true }
 
--- a/components/clarity-repl/src/analysis/linter.rs
+++ b/components/clarity-repl/src/analysis/linter.rs
@@ -4,7 +4,7 @@ use clarity_types::diagnostic::Level as ClarityDiagnosticLevel;
 #[cfg(feature = "json_schema")]
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
-use strum::{EnumString, VariantArray};
+use strum::{EnumMessage, EnumString, VariantArray};
 
 use crate::analysis::annotation::Annotation;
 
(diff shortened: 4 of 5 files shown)


components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -0,0 +1,78 @@
...
+        let minor = parts.next()?.parse().ok()?;
+        let patch = parts.next()?.split('-').next()?.parse().ok()?;
+        Some((major, minor, patch))
+    };
+
+    match (parse(latest), parse(current)) {
+        (Some(l), Some(c)) => l > c,
+        _ => false,
+    }
+}
```
hugo-stacks: Can it be replaced with the semver crates? It's already in our dependencies

hugo-stacks (changes requested):
I wanted to link the old issue in the issue desc, it had some good ideas:
- don't check every time to avoid being to noisy https://github.com/stx-labs/clarinet/issues/221#issuecomment-1365271346
- also, do we really want users to be able to disable it? If we only show it every 24 or 72 hours, or even less, maybe we could always show it (remove the config)
- I think we need to check when was the latest release published, if it was just a few minutes ago, some package managers won't be aware of it (see edge case in issue desc)

components/clarinet-cli/src/frontend/update_check.rs
brady-stacks: Yes!

brady-stacks:
@hugo-stacks I was overcomplicating the 72 hour check in my head. It is actually very straightforward. I updated it to use `~/.clarinet/.last_update_check`'s metadata and `touch` when it actually fetches the release ver.

As for the check when the release was published I think we'll be ok without checking that if we're only checking every 3 days. I could be convinced otherwise, lmk

I _did_ remove the ability to ignore the warning. It's almost definitely undesirable to set that `true` and then forget about it

components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -0,0 +1,145 @@
+use std::time::Duration;
+
+use clarinet_files::clarinetrc::ClarinetRC;
+
+const GITHUB_RELEASES_URL: &str = "https://api.github.com/repos/stx-labs/clarinet/releases/latest";
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(3);
+const CHECK_INTERVAL: Duration = Duration::from_secs(72 * 60 * 60);
```
hugo-stacks: Nit: There's a `from_hours`

components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -0,0 +1,193 @@
...
+        return;
+    };
+
+    let cache_path = config_dir.join(VERSION_CACHE_FILE);
+
+    let (latest, published_at) = if should_check(&cache_path) {
+        // Cache is missing or stale — fetch from GitHub (only happens every 72h)
+        let fetched = fetch_latest_version();
+        if let Some((ref version, ref published)) = fetched {
+            let cache_content = format!("{}\n{}", version, published);
```
jbencin-stacks: Inline the variables in this PR

components/clarinet-cli/src/frontend/cli.rs
```diff
@@ -593,6 +594,8 @@ pub fn main() {
 
     let clarinetrc = ClarinetRC::from_rc_file();
 
+    update_check::check_for_update();
```
jbencin-stacks: This is a blocking network request. Seems like a dangerous thing to put this in front of all CLI commands. If GitHub  or the user's network is having issues it breaks the CLI.

Can we do this in a separate thread? Before the CLI exits, check and print the result of that thread. If we don't have a result yet, wait like a second or two for it, and if we still don't have it, just exit or print a failure message

brady-stacks: Agreed. It's a bit less concise but definitely worth doing. I set it to wait at least 1 second for github fetch to return. I'd think that's enough. Anything over that seems likely to be a timeout

components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -0,0 +1,197 @@
...
+            .unwrap_or((None, None))
+    } else {
+        read_cached_version(&cache_path)
+    };
+
+    let latest = latest?;
+    if is_newer(&latest, current_version) && !is_within_cooldown(published_at.as_deref()) {
+        Some(format!(
+            "\nA new release of clarinet is available: {} -> {}",
+            current_version, latest,
```
jbencin-stacks: Inline

```suggestion
            "\nA new release of clarinet is available: {current_version} -> {latest}"
```

components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -0,0 +1,197 @@
...
+const VERSION_CACHE_FILE: &str = ".latest_version";
+
+/// Spawn the update check on a background thread so it doesn't block CLI startup.
+/// The message (if any) is printed to stderr from the background thread.
+/// The HTTP request has a 1-second timeout; if it doesn't complete in time,
+/// or if the process exits first, the check is silently skipped.
+pub fn check_for_update_async() {
+    std::thread::spawn(|| {
+        if let Some(message) = check_for_update() {
+            eprintln!("{message}");
```
jbencin-stacks: I think we should return the message and print from the main thread, AFTER the output of the command. This ensures that:
- The user sees it
- It doesn't appear in the middle of the command output
- Let's you print a message also if the HTTP request timed out

components/clarinet-files/src/clarinetrc.rs
```diff
@@ -13,6 +13,10 @@ impl ClarinetRC {
         "~/.clarinet/clarinetrc.toml"
     }
 
+    pub fn get_config_dir() -> Option<std::path::PathBuf> {
+        dirs::home_dir().map(|h| h.join(".clarinet"))
```
jbencin-stacks: `get_settings_file_path()` and `from_rc_file()` should use this function also, as there should be a single canonical way to get the config directory. In particular, `get_settings_file_path()`s hardcoded string with `~` and `/`s looks like something that won't work on Windows

If this is a simple fix it can go in this PR, otherwise can be separate

components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -0,0 +1,205 @@
...
+
+use chrono::{DateTime, Utc};
+use clarinet_files::clarinetrc::ClarinetRC;
+
+const GITHUB_RELEASES_URL: &str = "https://api.github.com/repos/stx-labs/clarinet/releases/latest";
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(1);
+const CHECK_INTERVAL: Duration = Duration::from_hours(72);
+const RELEASE_COOLDOWN: Duration = Duration::from_hours(1);
+const VERSION_CACHE_FILE: &str = ".latest_version";
+const UPDATE_CHECK_WAIT: Duration = Duration::from_millis(100);
```
jbencin-stacks, line 17: Can you add doc comments to these?

components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -0,0 +1,205 @@
+use std::sync::mpsc;
+use std::time::Duration;
+
+use chrono::{DateTime, Utc};
+use clarinet_files::clarinetrc::ClarinetRC;
+
+const GITHUB_RELEASES_URL: &str = "https://api.github.com/repos/stx-labs/clarinet/releases/latest";
+const REQUEST_TIMEOUT: Duration = Duration::from_secs(1);
+const CHECK_INTERVAL: Duration = Duration::from_hours(72);
```
jbencin-stacks, line 11: Seems kinda long, I would think 12 or 24 hours would be appropriate

brady-stacks:
@jbencin-stacks 
> some way to disable this check and/or other network calls

I did have a setting to disable the checks initially but I remove it since it's a good idea to upgrade, but I could see something like DISABLE_NETWORK being a nice add. It would have to be called something else since that's a bit too ambiguous, but I like it. Added here #2297

components/clarinet-cli/src/frontend/update_check.rs
brady-stacks: Well, considering we release roughly once a week, the 72 hours seems appropriate for catching the upgrades

components/clarinet-files/src/clarinetrc.rs
brady-stacks: done 👍

components/clarinet-cli/src/frontend/update_check.rs
```diff
@@ -9,6 +9,7 @@ const REQUEST_TIMEOUT: Duration = Duration::from_secs(1);
 const CHECK_INTERVAL: Duration = Duration::from_hours(72);
 const RELEASE_COOLDOWN: Duration = Duration::from_hours(1);
```
jbencin-stacks, line 13: Needs a doc comment also. I'd put one on all the duration constants

PR #2279, opened 2026-03-11
fix(linter): Traverse Clarity 4's `as-contract?`

### LINKED ISSUE #2272: linter: False `unused_*` positives for items inside of `as-contract?`
**Describe the bug**

@obycode reported false positives for `unused_*` lints in the [stackflow](https://github.com/obycode/stackflow) project. I tried it myself and was able to confirm that there were in fact false positives and they all came from items that were only used inside of `as-contract?` expressions

**Explanation**

The new `as-contract?` introduced in Clarinet 4 is handled in `ASTVisitor`, which is used to implement the lints. However, it is handled incorrectly:

```rust
AsContract | AsContractSafe => {
    self.traverse_as_contract(expr, args.get(0).unwrap_or(&DEFAULT_EXPR))
}
```

```rust
fn traverse_as_contract(
    &mut self,
    expr: &'a SymbolicExpression,
    inner: &'a SymbolicExpression,
) -> bool {
    self.traverse_expr(inner) && self.visit_as_contract(expr, inner)
}
```

This works correctly for the old `as-contract`, but will pass the arguments to `as-contract?` as `inner`, and completely ignore the actual inner expression

jbencin-stacks:
### Description

Fixes #2272 

The new `as-contract?` in Clarity 4 was being handled the same as the old `as-contract`, which was incorrect

This PR updates `ASTVisitor` to traverse all the allowances and the full expression list, so that the `LintGroup::Unused` lints don't emit false positives for items used inside `as-contract?`

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarity-repl/src/analysis/check_checker/mod.rs
+++ b/components/clarity-repl/src/analysis/check_checker/mod.rs
@@ -528,40 +528,38 @@ impl<'a> ASTVisitor<'a> for CheckChecker<'a> {
         true
     }
 
-    fn traverse_as_contract(
+    fn pre_traverse_as_contract(
         &mut self,
-        expr: &'a SymbolicExpression,
-        inner: &'a SymbolicExpression,
-    ) -> bool {
+        _expr: &'a SymbolicExpression,
+        _inner: &'a SymbolicExpression,
+    ) {
         self.in_as_contract = true;
-        let res = self.traverse_expr(inner) && self.visit_as_contract(expr, inner);
-        self.in_as_contract = false;
-        res
     }
 
-    fn traverse_as_contract_safe(
+    fn post_traverse_as_contract(
         &mut self,
-        expr: &'a SymbolicExpression,
-        allowances: &'a SymbolicExpression,
-        body: &'a [SymbolicExpression],
-    ) -> bool {
-        self.in_as_contract = true;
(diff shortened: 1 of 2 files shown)


components/clarity-repl/src/analysis/check_checker/mod.rs
```diff
@@ -539,6 +539,31 @@ impl<'a> ASTVisitor<'a> for CheckChecker<'a> {
         res
     }
 
+    fn traverse_as_contract_safe(
+        &mut self,
+        expr: &'a SymbolicExpression,
+        allowances: &'a SymbolicExpression,
+        body: &'a [SymbolicExpression],
+    ) -> bool {
+        self.in_as_contract = true;
```
xoloki-stacks, line 553: Though the code is superficially different, this override appears to function exactly the same as in the trait definition in `ast_visitor.rs`, with the sole exception of setting `self.in_as_contract` at the beginning and end of the function.  Why not add pre/post hooks to the default method that are empty in the trait, and implement them in the override?

xoloki-stacks (changes requested):
Looks good though I have one suggestion.

components/clarity-repl/src/analysis/check_checker/mod.rs
jbencin-stacks: Done b506e6b

PR #2283, opened 2026-03-13
Check client capabilities before running cost analysis

xoloki-stacks:
### Description

After commit 59653687be1007ca10e079bb5485cec970a5dc59 implemented static cost analysis and `codeLens` support, the `emacs` `LSP` package `eglot` began to hang when any contract was opened.  This PR addresses that with several fixes:

1. Unconditional code_lens_refresh() - server-to-client request that `eglot` couldn't answer                          
1. Silently dropped error responses - `if let Ok(response)` meant any `Err` from `process_notification` left the handler blocked on `recv()` forever     
1. Unconditional cost analysis - expensive computation running for every client even when nobody asked for it, blocking the single-threaded runtime

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarinet-cli/src/lsp/mod.rs
+++ b/components/clarinet-cli/src/lsp/mod.rs
@@ -305,7 +305,8 @@ async fn run_did_open_and_collect_server_requests(code_lens_refresh: bool) -> Ve
     let contract_path = std::env::current_dir()
         .unwrap()
         .join("examples/counter/contracts/counter.clar");
-    let uri = format!("file://{}", contract_path.to_str().unwrap());
+    let uri =
+        clarinet_files::paths::path_to_url_string(&contract_path).expect("failed to build URI");
 
     // 3. Send didOpen and concurrently handle server-to-client requests
     let did_open = build_did_open_notification(&uri);


hugo-stacks:
Good catch 👏 

Looking at the code, I'm not 100% confident that codelens would _not_ be computed when
`settings.static_cost_analysis == false ` and `client_supports_code_lens_refresh == true`
or when
`settings.static_cost_analysis == true` and `client_supports_code_lens_refresh == false`

Could 
`es.settings.static_cost_analysis` and `client_supports_code_lens_refresh` collapsed into one boolean that conditions it everywhere?
I guess that would be implemented here but I'm not sure it can be done properly
https://github.com/stx-labs/clarinet/blob/4f6591816239ab0fc16718ec314f50c05d7cfe5a/components/clarity-lsp/src/common/backend.rs#L607-L624

Ideally, we want to avoid as much as possible to have custom logic in native_bridge and vscode_brige.

hugo-stacks:
Good catch 👏 

Looking at the code, I'm not 100% confident that codelens would _not_ be computed when
`settings.static_cost_analysis == false ` and `client_supports_code_lens_refresh == true`
or when
`settings.static_cost_analysis == true` and `client_supports_code_lens_refresh == false`

Could 
`es.settings.static_cost_analysis` and `client_supports_code_lens_refresh` collapsed into one boolean that conditions it everywhere?
I guess that would be implemented here

https://github.com/stx-labs/clarinet/blob/4f6591816239ab0fc16718ec314f50c05d7cfe5a/components/clarity-lsp/src/common/backend.rs#L607-L624

Ideally, we want to avoid as much as possible to have custom logic in native_bridge and vscode_brige.

xoloki-stacks:
@hugo-stacks there seems to be a lot of strangeness around code lenses in the current code.  I think we should merge this PR so we can fix clients that don't support code lenses, and then file an issue to dig deeper.

hugo-stacks:
> this PR so we can fix clients that don't support code lenses

Fair 👍 

I'll let @brady-stacks review this one.

Can you open the issue in the meantime? I'm curious about "there seems to be a lot of strangeness around code lenses in the current code"

xoloki-stacks:
> Can you open the issue in the meantime? I'm curious about "there seems to be a lot of strangeness around code lenses in the current code"

https://github.com/stx-labs/clarinet/issues/2287

The last paragraph in the issue describes the general issue which the code lens support exacerbated.

PR #2284, opened 2026-03-13
Add a strength option for key derivation when using encrypted mnemonics

xoloki-stacks:
### Description

The original `encrypted_mnemonic` support was tuned for performance rather than security.  It was sufficient for strong passwords, but insufficient for weak passwords guarding wallets of high value.  

This PR adds `enum MnemonicEncryptionStrength` with variants `default`/`medium`/`high`, and a separate `TOML` keys for each.  The non-default variants will configure the `argon2` key derivation to use a per-ciphertext random salt and higher memory/iterations, at the cost of latency when encrypting and decrypting the mnemonic.

#### Breaking change?

No, any current `encrypted_mnemonic` fields will behave as before.

### Example

---

### Checklist

### DIFF
--- a/components/clarinet-utils/src/lib.rs
+++ b/components/clarinet-utils/src/lib.rs
@@ -379,26 +379,48 @@ mod tests {
     fn test_encrypt_mnemonic() {
         let phrase = "twice kind fence tip hidden tilt action fragile skin nothing glory cousin green tomorrow spring wrist shed math olympic multiply hip blue scout claw";
         let password = "foo";
-        let strength = MnemonicEncryptionStrength::Default;
 
+        // Test Default and Medium with cross-strength mismatch checks
+        for strength in [
+            MnemonicEncryptionStrength::Default,
+            MnemonicEncryptionStrength::Medium,
+        ] {
+            let encrypted = encrypt_mnemonic_phrase(phrase, password, strength)
+                .expect("encrypt_mnemonic_phrase should succeed");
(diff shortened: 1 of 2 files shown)


components/clarinet-utils/src/lib.rs
```diff
@@ -104,40 +141,89 @@ impl From<std::str::Utf8Error> for MnemonicEncryptionError {
...
-pub fn derive_key(password: &str, buf: &mut [u8]) -> Result<(), EncryptionError> {
-    let salt = b"clarinet_utils-derive_key_salt";
+pub fn derive_key(
+    password: &str,
+    buf: &mut [u8],
+    strength: MnemonicEncryptionStrength,
+    salt: &[u8],
+) -> Result<(), EncryptionError> {
+    let argon2 = match strength {
+        MnemonicEncryptionStrength::Default => Argon2::default(),
```
brady-stacks, line 151: I wonder if making this explicit would be good `argon2::Params::new(19456, 2, 2, Some(32))` ?

brady-stacks: Maybe even renaming Default to Low and setting it as the default

xoloki-stacks: I had thought about naming this variant `Low`, but decided against it because I wanted it clear that this was what `Argon` considered a default.  Also since we'd already used the `TOML` field `encrypted_mnemonic` without a suffix, and we needed to stay backwards compatible.  Same with making the parameters explicit.

PR #2292, opened 2026-03-16
Add tests covering every diagnostic path in collect_annotations

xoloki-stacks:
### Description

During my last PR it became apparent that none of the annotation code paths were being tested.  This PR adds test for all of them.

#### Breaking change?

No

### Checklist

### DIFF
--- a/components/clarity-repl/src/repl/interpreter.rs
+++ b/components/clarity-repl/src/repl/interpreter.rs
@@ -2262,4 +2262,112 @@ mod tests {
             })),
         );
     }
+
+    #[test]
+    fn eol_annotation_produces_ignored_warning() {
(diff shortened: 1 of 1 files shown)


xoloki-stacks:
> Approving, but might make more sense to put these in `annotation.rs`. I don't feel strongly about it, so your call

I put the tests here because they need an `Interpreter` so they can call `collect_annotations`.  If I put the tests in `annotation.rs` that would create a circular dependency between the modules, since `interpreter.rs` needs `analysis::annotation::{Annotation, AnnotationKind}`.

Rust does allow cyclic dependencies between modules in the same crate, but it's discouraged.  We might be able to get away with it with clever use of test submodules, but this just seemed cleaner.

PR #2295, opened 2026-03-17
Fix/pox cycle length check

hugo-stacks:
### Description

The constant were set to the wrong value
```
pub const DEFAULT_POX_PREPARE_LENGTH: u64 = 4;
pub const DEFAULT_POX_REWARD_LENGTH: u64 = 10;
```

Compared to the actual config values:
https://github.com/stx-labs/clarinet/blob/2e0783a3cc00ca49aefecdfd4ce373105438fed8/components/stacks-network/src/orchestrator.rs#L948-L949

Fix: Fix the constant and use them in the config.

The issue wasn't _that_ since it was only used in the config check. It could have impacted developers using a custom epoch height


hugo-stacks:
I think the Devnet it self has so many default config that it makes sense to keep those where they are. But it could be useful to centralize all as well. No strong opinion

Issue #2296, opened 2026-03-17
Discoverablilty for lints and lint groups


jbencin-stacks:
From @hstove :

> how can i get a list of all clarinet lint rules? specifically the "string" key.
>
> It would be great if the LSP and Clarinet check included this in the message - that's often the case with other linters

There is currently no way to discover these without reading the code or release notes. We need to add something, options include:
- [ ] Adding `--lints=<lint>` and `--lint-groups=<lint-group>` options to `clarinet check`, so that the enum variants would show in `clarinet check --help`
- [x] Adding `--show-lints` flag
  - #2298 
  - #2314
- [x] Adding the lint names to the printed warnings
  - #2307
- [ ] Adding `clarinet manifest schema` command that outputs JSON schema for `Clarinet.toml`
  - #2300

We may do more than one of these

PR #2300, opened 2026-03-17
feat: Add `clarinet manifest schema` to generate JSON schema for `Clarinet.toml`

jbencin-stacks:
### Description

Add `clarinet manifest schema` to generate JSON schema for `Clarinet.toml`. This is a minor change, as all the heavy lifing is done by `serde` and `schemars`, which we already had here. This will allow users to validate their `Clarinet.toml` files without running Clarinet, and will give LLMs guidance on how to modify the files

Also adds `clarinet manifest check [file]` to validate a `Cargo.toml` file without running anything else

### Related Issues

- #2296 

### Checklist

### DIFF
--- a/components/clarinet-cli/tests/cli.rs
+++ b/components/clarinet-cli/tests/cli.rs
@@ -731,12 +731,12 @@ fn test_deployment_environment_mapping() {
 
 #[test]
 fn test_schema_validates_all_manifests() {
-    // Generate schema from `clarinet schema`
+    // Generate schema from `clarinet manifest schema`
     let output = Command::new(env!("CARGO_BIN_EXE_clarinet"))
-        .args(["schema"])
+        .args(["manifest", "schema"])
         .output()
-        .expect("Failed to execute clarinet schema");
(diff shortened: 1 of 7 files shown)


components/clarinet-cli/tests/cli.rs
```diff
@@ -728,3 +728,56 @@ fn test_deployment_environment_mapping() {
...
+        jsonschema::validator_for(&schema_json).expect("Schema itself is not a valid JSON Schema");
+
+    // Find all Clarinet.toml files in the repo
+    let workspace_root = Path::new(env!("CARGO_MANIFEST_DIR"))
+        .parent()
+        .unwrap()
+        .parent()
+        .unwrap();
+
+    let manifest_paths = [
```
brady-stacks, line 753: we could probably glob this so it picks up future contracts and doesn't fail if we remove some of these. like `clarinet-cli/examples/**/Clarinet.toml` and `clarity-vscode/test-data/**/Clarinet.toml`

PR #2328, opened 2026-03-25
feat: add documentSymbols capability to LSP, #2327

hstove-stacks:
- Fixes https://github.com/stx-labs/clarinet/issues/2327

Turns out this was a very simple change! I verified this by running `npm run dev` in `components/clarity-vscode`, and then in the VSCode window that opened, I see both the outline, and the `@` search works too.

### DIFF
--- a/components/clarity-lsp/src/common/requests/capabilities.rs
+++ b/components/clarity-lsp/src/common/requests/capabilities.rs
@@ -35,7 +35,7 @@ impl Default for InitializationOptions {
             completion: true,
             completion_smart_parenthesis_wrap: true,
             completion_include_native_placeholders: true,
-            document_symbols: false,
+            document_symbols: true,
             formatting: true,
             go_to_definition: true,
             hover: true,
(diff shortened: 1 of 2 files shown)


hugo-stacks:
Hey Hank,
It's pretty simple change, because it's was already implemented 😄 But it's disabled by default.

Can you try enabling it `"clarity-lsp.documentSymbols": true`, or in the UI:

Screenshot 2026-03-25 at 10 03 20

And see if it works well and answer your needs?
I'll take a look at the code and try remember why I marked it as beta in the first place. I tink it doesn't traverse the all AST, which is fine in clarity where most things that matter for this use case are pretty much top level.

If you can test for a few days, we may be able to take it out of beta

hstove-stacks:
My update here is that this works great, and I think we should change the default to "enabled" IMO. There does seem to be a small lag before symbols are available. After I open a contract in VSCode, it takes maybe 10-15 seconds before I can see them in the outline or via `@` searching.

hugo-stacks:
Hey @hstove-stacks 
I'm still seeing unhandled errors when I have symbols activated. I'll explore why, but for nw, we'll keep it off by default.
The lag is a concern as well 🤔 

I'm closing the PR, but we'll definitely keep the issue opened and try to get it to the finish line

PR #2356, opened 2026-04-08
feat(linter): Add `unnecessary_tuple` lint

jbencin-stacks:
### Description

Add `unnecessary_tuple` to linter (#2028)

This lint is part of `LintGroup::Perf`

This lint checks for single-item tuples in type signatures and values. Single item-tuples are less efficient than using the inner value directly, and are generally unnecessary. The only exception being as an argument to the `merge` built-in function, since that function only accepts tuple arguments

#### Breaking change?

No

### Example

This lint will generate a diagnostic on the following code:

```clarity
(define-map my-map { key: uint } { a: uint, b: bool })
```

You can supress the diagnostic on a specific line by placing the comment:

```clarity
;; #[allow(unnecessary_tuple)]
```

### Checklist


brady-stacks (approved):
lgtm

PR #2357, opened 2026-04-09
feat: support pox-5

brady-stacks:
- Add PoX-5 support to Clarinet (pox-5 boot contract, epoch 3.5 config, updated stacking/signing logic for new PoX 
- devnet snapshot uses a SnapshotLevel enum now (epoch 3.5 burn height 163)
- Update default epoch timing constants (3.1–3.5) so epochs transition at consecutive burn heights, and make epoch 3.5 a non-optional default in DevnetConfig
- Add btc_balance field to AccountConfig for seeding BTC in stacker wallets during devnet genesis
- Add PoX-5 stacking transaction generation using `make_pox_5_signer_key_signature` / `make_pox_5_signer_grant_signature`
- Handle snapshot boot gaps in the observer's `StacksBlockPool` by creating a new fork when a block is far ahead (>10 blocks) of all existing fork tips
  - Skip Docker image pull when the image already exists locally
  - Skip devnet compatibility checks when `--from-genesis` is passed


components/clarinet-files/src/network_manifest.rs
```diff
@@ -1013,7 +1021,7 @@ impl NetworkManifest {
                 epoch_3_2: devnet_config.epoch_3_2.unwrap_or(DEFAULT_EPOCH_3_2),
                 epoch_3_3: devnet_config.epoch_3_3.unwrap_or(DEFAULT_EPOCH_3_3),
                 epoch_3_4: devnet_config.epoch_3_4.unwrap_or(DEFAULT_EPOCH_3_4),
-                epoch_3_5: devnet_config.epoch_3_5,
+                epoch_3_5: Some(devnet_config.epoch_3_5.unwrap_or(DEFAULT_EPOCH_3_5)),
```
hugo-stacks: We don't want to have a default value here.

At some point, we'll be in a state where Clarinet will have to support both epoch 3.4 (by default) and preview for epoch 3.5.

If/when the user sets an epoch 3.5 height (along with a 3.5 image), Clarinet will handle it, but otherwise it'll still default to 3.4 until 3.5 activates on mainnet.

Until will eventually default to 3.5, like we recently did for 3.4 https://github.com/stx-labs/clarinet/pull/2342

PR #2359, opened 2026-04-13
chore: use rust-overlay for nix pkg pinning

brady-stacks:
CI was failing with

```
Rust in nixpkgs: 1.94.0
Latest stable:   1.94.1 (released 2026-03-26)
Stable version was released 18 days ago.
Error: nixpkgs Rust (1.94.0) is behind latest stable (1.94.1) which is > 7 days old.
```

This decouples our Rust version from the `nixpkgs` release cycle which is the error above (nixos-unstable lagged behind the latest stable)

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -20,11 +20,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1775423009,
-        "narHash": "sha256-vPKLpjhIVWdDrfiUM8atW6YkIggCEKdSAlJPzzhkQlw=",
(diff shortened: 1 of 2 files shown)


adriano-stacks (approved):
Thank you! I was hoping that 1.94.1 would make it into `nixpkgs` unstable quickly, but it didn't -- the PR for it was merged weeks ago, but they probably slowed down moving stuff to unstable due to the upcoming 26.05 release they are doing.

Using `rust-overlay` works great for us and gives us more flexibility. Thank you!

PR #2361, opened 2026-04-16
fix: pin new rand and rustls-webpki

brady-stacks:
`cargo audit` exposed 2 vulnerabilities in `rand` and `rustls-webpki` respectively. Repinned versions and ran audit again to ensure they were fixed 👍 

Also deals with updated clippy changes (`rustup update`)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2698,7 +2698,7 @@ dependencies = [
  "hyper-util",
  "log",
  "pin-project-lite",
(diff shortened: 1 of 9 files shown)


jbencin-stacks (changes requested):
Looks good but can you either split into 2 PRs, or change the description to mention the Clippy changes as well

jbencin-stacks (approved):

PR #2365, opened 2026-04-21
perf: Remove unnecessary call to `Session::new()`, saving ~70ms per LSP response

jbencin-stacks:
### Description

Remove an unnecessary call to `initiate_session_from_manifest()` (which makes an expensive call to `Session::new()`). This shaves about 70ms off of each LSP response


hugo-stacks (commented):
Awesome!
Have you manually tested (and tried to break it) with `env(simnet)` annotation?

jbencin-stacks:
@hugo-stacks I tried it and seems to work the same as the existing extension with `#[env(simnet)]`

hugo-stacks (approved):
lgtm!

PR #2369, opened 2026-04-24
fix: simnet annotation dimming

### LINKED ISSUE #2368: Badly annotated "simnet only" code is wrongly dimmed
Conditional code is wrongly dimmed in VSCode on invalid `env(simnet)` annotations.
Following up of this PR I believe: https://github.com/stx-labs/clarinet/pull/2306

This code is dimmed
```
;; [env(simnet)]
  (begin
    (asserts! (is-admin) ERR_NOT_ADMIN)
    (ok (var-set contract-owner new-admin))
  )
)
```
As well as 
```
;; [#env(simnet)]
  (begin
    (asserts! (is-admin) ERR_NOT_ADMIN)
    (ok (var-set contract-owner new-admin))
  )
)
```

While the correct annotation is `;; #[env(simnet)]`

Image

brady-stacks:
fix the annotation matching to strip the `#[` before matching to ensure we don't included the wrong annotations

- closes #2368 
 
 ## correct
Screenshot 2026-04-24 at 10 16 54 AM

## incorrect
Screenshot 2026-04-24 at 10 17 06 AM

### DIFF
--- a/components/clarity-repl/src/utils.rs
+++ b/components/clarity-repl/src/utils.rs
@@ -1,5 +1,3 @@
-use std::str::FromStr;
-
 use ::clarity::types::StacksEpochId;
 use ::clarity::vm::ast::parser;
 use ::clarity::vm::ast::stack_depth_checker::StackDepthLimits;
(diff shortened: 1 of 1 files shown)


jbencin-stacks:
Here's a test case that this doesn't handle:

```rust
    #[test]
    fn ignores_malformed_env_annotation_trailing_garbage_inside_brackets() {
        #[rustfmt::skip]
        let source = indoc!(r#"
            ;; #[env(simnet) this is not a real annotation]
                (begin
                    (ok (var-set contract-owner new-admin))
                )
            )
        "#);

        let spans = get_env_simnet_spans(source).unwrap();
        assert!(spans.is_empty());
    }
```

The issue here is that `AnnotationKind::from_str()` is mess. All of this parsing needs to be re-written. I can approve this as-is, since it fixes the immediate bug, but we will need a follow-up PR to remove the mess of regex parsing used here

jbencin-stacks (approved):

PR #2371, opened 2026-04-24
fix: remove annotation regex complexity

brady-stacks:
followup to #2369 

The regex for env matching was too error-prone. Replaced with simpler string operations

### DIFF
--- a/components/clarity-repl/src/analysis/annotation.rs
+++ b/components/clarity-repl/src/analysis/annotation.rs
@@ -15,10 +15,14 @@ pub enum AnnotationKind {
 impl std::str::FromStr for AnnotationKind {
     type Err = String;
 
-    /// Parse an annotation body, e.g. `env(simnet)` or `allow(unchecked_data)`.
-    /// The `#[…]` wrapper must already be stripped by the caller.
+    /// Parse an annotation, e.g. `#[env(simnet)]` or `#[allow(unchecked_data)]`.
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let s = s.trim();
+        let s = s
+            .trim()
+            .strip_prefix("#[")
+            .and_then(|s| s.strip_suffix(']'))
+            .ok_or_else(|| "malformed annotation: expected #[...]".to_string())?
+            .trim();
         let (name, value) = match s.find('(') {
             Some(open) => {
                 let name = &s[..open];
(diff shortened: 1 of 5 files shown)


components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -16,58 +15,63 @@ pub enum AnnotationKind {
...
-            } else {
-                (&captures[1], "")
-            };
-            match base {
-                "allow" => {
-                    let params: Vec<WarningKind> = value
+        let s = s.trim();
+        let (name, value) = match s.find('(') {
+            Some(open) => {
+                let name = &s[..open];
```
jbencin-stacks: If we have `#[env (simnet)]`, this will capture `name` as `env ` and fail, since it's not trimmed. Do we want that?

components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -16,58 +15,63 @@ pub enum AnnotationKind {
...
+                let rest = &s[open + 1..];
+                let value = rest
+                    .strip_suffix(')')
+                    .ok_or_else(|| "malformed annotation: missing closing ')'".to_string())?;
+                (name, value.trim())
+            }
+            None => (s, ""),
```
jbencin-stacks: Using `""` is usually not the correct approach in Rust, `value` should be an `Option`

components/clarity-repl/src/analysis/annotation.rs
jbencin-stacks: I can see us not allowing this syntax, but the general pattern here is to use `trim()` and ignore whitespace in most places, so IMO makes sense here too, but I could be convinced otherwise

components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -16,58 +15,63 @@ pub enum AnnotationKind {
 impl std::str::FromStr for AnnotationKind {
     type Err = String;
 
+    /// Parse an annotation body, e.g. `env(simnet)` or `allow(unchecked_data)`.
+    /// The `#[…]` wrapper must already be stripped by the caller.
```
jbencin-stacks: This seems weird that we expect the comment to be partially processed here. Does it make sense to move the square bracket matching into this function, rather than doing it at the callsite?

components/clarity-repl/src/analysis/annotation.rs
brady-stacks: yea, makes sense. It's a good catch. Trimmed now 👌

jbencin-stacks (approved):
This is pre-existing behavior, so doesn't have to be fixed here, but we allow unknown values in `allow` annotations:

```rust
#[test]
    assert!(
        result.is_err(),
    );
}
```

So we should add a comment or GitHub issue for this

brady-stacks:
https://github.com/stx-labs/clarinet/issues/2372

PR #2375, opened 2026-04-27
feat(static-cost): add warnings and resolving for local traits

brady-stacks:
Static cost analysis now warns when trait-based contract-call? cannot be resolved, and supports automatic trait resolution when concrete implementations are provided within the clarinet project.

Still requires a mechanism for specifying external trait resolution (to be handled in a separate branch)

### DIFF
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -1101,13 +1101,17 @@ async fn get_cost_analysis(
             call_stack: &mut call_stack,
         };
 
-        // Use static_cost which returns (StaticCost, Option<TraitCount>)
-        // TraitCount is HashMap<String, (u64, u64)>, so we can use it directly
-        static_cost(&mut env, &invoke_ctx, contract_id).map_err(|e| {
+        let result = static_cost(&mut env, &invoke_ctx, contract_id).map_err(|e| {
             clarity_repl::uprint!("[LSP] static_cost failed with error: {}", e);
             let error_msg = format!("Cost analysis failed for contract {}: {}", contract_id, e);
             VmExecutionError::Internal(VmInternalError::Expect(error_msg))
-        })
+        })?;
+
+        for warning in &result.warnings {
+            clarity_repl::uprint!("[LSP] Cost warning: {}", warning);
+        }
+
+        Ok(result.costs)
     });
 
     cost_result
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -1025,7 +1025,7 @@ pub async fn build_state(
             // Run static_cost_tree for this contract
             let Some(cost_result) = get_cost_analysis(session, contract_id, clarity_version).await
             else {
-                clarity_repl::uprint!("[LSP] Cost analysis failed for contract: {}", contract_id);
+                clarity_repl::uprint!("[LSP] Cost analysis failed for contract: {contract_id}");
                 continue;
             };
             clarity_repl::uprint!(
@@ -1074,9 +1074,7 @@ async fn get_cost_analysis(
     use clarity_static_cost::static_cost::static_cost;
 
     clarity_repl::uprint!(
-        "[LSP] get_cost_analysis called for contract: {} (clarity version: {:?})",
-        contract_id,
-        clarity_version
+        "[LSP] get_cost_analysis called for contract: {contract_id} (clarity version: {clarity_version:?})"
     );
 
     let tx_sender: clarity_types::types::PrincipalData = session.interpreter.get_tx_sender().into();
@@ -1086,7 +1084,7 @@ async fn get_cost_analysis(
         .interpreter
         .get_global_context(epoch, false)
         .map_err(|e| {
-            clarity_repl::uprint!("[LSP] Failed to get global context: {}", e);
+            clarity_repl::uprint!("[LSP] Failed to get global context: {e}");
             e
         })
         .ok()?;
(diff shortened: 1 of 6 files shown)


components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -585,33 +650,114 @@ fn extract_function_name(expr: &SymbolicExpression) -> Option<String> {
...
-        _ => first.match_field()?.contract_identifier.clone(),
+        Some(Value::Principal(PrincipalData::Contract(id))) => Some(id.clone()),
+        _ => first.match_field().map(|f| f.contract_identifier.clone()),
     };
 
-    let costs = static_cost(env, invoke_ctx, &contract_id).ok()?;
-    let (cost, _trait_count) = costs.get(function_name.as_str())?;
-    Some(cost.clone())
+    if let Some(contract_id) = contract_id {
+        let result = static_cost(env, ctx.invoke_ctx, &contract_id).ok()?;
```
jbencin-stacks: Does it make sense to do something other than silently discard error?

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -585,33 +650,114 @@ fn extract_function_name(expr: &SymbolicExpression) -> Option<String> {
...
+        )) => trait_id,
+        TypeSignature::TraitReferenceType(trait_id) => trait_id,
+        _ => return None,
+    };
+
+    let implementations = ctx.trait_implementations.get(&trait_id.name.to_string())?;
+
+    let mut envelope: Option<StaticCost> = None;
+    for impl_contract in implementations {
+        let result = static_cost(env, ctx.invoke_ctx, impl_contract).ok()?;
```
jbencin-stacks: Does it make sense to do something other than silently discard error?

jbencin-stacks: Also, if we get a single error here, we get an early return of `None` for all `implementations`. Same problem with early return on next line. Probably want to run the whole loop and filter out `None` values.

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -148,6 +149,16 @@ pub struct AnalysisContext<'a> {
     pub clarity_version: &'a ClarityVersion,
     pub epoch: StacksEpochId,
     pub invoke_ctx: &'a InvocationContext<'a>,
+    /// Mapping from unqualified trait names to concrete contract implementations.
+    /// Used to resolve trait-based `contract-call?` targets during static analysis.
+    /// Note: keys are simple trait names (e.g. `"sip-010-trait"`), not fully-qualified
+    /// identifiers. This is sufficient for in-project analysis where trait names are
+    /// unique, but could collide if multiple traits share the same name.
+    pub trait_implementations: &'a HashMap<String, Vec<QualifiedContractIdentifier>>,
```
jbencin-stacks: Why not key on the fully-qualified `TraitIdentifier` which prevents collisions entirely?

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -396,11 +415,37 @@ pub fn static_cost_from_ast_with_source(
...
+        clarity_version,
+        epoch,
+        contract_source,
+        &HashMap::new(),
+        env,
+        invoke_ctx,
+    )
+}
+
+pub fn static_cost_from_ast_with_options(
```
jbencin-stacks: Should we be more specific in the naming here:


```suggestion
pub fn static_cost_from_ast_with_trait_impls(
```

brady-stacks: yeah, the naming is horrible. Maybe a config struct could get rid of all these variants. I haven't messed around with that much. Will give it a shot

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -585,33 +636,153 @@ fn extract_function_name(expr: &SymbolicExpression) -> Option<String> {
...
+    let mut envelope: Option<StaticCost> = None;
+    for impl_contract in implementations {
+        let result = match static_cost(env, ctx.invoke_ctx, impl_contract) {
+            Ok(r) => r,
+            Err(_) => continue,
+        };
+        let (cost, _) = match result.costs.get(function_name.as_str()) {
+            Some(entry) => entry,
+            None => continue,
+        };
```
jbencin-stacks: Use `let`/`else`:

```suggestion
        let Ok(result) = static_cost(env, ctx.invoke_ctx, impl_contract) else {
            continue;
        };
        let Some((cost, _)) = result.costs.get(function_name.as_str()) else {
            continue;
        };
```

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -235,14 +255,22 @@ fn make_ast(
...
+/// warnings about incomplete analysis (e.g. unresolved trait calls).
+#[derive(Debug, Clone)]
+pub struct StaticCostResult {
+    pub costs: HashMap<String, (StaticCost, Option<TraitCount>)>,
+    pub warnings: Vec<CostWarning>,
+}
+
 /// Static execution cost for functions within Environment.
 /// Returns the static cost for each function in the contract.
 /// {some-function-name: (StaticCost, Some({some-function-name: (1,1)}))}
```
jbencin-stacks: The doc comment here no longer matches the actual return type

components/clarity-lsp/src/common/state.rs
```diff
@@ -1083,36 +1093,47 @@ async fn get_cost_analysis(
...
-        Ok(result.costs)
-    });
+            let warnings = result
+                .warnings
+                .iter()
+                .map(|w| {
+                    clarity_repl::uprint!("[LSP] Cost warning: {}", w);
+                    ClarityDiagnostic {
+                        level: ClarityLevel::Warning,
+                        message: format!("{w}"),
```
jbencin-stacks: Minor, but have Claude use `.to_string()` here and inline variables in the format strings everywhere else in this PR

jbencin-stacks (approved):

PR #2376, opened 2026-04-28
fix: reject unknown warning kinds in allow annotations

### LINKED ISSUE #2372: don't allow unknown values in allow annotations
```rust
#[test]
    assert!(
        result.is_err(),
    );
}
```

context: https://github.com/stx-labs/clarinet/pull/2371/#pullrequestreview-4182160708

truffle-dev:
The `allow` arm in `AnnotationKind::from_str` used `filter_map(.. .ok())`, which silently dropped unknown warning kinds rather than reporting them. `#[allow(unused_const, not_a_real_warning)]` would parse as an `Allow` with just `unused_const\`, hiding the typo from the user.

Switched to a fail-fast `map(|s| s.parse().map_err(...)).collect::<Result<_, _>>()?` shape, mirroring the existing `filter` arm. The error message surfaces the unknown name so the diagnostic emitted by `Interpreter::collect_annotations` reads `unknown warning kind 'not_a_real_warning' in 'allow' annotation`.

Test from the issue body added verbatim. All 12 `analysis::annotation` tests pass and all 6 `repl::interpreter::tests::annotation_*` diagnostic tests still pass (`annotation_allow_missing_value`, etc.) — the empty-allow path still hits the existing `if params.is_empty()` branch.

Followup to the carve-out in #2371's review.

closes #2372

### DIFF
--- a/components/clarity-repl/src/analysis/annotation.rs
+++ b/components/clarity-repl/src/analysis/annotation.rs
@@ -40,9 +40,14 @@ impl std::str::FromStr for AnnotationKind {
                 let params: Vec<WarningKind> = value
                     .unwrap_or("")
                     .split(',')
+                    .map(str::trim)
                     .filter(|s| !s.is_empty())
-                    .filter_map(|s| s.trim().parse().ok())
-                    .collect();
+                    .map(|s| {
+                        s.parse().map_err(|_| {
+                            format!("unknown warning kind '{s}' in 'allow' annotation")
+                        })
+                    })
+                    .collect::<Result<_, _>>()?;
                 if params.is_empty() {
                     Err("missing value for 'allow' annotation".to_string())
                 } else {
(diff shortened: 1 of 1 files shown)


components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -40,9 +40,14 @@ impl std::str::FromStr for AnnotationKind {
                 let params: Vec<WarningKind> = value
                     .unwrap_or("")
                     .split(',')
+                    .map(str::trim)
```
brady-stacks: This is redundant, right?

components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -180,6 +185,20 @@ mod tests {
         };
     }
 
+    #[test]
+    fn parse_allow_rejects_unknown_warning_kind() {
+        // If the user includes an unknown warning kind alongside a valid one,
+        // parsing should fail rather than silently dropping the unknown name and
+        // accepting only the recognized parts. Today the `filter_map(.. .ok())`
```
brady-stacks: No need to reference the previous bug. You can remove the "Today the..." sentence

truffle-dev:
1. Yes. Collapsed the standalone `.map(str::trim)` into the parse closure so the chain matches the sibling `filter` arm and the original style from #2371. a29b9978
2. Removed. a29b9978

components/clarity-repl/src/analysis/annotation.rs
```diff
@@ -41,8 +41,13 @@ impl std::str::FromStr for AnnotationKind {
                     .unwrap_or("")
                     .split(',')
                     .filter(|s| !s.is_empty())
-                    .filter_map(|s| s.trim().parse().ok())
-                    .collect();
+                    .map(|s| {
+                        let s = s.trim();
+                        s.parse().map_err(|_| {
```
brady-stacks: why not s.trim().parse() ?

brady-stacks:
In order to merge this we need to have verified commits. Here's [a reference](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) from github

truffle-dev:
1. Chained `s.trim().parse()` and dropped the rebinding. f31c84cf
2. Set up SSH commit signing locally and rebased the branch with `-S`. All three commits now show `Verified` (52fc5026, f4c9be7e, f31c84cf).

brady-stacks (approved):

PR #2384, opened 2026-05-06
feat(static-cost): add external trait resolution

brady-stacks:
handles traits defined in external contracts

### DIFF
--- a/.github/workflows/ci-vscode.yaml
+++ b/.github/workflows/ci-vscode.yaml
@@ -78,7 +78,7 @@ jobs:
           WASM_PACK_VERSION: "0.14.0"
           WASM_PACK_SHA256: "278a8d668085821f4d1a637bd864f1713f872b0ae3a118c77562a308c0abfe8d"
         run: |
-          curl -fsSL "https://github.com/drager/wasm-pack/releases/download/v${WASM_PACK_VERSION}/wasm-pack-v${WASM_PACK_VERSION}-x86_64-unknown-linux-musl.tar.gz" -o wasm-pack.tar.gz
+          curl -fsSL "https://github.com/wasm-bindgen/wasm-pack/releases/download/v${WASM_PACK_VERSION}/wasm-pack-v${WASM_PACK_VERSION}-x86_64-unknown-linux-musl.tar.gz" -o wasm-pack.tar.gz
           echo "${WASM_PACK_SHA256}  wasm-pack.tar.gz" | sha256sum -c -
           tar xzf wasm-pack.tar.gz
           mv "wasm-pack-v${WASM_PACK_VERSION}-x86_64-unknown-linux-musl/wasm-pack" /usr/local/bin/
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -532,13 +532,16 @@ impl EditorState {
 
         // Create a CodeLens for each function definition
         for (function_name, saved_range) in &contract_state.definitions {
-            // If the function's range in the live source differs from the saved range,
-            // the function (or something above it) was edited — skip the stale lens.
-            if let Some(ref current) = current_ranges {
-                if current.get(function_name) != Some(saved_range) {
-                    continue;
-                }
-            }
+            // Use the live range so the lens tracks the function's current
+            // position as the user edits.  If we have live ranges but the
+            // function is missing (deleted/renamed), skip it.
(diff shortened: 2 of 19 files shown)


jbencin-stacks (commented):
There's a potential stack overflow if you have two contracts that can call into each other using traits. See brady-stacks/clarinet#5

components/clarity-lsp/src/common/state.rs
```diff
@@ -1066,11 +1078,51 @@ struct CostAnalysisResult {
     warnings: Vec<ClarityDiagnostic>,
 }
 
+/// Scan all contract ASTs for `impl-trait` declarations and build a mapping
+/// from trait identifiers to the contracts that implement them.
+fn collect_trait_implementations(
+    asts: &BTreeMap<QualifiedContractIdentifier, ContractAST>,
+) -> HashMap<clarity_types::types::TraitIdentifier, Vec<QualifiedContractIdentifier>> {
```
jbencin-stacks: This `HashMap` type appears a lot. How about doing:

```rust
pub type TraitImplementations = HashMap<clarity_types::types::TraitIdentifier, Vec<QualifiedContractIdentifier>>;
```

.github/workflows/ci-vscode.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
           WASM_PACK_VERSION: "0.14.0"
           WASM_PACK_SHA256: "278a8d668085821f4d1a637bd864f1713f872b0ae3a118c77562a308c0abfe8d"
         run: |
-          curl -fsSL "https://github.com/drager/wasm-pack/releases/download/v${WASM_PACK_VERSION}/wasm-pack-v${WASM_PACK_VERSION}-x86_64-unknown-linux-musl.tar.gz" -o wasm-pack.tar.gz
+          curl -fsSL "https://github.com/wasm-bindgen/wasm-pack/releases/download/v${WASM_PACK_VERSION}/wasm-pack-v${WASM_PACK_VERSION}-x86_64-unknown-linux-musl.tar.gz" -o wasm-pack.tar.gz
```
jbencin-stacks, line 81: This seems fine but put it in the description or make a separate PR for it

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -668,7 +718,16 @@ fn get_contract_call_target_cost(
     };
 
     if let Some(contract_id) = contract_id {
-        match static_cost(env, ctx.invoke_ctx, &contract_id) {
+        if !ctx.visited.borrow_mut().insert(contract_id.clone()) {
+            return None;
```
jbencin-stacks: I think the `HashSet` here is insufficient, because you always return `None` when found. That's good to prevent a cycle when you're analyzing a contract, but if you're not in a cycle, and need cost info for a contract again later, than `None` is not correct. What you'll need here is:

```rust
HashMap<QualifiedContractIdentifier, Option<StaticCostResult>>
```

Where the contract is inserted with a value of `None` when you start analysis to prevent a cycle, but is given a cost value when you finish, so it is available later if you actually need it again

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -416,6 +442,27 @@ pub fn static_cost_from_ast(
     config: Option<&StaticCostConfig>,
     env: &mut ExecutionState,
     invoke_ctx: &InvocationContext,
+) -> Result<StaticCostResult, StaticCostError> {
+    let visited = RefCell::new(HashSet::new());
```
jbencin-stacks: You need to add:

```rust
visited.borrow_mut().insert(contract_ast.contract_identifier.clone());
```

so it matches the logic in `static_cost()`

components/clarity-static-cost/src/static_cost/cost_analysis.rs
```diff
@@ -293,7 +294,7 @@ pub fn static_cost(
         epoch,
         Some(&StaticCostConfig {
             contract_size: contract_source.len() as u64,
-            trait_implementations: HashMap::new(),
+            trait_implementations: trait_implementations.clone(),
         }),
```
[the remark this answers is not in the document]
jbencin-stacks: Claude is flagging this too. Can it be changed to pass a reference or use a `RefCell`?

jbencin-stacks (approved):

PR #2385, opened 2026-05-06
feat: use codelens live-range to avoid flashing

brady-stacks:
When the user edited a file, `get_code_lenses` compared the live function ranges against the saved (last-compiled) ranges. If they differed (i.e., the user typed anything that shifted lines), all affected lenses were skipped entirely.. which returned empty results to VSCode, which cleared them from the display.

Fix: Instead of skipping lenses when positions change, use the live range from the current editor content to position the lens. The cost data is still from the last save (stale), but it stays visible at the correct line. When the user saves, fresh cost analysis runs and the data updates.

https://github.com/user-attachments/assets/29827c45-0410-43db-b8ba-a79cc6053311

### DIFF
--- a/components/clarity-lsp/src/common/state.rs
+++ b/components/clarity-lsp/src/common/state.rs
@@ -532,13 +532,16 @@ impl EditorState {
 
         // Create a CodeLens for each function definition
         for (function_name, saved_range) in &contract_state.definitions {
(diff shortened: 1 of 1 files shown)


jbencin-stacks (approved):
Looks good, re-run the CI to make sure it passes

PR #2389, opened 2026-05-12
chore: use epoch4 built-ins via pox-wf-integration

brady-stacks:
Prepare for pox-5 with a stacks-core update from [pox-wf-integration](https://github.com/stacks-network/stacks-core/tree/pox-wf-integration)


hugo-stacks (approved):
lgtm

PR #2394, opened 2026-05-15
chore: pnpm v11

hugo-stacks:
### Description

Upgrade to pnpm v11, it brings better security defaults (blockexoticsubdeps now defaults to true, it doesn't allow build scripts, unless the whitelisted ones, like esbuild in our case).

Also bumped `minimumReleaseAge` from 1 day to 5.
it's trade off, less likely to pull malicious dependencies, but slower to pull patches. 3 or 5 days sounds like a solid middle ground. And we don't upgrade our dependencies that often on this repo.

Additionally, pnpm v11 changed `verifyDepsBeforeRun` to default to install, so `pnpm run build:sdk-wasm`now auto-runs `pnpm install`, which fires the prepare script on @stacks/clarinet-sdk-browser before the wasm packages exist.
In the `build_wasm` CI job, run `pnpm install --frozen-lockfile --ignore-scripts` explicitly before building wasm, so the auto-check is satisfied without triggering prepare.

### DIFF
--- a/.github/workflows/ci-sdk.yaml
+++ b/.github/workflows/ci-sdk.yaml
@@ -72,6 +72,9 @@ jobs:
             cargo install wasm-pack
           fi
 
+      - name: Install dependencies (skip prepare scripts)
+        run: pnpm install --frozen-lockfile --ignore-scripts
+
       - name: Build Wasm packages
         run: pnpm run build:sdk-wasm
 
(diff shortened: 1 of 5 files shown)


brady-stacks (dismissed):

brady-stacks (dismissed):

brady-stacks (approved):
👍

PR #2398, opened 2026-05-19
chore(vscode-ext): upgrade pnpm and dependencies

hugo-stacks:
### Description
upgrade pnpm and rspack in clarinet vscode

- address breaking changes in configs
- set `minimumReleaseAge` (it expects minutes, not seconds...)

### DIFF
--- a/components/clarity-vscode/package.json
+++ b/components/clarity-vscode/package.json
@@ -47,7 +47,7 @@
     "theme": "dark"
   },
   "qna": "https://github.com/stx-labs/clarinet/issues",
-  "packageManager": "pnpm@10.33.0",
+  "packageManager": "pnpm@11.1.3",
   "engines": {
     "node": ">=24",
     "vscode": "^1.64.0"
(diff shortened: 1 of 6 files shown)


components/clarity-vscode/package.json
```diff
@@ -245,22 +245,22 @@
...
     "@typescript-eslint/parser": "8.58.0",
     "@vscode/test-web": "0.0.80",
-    "@vscode/vsce": "3.7.1",
+    "@vscode/vsce": "3.9.1",
     "@wasm-tool/wasm-pack-plugin": "1.7.0",
     "concurrently": "9.2.1",
     "eslint": "10.1.0",
-    "ovsx": "0.10.10",
+    "ovsx": "0.10.12",
     "playwright": "1.58.2",
```
jbencin-stacks, line 260: In the lockfile we updated to `@playwright/browser-chromium@1.60.0`, so we should update here also to match

hugo-stacks: Hmm good catch, I've been fighting this, let me check.

At the end of the day, if the lockfile isn't up to date `pnpm install --frozen-lockfile` should break and warn us in CI

components/clarity-vscode/package.json
```diff
@@ -47,7 +47,7 @@
     "theme": "dark"
   },
   "qna": "https://github.com/stx-labs/clarinet/issues",
-  "packageManager": "pnpm@10.33.0",
+  "packageManager": "pnpm@11.1.3",
```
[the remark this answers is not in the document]
hugo-stacks: Good catch, but instead let's use 11.1.3, committed

components/clarity-vscode/package.json
hugo-stacks: So, vscode-test-web is iporting both
```
> pnpm why playwright-core@1.58.2
playwright-core@1.58.2
└─┬ playwright@1.58.2
  └─┬ @vscode/test-web@0.0.80
    ├── @stacks/clarity-vs-code-web-client@0.0.1 (dependencies)
    └── clarity-stacks@3.17.0 (devDependencies)


> pnpm why playwright-core@1.60.0
playwright-core@1.60.0
└─┬ @playwright/browser-chromium@1.60.0
  └─┬ @vscode/test-web@0.0.80
    ├── @stacks/clarity-vs-code-web-client@0.0.1 (dependencies)
    └── clarity-stacks@3.17.0 (devDependencies)
```

But it does expect us to use @1.58.2  (or it [fails](https://github.com/stx-labs/clarinet/actions/runs/26100725034/job/76751078369))

It's okay, just dev dependencies, it doesn't impact the size of the build

jbencin-stacks: If it builds and tests pass I'm okay with it, but we should look into this and possibly do a follow-up PR

jbencin-stacks (approved):

PR #2401, opened 2026-05-19
test: split encryption tests

hugo-stacks:
### Description

This test can sometime time out.
https://github.com/stx-labs/clarinet/actions/runs/26110965106/job/76787785016?pr=2399

I would also be ok to just remove the "high" test, I'm not sure it has value


hugo-stacks:
Ready for re-review @brady-stacks

brady-stacks (approved):

PR #2408, opened 2026-05-22
feat: replace curl shellout with worker-thread sync HTTP in Node WASM

hugo-stacks:
### Description

The Node build of clarity-repl used to shell out to the system curl binary every time it needed to fetch data from the Stacks API (because WASM can't await Promises like normal JS code). That meant users needed curl installed on their machine, and Windows users sometimes hit weird issues with quoting and PATH.

This PR replaces the curl call with a small JS helper that uses Node's built-in fetch. To keep the API synchronous (which is what the WASM code needs), the helper runs fetch on a background worker thread and signals back to the main thread through shared memory.

With this change, it was easier to move the HTTP retry logic (3 attempts, rate-limit handling, 5xx backoff) into a shared file **so the native and Node code paths can't drift apart again**. The retry behavior used to only exist on the native side.

## What changed
- No more dependency on system curl for the Node SDK.
- Retry behavior is now identical between the CLI and the SDK.
- Bumped the SDK's minimum Node version to 20 (needed for built-in fetch).
- No new npm or Rust dependencies.

--- 
_tl;dr_
#### Context
- in MXS (mainnet execution simulation) we fetch RPC endpoints to get actual mainnet data
- the datastore is fully synchronous
- we can't perform sync HTTP calls in Node.js / WASM
- the hack was to shell out to curl to make it synchronous
Fix
- this PR spawns a worker thread that handles the http call (asynchronously, with fetch), and the main thread uses Atomics.wait on a SharedArrayBuffer to block until the worker is done

## Note on the PR

The code is quite low level (for JS I mean...) and heavily AI written.
The SharedArrayBuffer (SAB) data structure can seem a bit scary, but this is standard practices when working with child processes in Node (although relative new).

I think this is overall very solid, a net improvement compared to the previous previous on several aspects
- less hack
- more robust
- faster

--- 

### Benchmarks

before (main)
```
branch=main commit=7a1493b5 node=v24.16.0
api=https://api.testnet.hiro.so iterations=5 atBlockCalls=10 apiKey=yes
warmup… done
iteration 1/5… 7.54s
iteration 2/5… 7.20s
iteration 3/5… 6.93s
iteration 4/5… 6.99s
iteration 5/5… 7.93s

results (main @ 7a1493b5):
  initEmptySession       n=  5  min= 402.1ms  median= 475.7ms  mean= 516.4ms  p95= 747.3ms  max= 747.3ms  total=2.58s
  cold callReadOnlyFn    n=  5  min= 986.8ms  median=   1.08s  mean=   1.07s  p95=   1.16s  max=   1.16s  total=5.37s
  at-block call          n= 50  min= 499.9ms  median= 556.4ms  mean= 572.5ms  p95= 715.2ms  max=   1.04s  total=28.63s
  wall time: 44.61s  errors: 0
  ```
  
  after (this branch)
  ```
  branch=feat/replace-curl-shellout-with-worker-thread commit=b0399e26 node=v24.16.0
api=https://api.testnet.hiro.so iterations=5 atBlockCalls=10 apiKey=yes
warmup… done
iteration 1/5… 4.60s
iteration 2/5… 4.87s
iteration 3/5… 4.70s
iteration 4/5… 4.90s
iteration 5/5… 4.66s

results (feat/replace-curl-shellout-with-worker-thread @ b0399e26):
  initEmptySession       n=  5  min= 251.3ms  median= 257.2ms  mean= 286.4ms  p95= 401.3ms  max= 401.3ms  total=1.43s
  cold callReadOnlyFn    n=  5  min= 709.4ms  median= 756.2ms  mean= 787.9ms  p95= 940.8ms  max= 940.8ms  total=3.94s
  at-block call          n= 50  min= 342.1ms  median= 356.5ms  mean= 367.1ms  p95= 447.4ms  max= 567.9ms  total=18.35s
  wall time: 29.02s  errors: 0
  ```
  
  Summary
  
operation (median) | main (curl) | worker thread | delta
-- | -- | -- | --
initEmptySession | 476ms | 257ms | −46%
cold callReadOnlyFn | 1.08s | 756ms | −30%
at-block call | 556ms | 357ms | −36%
wall time | 44.6s | 29.0s | −35%

Node: `at-block` isn't relevant anymore, but was useful to bench un-cached data

### CODE REVIEW

File: components/clarinet-sdk/node/tests/sync-http.test.ts [hunk 1]
```diff
@@ -0,0 +1,257 @@
...
+let mockChild: ChildProcess;
+let mockBaseUrl: string;
+
+function spawnMockServer(): Promise<{ url: string; child: ChildProcess }> {
+  const child = spawn(
+    process.execPath,
+    [
+      "-e",
+      `
+      const http = require('node:http');
```
brady-stacks: wwould it be weird to put this in fixtures or something? Then read out the file into this argument?

brady-stacks: (non-blocking)

components/clarity-repl/src/repl/remote_data/http_request/retry.rs
```diff
@@ -0,0 +1,277 @@
...
+    T: DeserializeOwned,
+    F: Fn() -> Result<Response, String>,
+    S: FnMut(u32),
+{
+    let mut attempts: u32 = 0;
+    loop {
+        let response = transport()?;
+        let status = response.status;
+
+        if (200..300).contains(&status) {
```
brady-stacks, line 80: this should be 200..399 ?

components/clarity-repl/src/repl/remote_data/http_request/native.rs
```diff
@@ -1,88 +1,62 @@
...
+            v.to_str()
+                .ok()
+                .map(|s| (k.as_str().to_string(), s.to_string()))
+        })
+        .collect()
 }
 
 pub fn http_request<T: DeserializeOwned>(url: &str) -> Result<T, String> {
-    let client = Client::new();
+    let client = Client::builder()
```
brady-stacks:
```rust
   use std::sync::LazyLock;
   static CLIENT: LazyLock<Client> = LazyLock::new(|| {
       Client::builder().timeout(REQUEST_TIMEOUT).build().expect("build client")
   });
```
Maybe we don't need to build a client for each request?

components/clarity-repl/src/repl/remote_data/http_request/retry.rs
hugo-stacks: I think 2xx is good.
At the point, wether the request comme from node fetch or reqwest in native, both should have followed the redirects, the 3xx shouldn't make it there. 
Also, if any 3xx hits this code path, it wouldn't be json so we'd have parsing errors, and it doesn't make much sense to retry on a 3xx.

components/clarinet-sdk/node/tests/sync-http.test.ts
hugo-stacks: good call

components/clarity-repl/src/repl/remote_data/http_request/native.rs
hugo-stacks: yes, good call 💯

brady-stacks (approved):
Huge! Very nice

PR #2409, opened 2026-05-26
docs: add on-chain game scoring example to README

Dark-Brain07:
## Summary
Adds a practical Clarity smart contract example showing how to build game score submission with a fee using Clarinet.

This pattern is battle-tested in production — it's the same approach used by [StacksHurry](https://stackshurry.vercel.app), an active on-chain rocket shooter with 7 Clarity contracts deployed on Stacks mainnet.

Adding real-world usage examples helps developers understand what's possible with Clarinet beyond simple token contracts.

## Changes
- Added `Example: On-chain game scoring contract` subsection to README.md
- Includes a minimal Clarity contract with stx-transfer?, map-set, and print
- Includes the 3 key Clarinet commands (check, test, console)

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,44 @@ clarinet console
 (contract-call? .counter get-count tx-sender)
 ```
 
+### Example: On-chain game scoring contract
+
+Clarinet makes it easy to build and test game mechanics using Clarity.
+Here's a minimal example of a score-submission contract with a fee:
+
+```clarity
+;; game-score.clar
+(define-constant CONTRACT-OWNER tx-sender)
+(define-constant SUBMIT-FEE u5000)
+(define-constant ERR-INVALID-SCORE (err u100))
+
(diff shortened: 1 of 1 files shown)


hugo-stacks:
Hey @Dark-Brain07,

We are keeping this readme as lean as possible. There is already one simple Clarity example in the readme that is enough to demonstrate what Clarinet does. 
If you think that there is one particular smart contract pattern that should be explained somewhere, I think the [docs repo](https://github.com/stacks-network/docs) might be a more suitable place.
I would recommend starting by opening an issue explaining the gap you are trying to fill. 

--- 

Regarding the code snippet itself, letting a user submit an arbitrary number for a score is not a viable on-chain scoring system, that would be considered as an anti-pattern. Happy to chat more about it (on Discord for instance).

I'm closing this PR, please open an issue or reach out on Discord to discuss

PR #2424, opened 2026-06-08
fix: warn if remote_data enabled in optional epoch

brady-stacks:
Display a warning if a user attempts to run a project with remote_data enabled AND has enabled an optional (future) epoch


brady-stacks:
This should be handled by #2431 and pessimistically by #2429

PR #2425, opened 2026-06-09
fix: new epoch and clarity types

hugo-stacks:
### Description

sdk:
- use @stacks/transaction ClarityVersion enum, one less place to maintain on new ClarityVersions
- Fix tsconfig and node types (similar to #2423, remaining work from the pnpm migration)

sdk-wasm:
- Add epoch 4.0 to the exported `EpochString` (used in `simnet.setEpoch`)


brady-stacks (approved):
👍

PR #2426, opened 2026-06-09
feat(linter): flatten nested variadic functions

brady-stacks:
- Update stacks-core 6dd4b092, which introduces variadic concat in Clarity 6
  - Add `flatten_nested_variadic` lint that suggests flattening nested calls to variadic functions 
    - `(+ a (+ b c))` -> `(+ a b c)`
    - `(concat a (concat b c))` -> `(concat a b c))`
  - The concat suggestion only fires in Clarity 6+, arithmetic functions like +, -, *, / for all versions
  - Add formatter unit tests for nested and variadic concat

Screenshot 2026-06-11 at 10 01 00 AM


components/clarity-repl/src/analysis/lints/flatten_nested_variadic.rs
```diff
@@ -0,0 +1,340 @@
...
+        let mut session = Session::new_without_boot_contracts(settings);
+        session.update_epoch(StacksEpochId::Epoch40);
+        session
+            .formatted_interpretation(snippet, Some("checker".to_string()), false, None)
+            .expect("Invalid code snippet")
+    }
+
+    #[test]
+    fn warn_nested_add() {
+        let snippet = indoc!(
```
jbencin-stacks: Add `#[rustfmt(skip)]` or use the curly brackets and format the `indoc!` strings like they are in the other lints

components/clarity-repl/src/analysis/lints/flatten_nested_variadic.rs
```diff
@@ -0,0 +1,340 @@
...
+        expr: &'a SymbolicExpression,
+        operands: &'a [SymbolicExpression],
+    ) -> bool {
+        if self.clarity_version >= ClarityVersion::Clarity6 {
+            self.check_operands(expr, "concat", operands);
+        }
+        true
+    }
+}
+
```
jbencin-stacks, line 171: Also should implement `visit_lazy_logical()` to check `and` and `or` operations

components/clarity-repl/src/analysis/lints/flatten_nested_variadic.rs
```diff
@@ -0,0 +1,340 @@
...
+    fn visit_arithmetic(
+        &mut self,
+        expr: &'a SymbolicExpression,
+        func: NativeFunctions,
+        operands: &'a [SymbolicExpression],
+    ) -> bool {
+        // Only lint associative operators where flattening preserves semantics.
+        // e.g. (- a (- b c)) != (- a b c), so we skip Subtract, Divide, etc.
+        if matches!(func, NativeFunctions::Add | NativeFunctions::Multiply) {
```
jbencin-stacks, line 114: Flattening can actually, in some edge cases, lead to overflow and abort with an error, where nesting would not. Examples:

- **`+` on `int`** — `a = -10`, `b = i128::MAX`, `c = 5`:
  - `(+ a (+ b c))` → inner `MAX + 5` overflows → **runtime error**
  - `(+ a b c)` → `(-10 + MAX) = MAX-10`, then `+ 5 = MAX-5` → **succeeds**
- **`*` on `uint`** — `a = u128::MAX`, `b = u128::MAX`, `c = u0`:
  - `(* a (* b c))` → inner `u128::MAX * 0 = 0`, then `u128::MAX * 0 = 0` → **succeeds**
  - `(* a b c)` → `(u128::MAX * u128::MAX)` overflows → **runtime error**

We can't even check for potential overflow in Clarinet if the values are not constants, so perhaps the diagnostic message for arithmetic operations should carry a note about potential overflow

components/clarity-repl/src/analysis/lints/flatten_nested_variadic.rs
```diff
@@ -0,0 +1,340 @@
...
+                ;; #[allow(flatten_nested_variadic)]
+                (+ u1 (+ u2 u3))
+            )
+        "
+        )
+        .to_string();
+
+        let (_, result) = run_snippet(snippet);
+        assert_eq!(result.lint_diagnostics.len(), 0);
+    }
```
jbencin-stacks, line 386: I don't see any tests for deep nesting. Add some tests for 3 or 4 layers to see what happens

components/clarity-repl/src/analysis/lints/flatten_nested_variadic.rs
jbencin-stacks: Also can flatten:
- Bitwise operators (`bit-and`, `bit-or`, and `bit-xor`)
- `begin` blocks

components/clarity-repl/src/analysis/lints/flatten_nested_variadic.rs
jbencin-stacks: Also add a test for something like this, where we are already using variadic structure but can flatten it more:

```clarity
(+ u1 u2 u3 u4 (+ u5 u6 u7))
```

jbencin-stacks (approved):

PR #2428, opened 2026-06-10
feat: pox-5 devnet capabilities

### LINKED ISSUE #2438: `override_boot_contracts_source` causes duplicate items in `getContractInterfaces()`
When using `override_boot_contracts_source` in Clarinet.toml, it causes an issue where some boot contracts (requirements?) are added twice. Specifically, this happens when you have sBTC as a requirement in your Clarinet.toml file.

The result is that `simnet.getContractsInterfaces()` includes two contracts for `sbtc-token` (and `sbtc-registry`) - one deployed by `SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4` (which is "correct"), and one by the simnet deployer.

This ended up causing a bug in Clarigen, which is how I noticed this.

You can minimally reproduce this in the `pox-wf-integration` branch of Stacks Core by adding this to `pox-5.test.ts`:

```
test.only('demo of clarinet issue', () => {
  const interfaces = simnet.getContractsInterfaces();
  interfaces.forEach((_value, key) => {
    if (key.includes('sbtc-')) console.log(key);
  });
});
```

Then, run `npx vitest run pox-5.test.ts`. You'll see that it prints out:

```
ST1F7QA2MDF17S807EPA36TSS8AMEFY4KA9TVGWXT.sbtc-registry
SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-withdrawal
SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-registry
SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token
ST1F7QA2MDF17S807EPA36TSS8AMEFY4KA9TVGWXT.sbtc-token
SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-deposit
```

As you can see, `sbtc-withdrawal` and `sbtc-deposit` don't have this issue - maybe because they're explicitly set as requirements?

brady-stacks:
Devnet support for epoch 4.0 / pox-5 stacking contracts.

- closes #2438  (simnet duplicate address issue)
- Generate pox-5 node config (pox_5_sbtc_contract, pox_5_sbtc_registry_contract, pox_5_bond_admin) in the stacks-node toml _only_ when epoch 4.0 is configured. 
- Auto-deploy sbtc boot contracts: Automatically add sbtc-registry and sbtc-token as requirement-publish transactions in the epoch 3.0 deployment batch, since pox-5 depends on them. 
- Patch the sbtc-registry contract's current-aggregate-pubkey with the first signer's public key so the miner doesn't stall during the prepare phase.
- pox-5 stacking - deploy signer-manager contract, call register-self, call pox-5.stake 
- Run pox-5 initial staking threads detached to avoid deadlocking the coordinator.
- PoX contract transition detection: Detect when the active PoX contract changes (pox-4 → pox-5) and re-submit stacking orders so the chain doesn't stall at the next reward cycle
- Observer compatibility: Handle nullable fields (`parent_microblock`, `parent_microblock_sequence`, `pox_ustx_threshold`) in the new_block event payload


components/stacks-network/src/orchestrator.rs
```diff
@@ -1503,18 +1526,83 @@ impl DevnetOrchestrator {
...
             .ok_or("unable to boot container")?;
         let docker = self.docker_client.as_ref().ok_or(DOCKER_ERR_MSG)?;
+        let devnet_config = self.get_devnet_config()?;
 
         docker
             .start_container::<String>(container, None)
             .await
             .map_err(|e| formatted_docker_error("unable to start postgres container", e))?;
 
+        // The stacks-blockchain-api pox5 image requires the
```
brady-stacks: stacks api image issue or something else, I'm not sure. This should be deleted before merge

components/clarinet-deployments/src/lib.rs
```diff
@@ -543,6 +544,16 @@ pub async fn generate_default_deployment_with_cache(
...
+
+        // Add sbtc contract ASTs so they're available without API fetch.
+        // pox-5 (deployed at epoch 4.0) depends on sbtc-token.
+        // These are treated as boot contracts (added to boot_contracts_ids)
+        // because they are deployed during session boot, so the deployment
+        // plan should not attempt to emit them as emulated publishes.
+        for (id, (contract, ast)) in SBTC_BOOT_CONTRACTS.iter() {
+            boot_contracts_ids.insert(id.clone());
+            requirements_data.insert(id.clone(), (contract.clarity_version, ast.clone()));
+        }
```
hugo-stacks: I'm a bit confused by that, the comment say `plan should not attempt to emit them as **emulated** publishes`

Are we talking simnet here?
Also, I _believe_ sbtc boot contracts were already properly handled before (but they were optional), the difference is that now they are mandatory, but it shouldn't require more logic to get them deployed

components/clarinet-deployments/src/lib.rs
```diff
@@ -798,6 +809,90 @@ pub async fn generate_default_deployment_with_cache(
         }
     }
 
+    // When epoch 4.0 is configured for devnet, auto-deploy sbtc-registry and
+    // sbtc-token so that the pox-5 boot contract can resolve its references.
+    let has_epoch_4_0 = network_manifest
+        .devnet
+        .as_ref()
+        .is_some_and(|d| d.epoch_4_0.is_some());
+    if matches!(network, StacksNetwork::Devnet) && has_epoch_4_0 {
```
hugo-stacks: Let's always deploy sbtc smart contracts and stop making them optional, treat them as boot contracts for every situations (even if not using pox-5 / epoch 4.0)

They will be mandatory soon enough, let's skip the step where we make it epoch gated

components/clarinet-deployments/src/lib.rs
brady-stacks: Yea, I see what you mean. I moved the handling up so they get inserted into `boot_contracts_data` and tehn treated the same as any other boot contract. 👍

components/clarinet-deployments/src/lib.rs
```diff
@@ -530,13 +531,22 @@ pub async fn generate_default_deployment_with_cache(
...
                 &settings.override_boot_contracts_source,
             )
         };
+
+        // sbtc-registry and sbtc-token are deployed during session boot
+        // (pox-5 depends on sbtc-token). They live at a non-boot address
+        // (SM3VDX...) so they're in a separate static, but they receive the
+        // same treatment as regular boot contracts here.
+        for (id, (contract, ast)) in SBTC_BOOT_CONTRACTS.iter() {
+            boot_contracts_data.insert(id.clone(), (contract.clone(), ast.clone()));
```
hugo-stacks: For a follow up PR; if sbtc are treated as boot contracts, we may want to have the ability use `override_boot_contracts_source` on it. Not sure if anyone would use it today (sbtc contracts are stable), so definitely low priority

components/clarinet-files/src/network_manifest.rs
```diff
@@ -977,10 +979,15 @@ impl NetworkManifest {
...
-                    .unwrap_or(DEFAULT_STACKS_API_IMAGE.to_string()),
+                stacks_api_image_url: devnet_config.stacks_api_image_url.take().unwrap_or_else(
+                    || {
+                        if devnet_config.epoch_4_0.is_some() {
+                            DEFAULT_STACKS_API_IMAGE_EPOCH_4_0.to_string()
+                        } else {
+                            DEFAULT_STACKS_API_IMAGE.to_string()
+                        }
+                    },
+                ),
```
hugo-stacks: My take, let's not add transient complexity, if some one wants to use epoch 4 devnet, they need to set their custom node/miner/api image in their devnet.toml

components/clarinet-deployments/src/lib.rs
brady-stacks: 👍 https://github.com/stx-labs/clarinet/issues/2441

components/clarinet-deployments/src/lib.rs
```diff
@@ -798,6 +808,85 @@ pub async fn generate_default_deployment_with_cache(
...
+        );
+
+        // Deploy sbtc-registry before sbtc-token (order matters).
+        for name in ["sbtc-registry", "sbtc-token"] {
+            let contract_id = QualifiedContractIdentifier::new(
+                sbtc_mainnet_principal.clone(),
+                ContractName::try_from(name).unwrap(),
+            );
+
+            // Skip if already added as an explicit requirement
```
hugo-stacks: We don't want to do the `current-aggregate-pubkey` re-write in this case?

hugo-stacks (changes requested):

components/clarinet-deployments/src/lib.rs
brady-stacks: the check was redundant anyway. Now there's just 1 place for this logic. Nice simplification 👍

hugo-stacks (dismissed):
lgtm appart from the failing build

hugo-stacks (approved):

PR #2431, opened 2026-06-11
fix: remote_data fetch getContractAST

### LINKED ISSUE #2430: MXS doesn't fetch contracts meta data
## Issue

With remote data enabled, Clarinet doesn't fetch metadata for remote contract such as source, ast, and likely interface.

## Cause

We try to get the data from `session.contracts`, but it only contains the local contract.
That works for local contracts, but remote contract should go through the interpreter and datastore existing calls to fetch the data from http calls

### Note about building the AST
We want to build the ast locally from the fetched source. Instead of fetching the ast, the wouldn't contain the PSEs `span`s in prod. Related: https://github.com/stx-labs/clarinet/pull/1947

## Reproduce

In `remote-data.test.ts` add
```ts
  it.only("can get remote contracts ast", async () => {
    await simnet.initEmptySession({
      enabled: true,
      api_url,
      // the counter contract is deployed at 41613
      initial_height: 800000,
    });
    simnet.getContractAST("ST1S5ZGRZV5K4S9205RWPRTX9RGS9JV40KQMR4G1J.dia-oracle");
  });
```

Build the sdk with and run the test
```sh
$ pnpm run build:sdk-wasm:dev # get clean stack traces
# or 
$ pnpm run build:sdk-wasm

$ pnpm --filter ./components/clarinet-sdk/node t tests/remote-data.test.ts 
```

## Related

#2429 properly surfaces the error

## Initially reported in rendezvous
From @radu-stacks 
> Friedger ran into an issue with Rendezvous and it seems to be related to clarinet getContractAST parsing while using remote_data. Basically when remote data is on, getContractAST crashes on a mainnet sBTC requirement.
>Project: https://github.com/friedger/clarity-pox-5-pool/
>Comment explaining the issue: https://github.com/friedger/clarity-pox-5-pool/blob/d8a5b19029e17c15bad86dade9dfdd1cea2ba44c/tests/rv-fuzz.test.ts#L12
> Issue: https://github.com/stx-labs/rendezvous/issues/264

brady-stacks:
- closes #2430 

remote_data fetch can get mainnet contracts source AND ast

### DIFF
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -668,41 +668,17 @@ impl SDK {
     #[wasm_bindgen(js_name=getContractSource)]
     pub fn get_contract_source(&mut self, contract: &str) -> Option<String> {
         let contract_id = Session::desugar_contract_id(&self.deployer, contract).ok()?;
-        let session = self.get_session_mut();
-        if let Some(contract) = session.contracts.get(&contract_id) {
-            return Some(contract.code.clone());
-        }
-        session.interpreter.get_contract_source(&contract_id)
+        self.get_session_mut().get_contract_source(&contract_id)
     }
 
     #[wasm_bindgen(js_name=getContractAST)]
     pub fn get_contract_ast(&mut self, contract: &str) -> Result<IContractAST, JsError> {
         let contract_id =
             Session::desugar_contract_id(&self.deployer, contract).map_err(|e| JsError::new(&e))?;
-        let session = self.get_session_mut();
-        if let Some(contract) = session.contracts.get(&contract_id) {
-            return Ok(encode_to_js(&contract.ast)?.unchecked_into::<IContractAST>());
-        }
-
-        let source = session
-            .interpreter
-            .get_contract_source(&contract_id)
-            .ok_or_else(|| JsError::new(&format!("contract {contract_id} not found")))?;
-        let analysis = session
-            .interpreter
-            .get_contract_analysis(&contract_id)
-            .ok_or_else(|| {
-                JsError::new(&format!("contract analysis for {contract_id} not found"))
-            })?;
-        let ast = clarity::vm::ast::build_ast(
-            &contract_id,
-            &source,
-            &mut (),
-            analysis.clarity_version,
-            analysis.epoch,
-        )
-        .map_err(|e| JsError::new(&e.to_string()))?;
-
+        let ast = self
+            .get_session_mut()
+            .get_contract_ast(&contract_id)
+            .map_err(|e| JsError::new(&e))?;
         Ok(encode_to_js(&ast)?.unchecked_into::<IContractAST>())
     }
 
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -670,6 +670,7 @@ impl SDK {
         let contract_id =
             Session::desugar_contract_id(&self.deployer, contract).map_err(|e| JsError::new(&e))?;
         self.get_session_mut()
+            .interpreter
             .get_contract_source(&contract_id)
             .map_err(|e| JsError::new(&e))
     }
--- a/components/clarinet-sdk-wasm/src/core.rs
+++ b/components/clarinet-sdk-wasm/src/core.rs
@@ -666,9 +666,12 @@ impl SDK {
     }
 
     #[wasm_bindgen(js_name=getContractSource)]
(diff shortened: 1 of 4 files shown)


components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -666,24 +666,44 @@ impl SDK {
...
+            &contract_id,
+            &source,
+            &mut (),
+            analysis.clarity_version,
+            analysis.epoch,
+        )
+        .map_err(|e| JsError::new(&e.to_string()))?;
 
-        Ok(encode_to_js(&contract.ast)?.unchecked_into::<IContractAST>())
+        Ok(encode_to_js(&ast)?.unchecked_into::<IContractAST>())
```
hugo-stacks, line 686: Conceptually, core.rs' SDK tries to be just a wasm wrapper around Session (which his itself a wrapper around Interpreter).
Could you check if most of this logic could be move to session.rs instead?
ignore if it seems less clean to put it there

components/clarity-repl/src/repl/session.rs
```diff
@@ -1067,6 +1067,42 @@ impl Session {
...
 
+    pub fn get_contract_source(
+        &mut self,
+        contract_id: &QualifiedContractIdentifier,
+    ) -> Result<Option<String>, String> {
+        if let Some(contract) = self.contracts.get(contract_id) {
+            return Ok(Some(contract.code.clone()));
+        }
+        self.interpreter.get_contract_source(contract_id)
+    }
```
hugo-stacks: Actually, could that just be it:

```suggestion
    pub fn get_contract_source(
        &mut self,
        contract_id: &QualifiedContractIdentifier,
    ) -> Result<Option<String>, String> {
        self.interpreter.get_contract_source(contract_id)
    }
```

I think the local contracts do leave in the datastore / interpreter as well, and it would be same performance wise. Just simpler, essentially removing the need for `session.get_contract_source()` in favor of `session.interpreter.get_contract_source()`
To be confirmed with remote-data both enabled and disabled

components/clarinet-sdk/node/tests/remote-data.test.ts
```diff
@@ -106,6 +106,33 @@ describe("simnet remote interactions", { retry: 2 }, async () => {
...
   });
 
+  it("can get remote contract source", async () => {
+    await simnet.initEmptySession({
+      enabled: true,
+      api_url,
+      initial_height: 56230,
+    });
+    // trigger the remote contract to be fetched
+    simnet.callReadOnlyFn(counterAddress, "get-count", [], sender);
```
hugo-stacks:
```suggestion
```

Calling it first would indeed fetch and cache the contract source, but it's not _needed_.
Here it adds confusion to the test, I think the unit test should directly test `const source = simnet.getContractSource(counterAddress);`

components/clarinet-sdk/node/tests/remote-data.test.ts
```diff
@@ -106,6 +106,33 @@ describe("simnet remote interactions", { retry: 2 }, async () => {
...
+  });
+
+  it("can get remote contract AST", async () => {
+    await simnet.initEmptySession({
+      enabled: true,
+      api_url,
+      initial_height: 56230,
+    });
+    // trigger the remote contract to be fetched
+    simnet.callReadOnlyFn(counterAddress, "get-count", [], sender);
```
hugo-stacks:
```suggestion
```

Same as above

hugo-stacks (changes requested):
Just some nits regarding the tests.

Now that local and remote contract basically take the same path, I would add a test like that at the end of remote-data.test.ts

Locally, we can even observe that the first one is fast and the second is slow (after cleaning `tests/fixtures/.cache`), but there is no real clean way to test for that I think (without making it flaky). And testing that it works is enough in itself

```ts
  it("can get local and remote contract sources", async () => {
    const simnet = await initSimnet("tests/fixtures/ManifestWithMXS.toml", false);

    const localContractSource = simnet.getContractSource(`${simnet.deployer}.counter`);
    expect(localContractSource).toBeDefined();
    expect(localContractSource).toContain("define-data-var");

    const remoteContractSource = simnet.getContractSource(
      "SP1H1733V5MZ3SZ9XRW9FKYGEZT0JDGEB8Y634C7R.miamicoin-token-v2",
    );
    expect(remoteContractSource).toBeDefined();
    expect(remoteContractSource).toContain("define-data-var");
  });
```

brady-stacks:
@hugo-stacks local+remote test added 👍

hugo-stacks (approved):
lgtm!

PR #2439, opened 2026-06-16
fix: lcov asserts and BRDA numbering

### LINKED ISSUE #2437: `asserts!`not properly handled in code coverage
The code coverage report is invalid when an asserts error branch isn't hit.
I don't think that's a regression on our side, but a breaking change in lcov that is stricter now.

## Reproduce

```console
clarinet new assert
cd assert
clarinet contract new tstassert
npm i
```

```clar
# contracts/tstassert.clar
(define-public (check (n uint))
  (begin
    (asserts! (> n u1) (err u501))
    (ok true)
  )
)
```

```ts
// tests/testassert.test.ts
import { Cl } from "@stacks/transactions";
import { expect, it } from "vitest";

it("can call check", () => {
  const { result } = simnet.callPublicFn(
    "tstassert",
    "check",
    [Cl.uint(10)],
    simnet.deployer,
  );
});
```

```console
npm run test:report
genhtml lcov.info --branch-coverage -o coverage/
```

Error:
```
Reading tracefile lcov.info.
genhtml: WARNING: (format) invalid characters removed from testname in tracefile lcov.info: 'tests/tstassert.test.ts__can call check'->'tests_tstassert_test_ts__can_call_check'
        (use "genhtml --ignore-errors format,format ..." to suppress this warning)
genhtml: ERROR: (corrupt) unable to read trace file 'lcov.info': genhtml: ERROR: (inconsistent) "/Users/hugo/Sites/stacks-labs/tmp/tstassert/contracts/tstassert.clar":3: line is hit but no branches on line have been evaluated.
        To skip consistency checks, see the 'check_data_consistency' section in man lcovrc(5).
        (use "genhtml --ignore-errors inconsistent ..." to bypass this error)
        (use "genhtml --ignore-errors corrupt ..." to bypass this error)
```

## Fix

Let's ignore the WARNING for now (but could be fixed in an other PR as well).

`asserts!` acts as en early return, from the branch-coverage point of view that's like a two-way branch (`if (cond) { continue } else { return err }`)

So we need to tell lcov that the first branch is taken in this case.

Double check if other branching functions (if, match, etc)
Make sure the lcov file is consistent, Claude highlighted some other issues on my end
> Fix your block numbering while you're in there. You wrote block 10. The manpage requires within a particular line, block is an integer numbered from zero with no gaps. A lone block 10 can trip the "branch ID does not follow expected integer sequence" check in strict LCOV 2.x. Number branch blocks per line starting at 0.

brady-stacks:
- closes #2437 

- BRDA uses sequential block ids (lcov 2.x)
- asserts work similar to other branching fns
- did_finish_Eval was double counting with new will_begin_eval counting

### DIFF
--- a/components/clarity-repl/src/analysis/coverage.rs
+++ b/components/clarity-repl/src/analysis/coverage.rs
@@ -318,8 +318,8 @@ fn retrieve_executable_lines_and_branches(
                         }
                         NativeFunctions::Asserts => {
                             // asserts!(condition, error-value)
-                            // Branch 0 (success/continue): the condition expression
-                            // Branch 1 (error): the error value expression
+                            // Branch 0 (condition evaluated): the condition expression (always evaluated)
+                            // Branch 1 (error path): the error value expression (evaluated when condition is false)
                             let (Some(cond), Some(error_value)) = (args.first(), args.get(1))
                             else {
                                 continue;
--- a/components/clarity-repl/src/analysis/coverage.rs
+++ b/components/clarity-repl/src/analysis/coverage.rs
@@ -3,7 +3,7 @@ use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use clarity::vm::ast::ContractAST;
 use clarity::vm::errors::VmExecutionError;
 use clarity::vm::functions::define::DefineFunctionsParsed;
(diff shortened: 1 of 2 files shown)


hugo-stacks (approved):
lgtm

PR #2440, opened 2026-06-17
fix: include suggestion text in lsp diagnostics

brady-stacks:
Previously the suggestion text was only displayed in `clarinet check`

Screenshot 2026-06-17 at 9 14 47 AM

### DIFF
--- a/components/clarity-repl/src/repl/diagnostic.rs
+++ b/components/clarity-repl/src/repl/diagnostic.rs
@@ -57,10 +57,26 @@ pub fn output_diagnostic(
     output.append(&mut output_code(diagnostic, lines));
 
     if let Some(ref suggestion) = diagnostic.suggestion {
-        output.push(suggestion.to_string());
+        output.push(colorize_suggestion(suggestion));
     }
 
     output
+}
+
+/// Apply terminal colors to a suggestion string.
+/// Lines starting with "Note:" are highlighted in yellow.
+fn colorize_suggestion(suggestion: &str) -> String {
+    suggestion
+        .lines()
+        .map(|line| {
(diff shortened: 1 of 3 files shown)


jbencin-stacks:
Can you bold "Note" and/or put another newline between the diagnostic and suggestion? Just something to visually separate the two

brady-stacks:
@jbencin-stacks vscode doesn't let you markdown or stylize that text, but I did add an extra newline 👍

components/clarity-repl/src/repl/diagnostic.rs
```diff
@@ -57,12 +57,28 @@ pub fn output_diagnostic(
...
     output
 }
 
+/// Apply terminal colors to a suggestion string.
+/// Lines starting with "Note:" are highlighted in yellow.
+fn colorize_suggestion(suggestion: &str) -> String {
+    suggestion
+        .lines()
+        .map(|line| {
+            if let Some(rest) = line.strip_prefix("Note:") {
```
jbencin-stacks, line 72: I really don't like this overloading of `diagnostic.suggestion` to contain both suggestions and notes, depending on the prefix, but I guess we're kinda stuck with it since `Diagnostic` is defined externally

Ask Claude if there's a simple way to avoid this, but if not, it's fine

jbencin-stacks (approved):

components/clarity-repl/src/repl/diagnostic.rs
brady-stacks: Yeah, let's handle that here: https://github.com/stx-labs/clarinet/issues/2334

PR #2445, opened 2026-06-23
fix: avoid capacity overflow bug in static cost

brady-stacks:
The `calculate_total_cost_with_summing` had a bug that resulted in the "capacity overflow" bug that popped up from the pyth-lazer-decoder contract. This removes that function in favor of the branching function that doesn't have a summing explosion on deeply nested clarity

### screenshot showing static costs on the pyth decode contract working

Screenshot 2026-06-23 at 2 23 13 PM

Also fixes a vuln exposed by `cargo audit` - quinn-proto transitive dep for reqwest

### DIFF
--- a/components/clarity-static-cost/src/static_cost/mod.rs
+++ b/components/clarity-static-cost/src/static_cost/mod.rs
@@ -173,34 +173,39 @@ pub(crate) fn calculate_total_cost_with_branching(node: &CostAnalysisNode) -> Su
                         calculate_total_cost_with_branching(&node.children[0]);
                     let condition_static: StaticCost = condition_branching.into();
 
-                    // Add the root cost + condition's worst-case cost to each branch.
+                    // Build separate min/max bases: root cost + condition cost.
                     // Each branch is mutually exclusive, so we add root+condition to each.
-                    let mut root_and_condition = node.cost.min.clone();
-                    saturating_add_cost(&mut root_and_condition, &condition_static.max);
+                    let mut root_and_condition_min = node.cost.min.clone();
+                    saturating_add_cost(&mut root_and_condition_min, &condition_static.min);
+                    let mut root_and_condition_max = node.cost.max.clone();
+                    saturating_add_cost(&mut root_and_condition_max, &condition_static.max);
 
                     for child_cost_node in node.children.iter().skip(1) {
                         // Use _branching to avoid cartesian product explosion.
                         let branch_cost = calculate_total_cost_with_branching(child_cost_node);
                         let branch_static: StaticCost = branch_cost.into();
                         // Add min path for this branch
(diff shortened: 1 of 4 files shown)


components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -156,43 +156,6 @@ pub(crate) fn calculate_function_cost_from_native_function(
     }
 }
 
-/// total cost handling branching
-/// For non-branching we combine all paths
-pub(crate) fn calculate_total_cost_with_summing(node: &CostAnalysisNode) -> SummingExecutionCost {
```
brady-stacks, line 161: this function was causing the capacity overflow. Basically the rest of the changes are just to use the _branching variant that doesn't break

components/clarity-static-cost/src/static_cost/mod.rs
```diff
@@ -206,34 +169,43 @@ pub(crate) fn calculate_total_cost_with_branching(node: &CostAnalysisNode) -> Su
                 // Match expressions work similarly to If: evaluate condition, then evaluate matching branch
                 // The cost includes the condition evaluation plus the cost of the selected branch
                 if node.children.len() >= 2 {
-                    let condition_cost = calculate_total_cost_with_summing(&node.children[0]);
-                    let condition_total = condition_cost.add_all();
```
jbencin-stacks, line 210: This function is now unused but still defined at `cost_analysis.rs` line 230. Should it be removed?

jbencin-stacks: Actually, if we're not using `add_all()`, we may not even need `SummingExecutionCost` any more

components/clarity-static-cost/tests/mod.rs
```diff
@@ -2587,3 +2587,105 @@ fn test_static_cost_handles_mutual_trait_impl_cycle() {
...
+                    }))))
+
+        (define-public (decode-and-verify-price-feeds (update (buff 8192)))
+            (let ((verified (try! (recover-signer update))))
+                (decode-payload (get payload verified))))
+    "#};
+
+    let contract_id = QualifiedContractIdentifier::transient();
+    let epoch = StacksEpochId::Epoch2_05;
+    let clarity_version = ClarityVersion::Clarity5;
```
jbencin-stacks: Clarity 5 on Epoch 2.05?

jbencin-stacks (dismissed):
I made a couple comments but they're not blockers, address if you want

Also, in the future I'd prefer having the `cargo audit` fixes in their own PRs. We can merge those quickly and then rebase existing PRs

jbencin-stacks (approved):

PR #2447, opened 2026-06-30
fix: Override boot contracts coverage

brady-stacks:
- fix contract source fetch for boot contract overrides in coverage reports


components/clarity-repl/src/analysis/coverage.rs
```diff
@@ -63,18 +63,31 @@ impl CoverageHook {
     ) -> String {
         let mut file_content = String::new();
 
-        let mut filtered_asts = HashMap::new();
+        // Functions only depend on source text, so one copy per contract name
+        // suffices. Executable lines/branches contain expression IDs that differ
+        // between ASTs built from the same source (e.g. a boot contract override
+        // and a user-deployed contract), so we store those per contract ID.
+        let mut functions_by_name: HashMap<String, Vec<(String, u32, u32)>> = HashMap::new();
```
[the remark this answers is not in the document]
brady-stacks: I don't think this is a real issue

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -472,6 +472,21 @@ impl SDK {
...
+        // reports the correct source file for overridden contracts.
+        let project_root = manifest
+            .location
+            .parent()
+            .map(|p| p.to_path_buf())
+            .unwrap_or_default();
+        for (name, entry) in &manifest.project.override_boot_contracts_source {
+            let contract_id = QualifiedContractIdentifier::new(
+                clarity_repl::repl::boot::BOOT_TESTNET_PRINCIPAL.clone(),
+                clarity::vm::ContractName::try_from(name.as_str()).unwrap(),
```
jbencin-stacks: This will panic if user has a filename with character that's not legal in `ContractName`. I think this could be handled better, either:
- Skip file and warn
- Fail with clear error message

jbencin-stacks:
I don't see any changes to unit/integration tests here. Can we get a test or two?

brady-stacks:
@jbencin-stacks added tests for unwrap and unwrap-err 👍

components/clarity-repl/src/analysis/coverage_tests.rs
```diff
@@ -100,6 +100,94 @@ fn line_count_in_iterator() {
...
+        "  (ok (unwrap-err! res (err u1)))",
+        ")",
+    ]
+    .join("\n");
+
+    // NOTE: unwrap-err! incorrectly always evaluates both branches. This will
+    // need to change if it's fixed upstream
+    let snippets: Vec<String> = vec!["(contract-call? .contract-0 must-fail (err
+    u42))"
+        .into()];
```
jbencin-stacks: Weird formatting here, how did we get a line break in the middle of a string?

brady-stacks: 80 character max line width

brady-stacks: removed the type annotation so it fits on one line now

components/clarity-repl/src/analysis/coverage_tests.rs
```diff
@@ -100,6 +100,92 @@ fn line_count_in_iterator() {
...
+fn unwrap_ret_branching() {
+    let contract = [
+        "(define-public (must-unwrap (opt (optional uint)))",
+        "  (ok (unwrap! opt (err u1)))",
+        ")",
+    ]
+    .join("\n");
+
+    // NOTE: unwrap! incorrectly always evaluates both branches. This will
+    // need to change if it's fixed upstream
```
jbencin-stacks: This looks wrong to me. It looks like we are always counting both branches as "covered", even if one is never actually taken, because `unwrap!` uses eager evaluation.

This is not something that can just be "fixed", this could potentially alter the behavior of a deployed contract, and therefore would be a consensus-breaking change

jbencin-stacks (commented):
There's 2 things in this PR:

> fix contract source fetch for boot contract overrides in coverage reports

Looks correct, but should have a test case

> add unwrap native functions

This looks incorrect (always evaluates both branches as taken), but also why is it not a separate PR?

brady-stacks:
@jbencin-stacks removed the unwrap/unwrap-err additions 👍

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -472,6 +472,31 @@ impl SDK {
...
 
+        // Register override boot contract file paths so LCOV coverage
+        // reports the correct source file for overridden contracts.
+        let project_root = manifest
+            .location
+            .parent()
+            .map(|p| p.to_path_buf())
+            .unwrap_or_default();
+        for (name, entry) in &manifest.project.override_boot_contracts_source {
+            let contract_name = match clarity::vm::ContractName::try_from(name.as_str()) {
```
jbencin-stacks: This `match` could be simplified to `let`/`else` I think

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -472,6 +472,31 @@ impl SDK {
...
+                Ok(name) => name,
+                Err(_) => {
+                    log!(
+                        "Warning: Skipping override boot contract '{}' - invalid contract name",
+                        name
```
jbencin-stacks: We need a note about this in `CLAUDE.md`...

```suggestion
                        "Warning: Skipping override boot contract '{name}' - invalid contract name"
```

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -1199,11 +1224,30 @@ impl SDK {
...
-        if include_boot_contracts {
-            for (contract_id, (_, ast)) in clarity_repl::repl::boot::BOOT_CONTRACTS_DATA.iter() {
-                asts.insert(contract_id.clone(), ast.clone());
+        let overrides = &session.settings.override_boot_contracts_source;
+        for (contract_id, result) in session.boot_contracts.iter() {
+            let name = contract_id.name.to_string();
+            let is_override = overrides.contains_key(&name);
+
+            // Always include overridden boot contracts Only include default
+            // boot contracts when explicitly requested.
```
jbencin-stacks: Fix formatting

```suggestion
            // Always include overridden boot contracts
            // Only include default boot contracts when explicitly requested.
```

jbencin-stacks (commented):
Just a couple minor changes and I'll approve

jbencin-stacks (approved):

PR #2457, opened 2026-07-10
fix: boot contracts deploy during correct epoch

### LINKED ISSUE #1988: Boot contracts should be deployed on epoch update in Simnet
In Simnet, all boot contracts are deployed at the beginning of a session, in epoch 2_0.

But contracts such as `pox-4` shouldn't be available at this point.

Ideally, boot contracts should be deployed in `Session::update_epoch()`:
- `update_epoch` should increment epochs one by one (instead of allowing a jump from epoch 2.0 to epoch 3.2, it should go through each epoch in between (`while self.interpreter.datastore.get_current_epoch() < epoch {...}`)
- it would introduce two important breaking changes:
  - the intended change is that all boot contracts wouldn't be available at all time 
  - a side effect is the block height changes would be impacted since each epoch change is a new burn block
    - a few projects have hard coded expected block heights and would therefore be impacted

brady-stacks:
- closes #1988


components/clarity-repl/src/repl/session.rs
```diff
@@ -1126,10 +1214,25 @@ impl Session {
...
         }
+
+        let newly_deployed = deploy_boot_contracts_for_range(
+            current,
+            epoch,
+            &self.settings,
+            &mut self.interpreter,
+            &self.boot_contracts,
+        );
+        self.boot_contracts.extend(newly_deployed);
```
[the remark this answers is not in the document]
brady-stacks: Lowering the epoch is not a real use case we're concerned with

components/clarity-repl/src/repl/session.rs
```diff
@@ -1126,10 +1214,25 @@ impl Session {
...
 
     pub fn update_epoch(&mut self, epoch: StacksEpochId) {
+        // deploys contracts with epochs between the current epoch
+        // and the target epoch (exclusive of current, inclusive of target).
+        let current = self.interpreter.datastore.get_current_epoch();
+
+        // Set the epoch first so boot contracts are deployed at the correct
+        // interpreter epoch (e.g. Clarity3 contracts must deploy at Epoch30+).
         self.interpreter.set_current_epoch(epoch);
         if epoch >= StacksEpochId::Epoch30 {
```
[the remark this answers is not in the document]
brady-stacks: To avoid the breaking change we'll avoid step by step epoch updates

components/clarity-repl/src/repl/boot/mod.rs
```diff
@@ -329,3 +330,15 @@ pub fn get_boot_contract_epoch_and_clarity_version(
     };
     (epoch, clarity_version)
 }
+
+/// Returns all boot contract names, grouped by their associated epoch,
+/// sorted by epoch order. Includes sbtc-registry and sbtc-token at epoch 3.0.
+pub fn boot_contracts_by_epoch() -> BTreeMap<StacksEpochId, Vec<&'static str>> {
```
jbencin-stacks: This function is never called

components/clarinet-sdk-wasm/src/core.rs
```diff
@@ -651,8 +651,35 @@ impl SDK {
             DEFAULT_EPOCH
         });
 
-        let session = self.get_session_mut();
-        session.update_epoch(epoch);
+        self.get_session_mut().update_epoch(epoch);
+
+        // Propagate any newly deployed boot contracts into contracts_interfaces.
+        let new_entries: Vec<_> = self
```
jbencin-stacks, line 672: This does pretty much the same thing as lines 441-468. Can it be factored out to a separate function?

components/clarinet-sdk/node/tests/deployment-plan.test.ts
```diff
@@ -7,7 +7,12 @@ import { describe, expect, it, beforeEach, afterEach } from "vitest";
 import { initSimnet, generateDeployement } from "..";
 import { Cl } from "@stacks/transactions";
 
-const nbOfBootContracts = 30;
+// Boot contracts with valid interfaces deployed through Epoch31 (Clarinet.toml highest epoch):
+// 12 names × 2 addresses (genesis has no functions) + 2 sbtc contracts = 26
+const nbOfBootContractsEpoch31 = 26;
```
jbencin-stacks: Does this include the contracts from `nbOfBootContractsEpoch24` also? If so, define as:

```rust
const nbOfBootContractsEpoch31 = nbOfBootContractsEpoch24 + 8;
```

components/clarinet-sdk/node/tests/simnet-usage.test.ts
```diff
@@ -13,7 +13,12 @@ const address2 = "ST2CY5V39NHDPWSXMW9QDT3HC3GD6Q6XX4CFRK9AG";
 
 let simnet: Simnet;
 
-const nbOfBootContracts = 30;
+// Boot contracts with valid interfaces deployed through Epoch31 (the highest epoch in Clarinet.toml):
+// 12 names × 2 addresses (genesis has no functions and deploys as a snippet)
+// + 2 sbtc contracts at Epoch30 = 26
+const nbOfBootContracts = 26;
```
jbencin-stacks: Is this at genesis? If so, name it so we don't confuse with other epochs:


```suggestion
const nbOfBootContractsAtGenesis = 26;
```

jbencin-stacks: Also, this name is a bit long, can we rename all these to like `bootContractsAtGenesis` instead?

I know you didn't name this initially, but it's generally not recommended to use [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) in strongly-typed languages like Rust, it just adds noise

components/clarity-repl/src/repl/session.rs
```diff
@@ -178,7 +265,9 @@ impl Session {
 
         set_up_accounts(&settings.initial_accounts, &mut interpreter);
         let boot_contracts = if with_boot_contracts {
-            deploy_boot_contracts(&settings, &mut interpreter)
+            // Deploy all boot contracts whose epoch <= the interpreter's current epoch.
+            let initial_epoch = interpreter.datastore.get_current_epoch();
+            deploy_boot_contracts(&settings, &mut interpreter, Some(initial_epoch))
```
jbencin-stacks, line 270: This will use Epoch 2.05. We use two different default Epochs in this project:

 - The session/datastore default: Epoch2_05 (datastore.rs:764)
 - `DEFAULT_EPOCH` constant: Epoch34 (clarinet-defaults/src/lib.rs:4)

We should pick one for all initialization paths. I don't know which one you intended to use here, you might want to change this, or maybe not, but we should open an issue for this

brady-stacks: Made an issue for this: https://github.com/stx-labs/clarinet/issues/2466
Most likely we can just bump anyhting using 2.05 up 👍

hugo-stacks (approved):
lgtm! Tested in CLI (not sdk, but unit tests look good)

PR #2463, opened 2026-07-15
chore: update node and signer images to 4.0.0

brady-stacks:


jbencin-stacks (dismissed):

components/clarinet-defaults/src/lib.rs
```diff
@@ -1,5 +1,5 @@
 use clarity::types::StacksEpochId;
 use clarity::vm::ClarityVersion;
 
-pub const DEFAULT_EPOCH: StacksEpochId = StacksEpochId::Epoch34;
+pub const DEFAULT_EPOCH: StacksEpochId = StacksEpochId::Epoch40;
```
hugo-stacks: We want to wait for a bit (might a week or two) to release default to 4.0.
Make it closer to activation date on mainnet.

I recall there might be a unit test that test the default epoch against the stacks-core image version? We may want to skip it for now

hugo-stacks (changes requested):

components/clarinet-defaults/src/lib.rs
brady-stacks: Maybe it makes more sense to wait on both image and epoch defaults then?

brady-stacks: Ok, changed it back to 3.4 and skipped the test 👍

hugo-stacks (approved):
lgtm

PR #2475, opened 2026-07-23
feat: Add rename_builtin lint 

### LINKED ISSUE #2472: Clarity 6 lint warning that "with-stacking" has been replaced with "with-staking"
`with-stacking` has been removed in favor of `with-staking` in Clarity 6.
Clarity will properly reject the old `with-stacking`, but a Clarity 6 and above lint rule to explain that it's now `with-staking` could help

brady-stacks:
New lint: `rename_builtin`

Will catch `with-stacking -> with-staking`

- closes #2472


components/clarity-repl/src/analysis/lints/rename_builtin.rs
```diff
@@ -0,0 +1,235 @@
...
+    fn run(mut self, contract_analysis: &'a ContractAnalysis) -> AnalysisResult {
+        if contract_analysis.clarity_version < ClarityVersion::Clarity6 {
+            return Ok(self.diagnostics);
+        }
+
+        traverse(&mut self, &contract_analysis.expressions);
+        Ok(self.diagnostics)
+    }
+
+    fn run_expressions(mut self, expressions: &'a [SymbolicExpression]) -> Vec<Diagnostic> {
```
jbencin-stacks: Do we need `run` and `run_expressions()`? They are the same thing but one wrapped in an `Ok(...)`

components/clarity-repl/src/analysis/lints/mod.rs
```diff
@@ -10,6 +10,7 @@ mod error_const;
 mod flatten_variadic;
 mod noop;
 mod panic;
+mod renamed_with_stacking;
```
jbencin-stacks: This could be more generic, so we can add to this lint in the future without doing additional AST passes. How about `rename_builtin`?

components/clarity-repl/src/analysis/lints/rename_builtin.rs
```diff
@@ -0,0 +1,432 @@
...
+    for expr in expressions {
+        let Some(list) = expr.match_list() else {
+            continue;
+        };
+        let Some(first) = list.first().and_then(|e| e.match_atom()) else {
+            continue;
+        };
+        if !matches!(
+            first.as_str(),
+            "define-private" | "define-public" | "define-read-only"
```
jbencin-stacks: Can we match on enums instead of hard-coded strings?

components/clarity-repl/src/analysis/lints/rename_builtin.rs
```diff
@@ -0,0 +1,432 @@
...
+        self.active_annotation = get_index_of_span(self.annotations, &expr.span);
+        if self.allow() {
+            return true;
+        }
+
+        self.diagnostics.push(Diagnostic {
+            level: self.level.clone(),
+            message: format!(
+                "`{}` was renamed to `{}` in Clarity {}. Replace this call with `{}`.",
+                entry.old_name, entry.new_name, entry.since, entry.new_name,
```
jbencin-stacks: `Display` for `ClarityVersion` will print "Clarity $version", so this line will print "in Clarity Clarity $version"

components/clarity-repl/src/analysis/lints/mod.rs
```diff
@@ -10,6 +10,7 @@ mod error_const;
 mod flatten_variadic;
 mod noop;
 mod panic;
+mod rename_builtin;
```
jbencin-stacks: Claude brought up a good point

> `rename_builtin` reads as an imperative. `#[allow(rename_builtin)]` parses as a command to rename something

How about `renamed_builtin` or `builtin_renamed`? I think the first one, since we generally use verb-noun order

jbencin-stacks:
It looks like this PR also fixes `restrict-assets?` in other lints. Add these tests to `unused_const` to confirm:

```rust
      #[test]
      fn used_in_restrict_assets_allowances() {
          let mut settings = SessionSettings::default();
          settings
              .repl_settings
              .analysis
              .enable_lint(UnusedConst::get_name(), Level::Warning);

          let mut session = Session::new_without_boot_contracts(settings);
          session.update_epoch(clarity::types::StacksEpochId::Epoch33);

          #[rustfmt::skip]
          let snippet = indoc!("
              (define-constant TRANSFER_AMOUNT u1000)

              (define-public (my-func)
                  (restrict-assets? tx-sender ((with-stx TRANSFER_AMOUNT)) true))
          ").to_string();

          let (_, result) = session
              .formatted_interpretation(snippet, Some("checker".to_string()), false, None)
              .expect("Invalid code snippet");
          assert_eq!(result.lint_diagnostics.len(), 0);
      }

      #[test]
      fn used_as_restrict_assets_owner() {
          let mut settings = SessionSettings::default();
          settings
              .repl_settings
              .analysis
              .enable_lint(UnusedConst::get_name(), Level::Warning);

          let mut session = Session::new_without_boot_contracts(settings);
          session.update_epoch(clarity::types::StacksEpochId::Epoch33);

          #[rustfmt::skip]
          let snippet = indoc!("
              (define-constant CONTRACT_OWNER tx-sender)

              (define-public (my-func)
                  (restrict-assets? CONTRACT_OWNER ((with-stx u1000)) true))
          ").to_string();

          let (_, result) = session
              .formatted_interpretation(snippet, Some("checker".to_string()), false, None)
              .expect("Invalid code snippet");
          assert_eq!(result.lint_diagnostics.len(), 0);
      }

```

jbencin-stacks (approved):

PR #2479, opened 2026-07-28
refactor: Replace `stackslib` with `stacks-codec` where possible

jbencin-stacks:
### Description

Since we've moved a lot of structs and ser/deser logic from `stackslib` into `stacks-codec`, we can remove the dependency on `stackslib` for a couple crates in this repo

This doesn't reduce compile time when building the full `clarinet` binary, since it still requires `stackslib`, but it could if we're building some subset that doesn't require it

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -715,9 +715,9 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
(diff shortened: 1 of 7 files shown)


brady-stacks (approved):
lgtm

PR #2483, opened 2026-07-31
feat: sdk debug adapter

brady-stacks:
Debugging works with any existing clarity test via the same vitest mechanism existing previously. The debug codelens allows running the debugger on specific `it` blocks and the debugger can be set up manually in other editors via the `clarinet dap` cli command

Note: "Debug" code lens is turned off by default

https://github.com/user-attachments/assets/223798a6-acf7-415c-95c8-61fe70cef524

Followup work: https://github.com/stx-labs/clarinet/issues/2485


brady-stacks:
We might want to change the name `startDebugServer` to `initSimnetDebug` or similar (to match `initSimnet`)

components/clarity-vscode/client/src/clientNode.ts
```diff
@@ -1,36 +1,304 @@
...
+          const disposable = vscode.debug.onDidTerminateDebugSession((session) => {
+            if (session.name !== sessionName) return;
+            debugOutput.info("  debug session terminated — killing dap server");
+            dapProcess.kill();
+            disposable.dispose();
+          });
+
+          // Run the selected test file directly. Passing the file path avoids
+          // relying on project-specific Vitest include patterns or extra config files.
+          const cmd = `CLARINET_DEBUG_PORT=${sdkPort} npx vitest run ${JSON.stringify(relativeFile)} -t ${JSON.stringify(testName)}`;
```
jbencin-stacks: Using the output of `JSON.stringify()` directly in a shell command, with no quoting or escaping, is kind of dangerous

components/clarity-repl/src/repl/debug/dap/mod.rs
```diff
@@ -103,12 +153,38 @@ impl DAPDebugger {
...
+            Some(state) => state.reset_for_new_call(contract_id, snippet),
+            None => self.state = Some(DebugState::new(contract_id, snippet)),
+        }
+    }
+
+    /// Run the DAP attach handshake: process `initialize`, `attach`,
+    /// `setBreakpoints`, and `configurationDone`, then return.
+    /// After this returns the server is ready to accept SDK contract calls.
+    pub fn init_attach(&mut self) -> Result<(), ParseError> {
+        while !self.config_done {
```
jbencin-stacks, line 198: This can occupy 100% of the CPU on a graceful disconnect. `wait_for_command()` will return `Ok(true)` indefinitely.

Should return an error to exit the loop in this case instead

jbencin-stacks (approved):
I'm going to approve this as-is, since we have something to ship that does work, but I have a document of follow-up items that we can work on in future PRs

jbencin-stacks (approved):

PR #2495, opened 2026-08-14
fix(static-cost): price epoch 4.0 with costs-5

jbencin-stacks:
### Description

Static cost analysis prices epoch 4.0 with the `Costs4` table, but the VM prices it with `Costs5`. Since `DEFAULT_EPOCH` is `Epoch40`, reported costs are wrong for most users.

This PR:

- maps epoch 4.0 to `Costs5`
- adds a test comparing our epoch → cost table mapping against `LimitedCostTracker::default_cost_contract_for_epoch`, which is what the VM consults, so the two can't silently diverge again

Three hash-cost tests compared a 1-byte input against a 32-byte one. `Costs5` prices hashing per 512-byte block, so those now cost the same and the assertion no longer proved anything; they compare 1 byte against 2048 bytes instead.

#### Breaking change?

No API change, but reported costs for epoch 4.0 change — that's the fix.

---

### Checklist

### DIFF
--- a/components/clarity-static-cost/src/static_cost/cost_analysis.rs
+++ b/components/clarity-static-cost/src/static_cost/cost_analysis.rs
@@ -2424,27 +2424,27 @@ mod tests {
         assert_eq!(result, None);
     }
 
-    /// Verify that hash function costs vary with input buffer size.
-    /// The runtime charges based on the serialized size of the input value,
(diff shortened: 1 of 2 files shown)


brady-stacks (approved):
LGTM 👍

brady-stacks (approved):

PR #2505, opened 2026-08-25
fix: vitest-helper path derivation

hugo-stacks:
More robust `sdkDir` package path derivation.

Especially useful when working with a local build of the SDK


brady-stacks (approved):
LGTM

Issue #1655, opened 2025-02-01
Improve clarinet requirement handling


hugocaillard:
## Context

Clarinet requires developers to manually add requirements with:
```
clarinet requirements add <contract principal>
```
Clarinet will fetch the contract on mainnet on testnet, detect all of requirements dependencies (other contract fetched by the requirements)

## Caveats

- Developers can often forget to do so
  - It's even more easy to forgot to remove a dependency that was wrongly added (or not needed anymore
- In the context of remote data fetching (#1503), it's possible that a contract is in the requirements (making clarinet check valid), but was actually deployed at a higher height than the once specified in the remote data settings

## Ideas

- Clarinet could aumatically detect local contract dependencies and load them, without the need to manually add or list requirements in the manifest. We could fully remove the concept of "dependencies"
- If `remote_data` is enabled, clarinet should check that the mainnet/testnet dependencies are valid (available on the right network and deployed at a height lower than the one specified in the remote data settings)

obycode:
Agreed, I think this makes it easier for users with not much downside. Maybe `clarinet` can output some info about dependencies that it is pulling in whenever it adds a new contract into the deployment plan, so that way users can still notice that they are interacting with additional contracts.

PR #2508, opened 2026-08-27
feat: remove `requirements` in favor of auto-detection

brady-stacks:
- related to #1655  (partial fix)

- Removes the need to declare requirements explicitly. Clarinet now auto-detects external contract dependencies by scanning contract ASTs before seeding the requirements queue, so references are pulled in automatically.
  - Adds per-network address remapping for the "contract already exists on testnet" case: when a testnet override differs from `contract_id`, Clarinet skips re-deploying the contract and instead rewrites its identifier in source text at broadcast time, same as for boot contracts and sBTC.
 
- `generate_deployment` reads user contract sources early, builds lightweight ASTs (skip_analysis: true), and calls `ASTDependencyDetector` to collect external references as auto-detected requirements.


jbencin-stacks:
Opened PR here with some issues and associated tests: brady-stacks/clarinet#12

components/clarinet-deployments/src/lib.rs
```diff
@@ -777,15 +893,21 @@ pub async fn generate_default_deployment_with_cache(
...
                             &manifest.project.cache_location,
                             &file_accessor,
                             api_base_url,
                         )
-                        .await?;
+                        .await
+                        {
+                            Ok(contract) => contract,
+                            Err(error) if explicit_ids.contains(&contract_id) => return Err(error),
+                            Err(_) => continue,
```
jbencin-stacks: This can trigger an infinite loop if a contract cannot be retrieved

I recommend tracking which requests end up here, and not attempting them again

components/clarinet-files/src/project_manifest.rs
```diff
@@ -752,6 +753,35 @@ mod tests {
             .contains_key("pox-x"));
     }
 
+    #[test]
+    fn test_requirements_parsing() {
+        let manifest_str = r#"
```
jbencin-stacks: Use `indoc!` here

jbencin-stacks:
Some more tests that should pass before merging: brady-stacks/clarinet#13

PR #2513, opened 2026-08-31
fix: sbtc funding

### LINKED ISSUE #2511: sbtc_balance is a no-op on a stock project: sbtc-deposit is not a boot contract, and genesis funding is gated on an unrelated requirement
## Summary

`sbtc_balance` in `settings/Devnet.toml` is a no-op on a stock `clarinet new` project, on **both simnet and devnet**. Every account holds 0 sBTC and `get-total-supply` is `(ok u0)`, despite the scaffold writing `sbtc_balance = 1_000_000_000` for all ten accounts.

The root cause is that `sbtc-deposit` is not treated as a boot contract, even though it is the only path by which genesis sBTC balances are minted. Since PoX-5, the sBTC contracts are part of the protocol surface — `pox-5` holds a `contract-call?` and a `with-ft` reference to `sbtc-token`, and the bond flow's sBTC leg is unusable without them. `sbtc-registry` and `sbtc-token` are already handled this way; `sbtc-deposit` was left behind.

Tested with `clarinet` 3.23.1 / `@stacks/clarinet-sdk` 3.23.2, node v24.20.0, darwin arm64. Line references below are against `stx-labs/clarinet@a62483f7`.

## Reproduction

```bash
clarinet new g01repro && cd g01repro
printf '(define-read-only (noop) u1)\n' > contracts/noop.clar
cat >> Clarinet.toml <<'EOF'

[contracts.noop]
path = "contracts/noop.clar"
epoch = "latest"
EOF
grep -c 'sbtc_balance = 1_000_000_000' settings/Devnet.toml   # -> 10
npm i
node -e 'import("@stacks/clarinet-sdk").then(async m=>{
  const s = await m.initSimnet();
  const A = "SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4";
  for (const c of ["sbtc-registry","sbtc-token","sbtc-deposit"]) {
    let src = null; try { src = s.getContractSource(`${A}.${c}`) } catch {}
    console.log(c, src ? "PRESENT" : "ABSENT");
  }
  console.log("epoch:", s.currentEpoch);
  console.log(s.callReadOnlyFn(`${A}.sbtc-token`, "get-total-supply", [], s.deployer).result);
})'
```

Actual:

```
10
sbtc-registry PRESENT
sbtc-token    PRESENT
sbtc-deposit  ABSENT
epoch: 4.0
total-supply: (ok u0)
```

Expected: `(ok u10000000000)` — ten accounts × 10 sBTC.

`deployments/default.simnet-plan.yaml` faithfully carries `sbtc-balance: '1000000000'` into the genesis block, so the plan is right; nothing consumes it.

## Why it happens

**1. `sbtc-deposit` is excluded from the boot set.** `SBTC_BOOT_CONTRACTS` contains only two entries ([`clarity-repl/src/repl/boot/mod.rs:287-288`](https://github.com/stx-labs/clarinet/blob/main/components/clarity-repl/src/repl/boot/mod.rs#L287-L288)):

```rust
let contracts: [(&str, &str); 2] = [
    ("sbtc-registry", SBTC_REGISTRY_SOURCE),
    ("sbtc-token", SBTC_TOKEN_SOURCE),
];
```

and the exclusion is deliberate ([`boot/mod.rs:48-50`](https://github.com/stx-labs/clarinet/blob/main/components/clarity-repl/src/repl/boot/mod.rs#L48-L50)):

```rust
// sbtc-registry and sbtc-token are deployed as boot contracts. sbtc-deposit is
// not a boot contract, but we include it here for contract-address mapping.
pub const SBTC_CONTRACTS_NAMES: &[&str] = &["sbtc-registry", "sbtc-token", "sbtc-deposit"];
```

There is no vendored `sbtc-deposit.clar` next to `sbtc-registry.clar` and `sbtc-token.clar` in `components/clarity-repl/src/repl/boot/`.

**2. Genesis funding calls a contract that isn't there.** `fund_genesis_account_with_sbtc` mints by calling `SM3VDX….sbtc-deposit::complete-deposit-wrapper` ([`clarinet-deployments/src/lib.rs:157-203`](https://github.com/stx-labs/clarinet/blob/main/components/clarinet-deployments/src/lib.rs#L157-L203)).

**3. The funding call is gated on an unrelated opt-in.** It only runs when the deployment plan contains a publish transaction for `sbtc-deposit` ([`lib.rs:236-237`](https://github.com/stx-labs/clarinet/blob/main/components/clarinet-deployments/src/lib.rs#L236-L237), [`lib.rs:261`](https://github.com/stx-labs/clarinet/blob/main/components/clarinet-deployments/src/lib.rs#L261)):

```rust
if !should_mint_sbtc && contract_id == *SBTC_DEPOSIT_MAINNET_ADDRESS {
    should_mint_sbtc = true;
}
…
if should_mint_sbtc {
    fund_genesis_account_with_sbtc(session, deployment);
}
```

and such a transaction only exists if the user declares an sBTC **project requirement**, which triggers the auto-add at [`lib.rs:658-670`](https://github.com/stx-labs/clarinet/blob/main/components/clarinet-deployments/src/lib.rs#L658-L670). So a genesis field that the scaffold populates by default silently depends on a `[[project.requirements]]` entry the scaffold does not write.

**4. Failure is swallowed.** Even when the call is made, its result is discarded ([`lib.rs:193`](https://github.com/stx-labs/clarinet/blob/main/components/clarinet-deployments/src/lib.rs#L193)):

```rust
let _ = session.call_contract_fn(
    &SBTC_DEPOSIT_MAINNET_ADDRESS.to_string(),
    "complete-deposit-wrapper",
    …
);
```

No warning is printed on any path. A user sees a configured balance, a plan that carries it, and a zero balance, with nothing in between to explain it.

**5. Devnet has the same hole, one layer down.** On devnet the coordinator funds accounts only when it observes a `<deployer>.sbtc-deposit` contract-deployment transaction in a block ([`stacks-network/src/chains_coordinator.rs:474-494`](https://github.com/stx-labs/clarinet/blob/main/components/stacks-network/src/chains_coordinator.rs#L474-L494)). But a stock devnet plan does not deploy it — `clarinet deployments generate --devnet` on the repro project above emits:

```yaml
- transaction-type: requirement-publish
  contract-id: SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-registry
- transaction-type: requirement-publish
  contract-id: SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token
- transaction-type: contract-publish
  contract-name: noop
```

`sbtc-registry` and `sbtc-token` are published unconditionally — i.e. devnet already treats them as always-present — but `sbtc-deposit` is absent, so `fund_genesis_account` never fires and `sbtc_balance` no-ops there too. Meanwhile the orchestrator unconditionally points the node's PoX-5 at these contracts ([`orchestrator.rs:894-896`](https://github.com/stx-labs/clarinet/blob/main/components/stacks-network/src/orchestrator.rs#L894-L896)):

```rust
pox_5_sbtc_contract = "{deployer_addr}.sbtc-token"
pox_5_sbtc_registry_contract = "{deployer_addr}.sbtc-registry"
pox_5_bond_admin = "{deployer_addr}"
```

**6. The existing test can't catch it.** `can_fund_initial_sbtc_balance` ([`clarinet-deployments/tests/genesis_accounts_funding.rs:84-127`](https://github.com/stx-labs/clarinet/blob/main/components/clarinet-deployments/tests/genesis_accounts_funding.rs#L84-L127)) hand-builds a batch containing `sbtc-registry`, `sbtc-token` **and** `sbtc-deposit` from `./fixtures/`. It exercises the funding code with the precondition already satisfied, which is exactly the precondition a real project lacks.

## How it should work

The sBTC contracts are now part of the PoX-5 / Bitcoin bonds protocol, not an optional third-party dependency. They should be boot contracts on every network, and genesis funding should be unconditional.

Concretely:

1. **Vendor `sbtc-deposit.clar`** into `components/clarity-repl/src/repl/boot/` alongside `sbtc-registry.clar` and `sbtc-token.clar`, and add it to `SBTC_BOOT_CONTRACTS` after `sbtc-token` (it depends on both registry and token). The ordering machinery in `SBTC_BOOT_CONTRACTS` and `deploy_boot_contracts` already handles this — `session.rs:110` deploys the sBTC set before the regular boot contracts precisely so `pox-5` can resolve against `sbtc-token`.

2. **Drop the `should_mint_sbtc` gate.** With `sbtc-deposit` always deployed, `fund_genesis_account_with_sbtc` can run whenever the genesis spec has a wallet with a non-zero `sbtc_balance`. Delete the flag and the scan at `lib.rs:212/236-237/261`.

3. **Stop discarding the result.** Replace `let _ =` at `lib.rs:193` with a check that prints a warning (or fails the session) naming the wallet and the underlying error. A silent mint failure is indistinguishable from a mint that never ran.

4. **Do the same on devnet.** Emit `sbtc-deposit` as a `requirement-publish` unconditionally, next to the `sbtc-registry` and `sbtc-token` entries the plan already generates, so the coordinator's existing detection at `chains_coordinator.rs:484` fires on a stock project. This also makes the node's `pox_5_sbtc_contract` / `pox_5_sbtc_registry_contract` settings meaningful by default rather than pointing at contracts a stock devnet may not have.

5. **Drop the `[[project.requirements]]` auto-add** at `lib.rs:658-670` once the above lands — it becomes dead, and it is the mechanism that made the dependency invisible in the first place. Declaring the sBTC requirement explicitly should remain harmless (the `session.boot_contracts.contains_key` skip at `lib.rs:243` already de-duplicates), but should no longer be load-bearing.

6. **Update the test.** Rewrite `can_fund_initial_sbtc_balance` to run against a plan with **no** sBTC transactions, so it asserts the property that actually matters: a stock project with `sbtc_balance` set gets funded.

## Current workaround

Declare sBTC as a project requirement, which pulls in `sbtc-deposit` via the auto-add:

```toml
[[project.requirements]]
contract_id = "SM3VDXK3WZZSA84XXFKAFAF15NNZX32CTSG82JFQ4.sbtc-token"
```

After this, simnet reports `total-supply: (ok u10000000000)`. Nothing in the scaffold, the `Devnet.toml` comments, or https://docs.stacks.co/clarinet/local-blockchain-development mentions that `sbtc_balance` depends on this.

hugo-stacks:
Fix #2511


brady-stacks (approved):
Looks great 👍

PR #2517, opened 2026-09-01
feat: Post-conditions in simnet

jbencin-stacks:
### Description

Add optional post-conditions in simnet

The following Clarinet SDK methods can now accept the optional parameter `postConditions`:

  - `callPublicFn()`
  - `callPrivateFn()`
  - `deployContract()`
  - `transferSTX()`

where `postConditions` is an array of objects or hex strings, defined in as `@stacks/transactions` as:

```typescript
postConditions?: (PostCondition | PostConditionWire | string)[];
```

To preserve backwards compatibility, using post-conditions in Clarinet SDK methods is completely optional and never required

#### Breaking change?

No

#### Using Structured Objects

```typescript
  import { Cl, Pc } from "@stacks/transactions";
  import { initSimnet } from "@stacks/clarinet-sdk";

  const simnet = await initSimnet();
  const sender = "ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5";

  const receipt = simnet.callPublicFn(
    "counter",
    "increment",
    [],
    sender,
    {
      postConditions: [
        Pc.principal(sender)
          .willSendEq(1_000_000)
          .ustx(),
      ],
    },
  );

  console.log(Cl.prettyPrint(receipt.result));
```

#### Using Hex Strings

```typescript
  import {
    Cl,
    Pc,
    postConditionToHex,
  } from "@stacks/transactions";
  import { initSimnet } from "@stacks/clarinet-sdk";

  const simnet = await initSimnet();
  const sender = "ST1SJ3DTE5DN7X54YDH5D64R3BCB6A2AG2ZQ8YPD5";

  const postConditionHex = postConditionToHex(
    Pc.principal(sender)
      .willSendEq(1_000_000)
      .ustx(),
  );

  const receipt = simnet.callPublicFn(
    "counter",
    "increment",
    [],
    sender,
    {
      postConditions: [postConditionHex],
    },
  );

  console.log(Cl.prettyPrint(receipt.result));
```

### Checklist

### DIFF
--- a/components/clarity-repl/src/repl/session.rs
+++ b/components/clarity-repl/src/repl/session.rs
@@ -726,6 +726,19 @@ impl Session {
             }]);
         }
 
+        // TokenTransfer payloads do not run the post-condition checker. The
+        // mode is still part of the transaction wire and must be valid for the
+        // current epoch, but it has no effect when the condition list is empty.
+        let current_epoch = self.interpreter.datastore.get_current_epoch();
+        if let Err(message) = post_conditions.validate_for_epoch(current_epoch) {
+            return Err(vec![Diagnostic {
+                level: Level::Error,
+                message,
+                spans: vec![],
+                suggestion: None,
+            }]);
+        }
+
         let sender = self.interpreter.get_tx_sender();
         let snippet = format!("(stx-transfer? u{amount} tx-sender '{recipient})");
 
@@ -734,7 +747,7 @@ impl Session {
         // which is one, so the terms are named here.
         let terms = TransactionTerms {
             charge: NonceCharge::Sender(sender.into()),
-            post_conditions,
+            post_conditions: PostConditionCheck::Unchecked,
         };
         self.eval_with_inclusion(snippet, false, terms)
             .map_err(Vec::from)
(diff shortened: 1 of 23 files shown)


components/clarity-repl/src/repl/session.rs
```diff
@@ -712,14 +713,43 @@ impl Session {
         &mut self,
         amount: u64,
         recipient: &str,
+        post_conditions: PostConditionCheck,
     ) -> Result<AnnotatedExecutionResult, Vec<Diagnostic>> {
+        // stacks-core rejects TokenTransfer payloads with post-conditions
+        // before execution. Keep simnet aligned with mainnet/testnet here.
```
brady-stacks, line 719: What does this mean, actually? That token transfer doesn't have postconditions? Do we have to provide a post_conditions arg here if it just fails with any non-empty list?

jbencin-stacks: Yeah `TokenTransfer` is just an STX transfer, which can only do one thing, so there's no point in having additional postconditions on it. Actually, if you submit a `TokenTransfer` with post-conditions, it will be rejected by a Stacks node (but I think an empty post-condition list is fine)

I told Claude to match the behavior on mainnet and add tests for it, and had both Claude and Codex repeatedly go over and make sure the behavior matched mainnet. You can double check that it's correct if you want

jbencin-stacks: Also to preserve backwards compatibility we don't worry about post-conditions if `postConditions` is not present (different from empty list)

brady-stacks (approved):
LGTM

PR #2520, opened 2026-09-03
fix(clarity-vscode): package browser wasm, stop emitting async chunks

hugo-stacks:
### Description

Fix the web extension (vscode.dev / github.dev) shipping without its language server: production tree-shaking dropped the only reference to lsp-browser_bg.wasm, so it has been missing from every published .vsix since 3.14.0.

Emit the wasm with an explicit copy step, disable async chunks everywhere since .vscodeignore is an exact-file allowlist, and remove the dev config's redundant plugin overrides.

Reproduce the bug here: https://vscode.dev/github/stx-labs/clarity-starter

### DIFF
--- a/components/clarity-vscode/rspack.config.dev.js
+++ b/components/clarity-vscode/rspack.config.dev.js
@@ -1,31 +1,15 @@
 // @ts-check
-"use-strict";
+"use strict";
 
-const path = require("path");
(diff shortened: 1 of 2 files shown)


brady-stacks (approved):
👍

PR #2522, opened 2026-09-03
fix(clarity-vscode): load the browser LSP wasm from extensionUri

hugo-stacks:
### Description

The browser language server fetched its wasm from a build-time URL hardcoded to `stackslabs.vscode-unpkg.net`, which is the wrong host on Open VSX (VSCodium web, Gitpod, Theia) and since the release workflow ships the same vsix to both registries, the web LSP could never start there.

`clientBrowser.ts` now derives the URL from `context.extensionUri` and posts it to the worker, which consumes it before opening the LSP connection; relative resolution isn't an option because VS Code starts that worker from a blob URL on an isolated origin. Also scopes #2520's `parser.url` to the wasm-bindgen glue rather than the whole server bundle.

### DIFF
--- a/components/clarity-vscode/client/src/clientBrowser.ts
+++ b/components/clarity-vscode/client/src/clientBrowser.ts
@@ -1,6 +1,8 @@
 import { ExtensionContext, Uri } from "vscode";
 import { LanguageClient } from "vscode-languageclient/browser";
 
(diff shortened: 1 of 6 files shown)


brady-stacks (approved):
💯

PR #2532, opened 2026-09-10
feat(clarity-vscode): log LSP notification timings

hugo-stacks:
### Description

Log lsp notification when `logRequestsTimings` is true.


hugo-stacks:
### Re: the notification queue finding — real, reproduced, deferred

Copilot's "needs a closer look" on `consumeNotification` is correct, and I want to record the reproduction so it doesn't get lost, but I'm deliberately **not** fixing it in this PR.

**The bug.** [`common.ts:57`](https://github.com/stx-labs/clarinet/blob/1585703c1cbe760170605b17e5bd18e39216517f/components/clarity-vscode/server/src/common.ts#L57) reads `notifications[notifications.length - 1]`, but [`common.ts:68`](https://github.com/stx-labs/clarinet/blob/1585703c1cbe760170605b17e5bd18e39216517f/components/clarity-vscode/server/src/common.ts#L68) pops *after* the await. If a notification arrives while the previous one is in flight, `pop()` removes the newly-queued entry instead of the one just handled, and the `notifications.length > 0` recursion then reprocesses the original.

**Reproduced** by driving the built `server/dist/serverNode.js` over node-ipc (real `initialize`, real `vfs/*` callbacks): send `didSave` on a contract, then `didClose` 5 ms later, i.e. inside the ~100 ms `build_state` window.

```
textDocument/didSave (4c900aae67d148): 102.671ms
textDocument/didSave (c89015a5f488b): 75.059ms
```

Two `didSave` lines with different ids, and no `didClose` line at all — the second notification was dropped and the first was processed twice.

**Impact.** An edit or save landing during a slow build is silently discarded, so diagnostics can go stale, and ~100–240 ms of `build_state` is repeated. While the queue is doubled up, `onRequest`'s `notifications.length > 0` guard also returns `null` for hover and completion.

**Why not here.** It predates this PR — the timing logs are just what made it visible; the two lines above are the symptom. Fixing it means changing the queue's consumption order, which is a behaviour change with its own testing needs (concurrent didOpen/didChange/didSave interleavings) and doesn't belong in a diff that only adds instrumentation.

**Suggested fix**, matching Copilot's: consume FIFO — read index `0` and remove with `shift()` — so each notification is handled exactly once. Worth pairing with moving `logTimings?.()` out of the `finally`, since a throw there currently skips `pop()` and stalls the queue permanently.

Two smaller notes on the current diff, for reviewers:

- Copilot's other suggestion (the request-path `try`/`finally`) **is** fixed here. A request whose params fail serde decoding used to produce no timing line; verified it now does.
- Timings only reach the **Clarity LSP** output channel on the desktop build. `vscode-languageclient` pipes server stdout into the channel for `TransportKind.ipc`, but its browser transport has no equivalent, so on vscode.dev these land in the worker's devtools console. That's equally true of the pre-existing request timings this replaces; `connection.console.log` would fix both, as a follow-up.

jbencin-stacks (approved):
I'm approving since the issue Copilot found isn't new to this PR, but make sure to create a GitHub issue for it

PR #2536, opened 2026-09-11
fix(linter): suppress unnecessary tuple warnings in print

### LINKED ISSUE #2462: linter: Surpress `unnecessary_tuple` warnings inside of `print` function
`print` is how we communicate off-chain. The structure of the data we print is determined, at least in part, by protocols and by off-chain consumers, so generating a warning in the Clarity code is not helpful

Jainakin:
### Description

Closes #2462.

`print` payloads are an off-chain interface, and their tuple structure can be prescribed by a protocol or consumer. The `unnecessary_tuple` lint previously descended into those payloads and recommended shape changes that could break that interface.

This change keeps traversing the printed expression but suppresses tuple diagnostics while inside the full `print` payload. The prior traversal state is restored afterward, including across nested `print` calls, so equivalent tuples outside the payload continue to warn.

#### Breaking change?

No. This only removes an inapplicable lint warning inside `print`; runtime behavior and warnings outside `print` are unchanged.

### Example

This no longer emits `unnecessary_tuple` warnings:

```clarity
(print { event: { id: u1 } })
```

The same single-field tuple outside `print` still emits the warning.

### Verification

- `cargo test -p clarity-repl analysis::lints::unnecessary_tuple::tests --locked` — 30 passed
- `cargo test -p clarity-repl --locked` — 660 library tests passed, 2 ignored; integration and doc tests passed
- `RUSTFLAGS="-Dwarnings" cargo clippy --workspace --tests --exclude clarinet-sdk-wasm --locked`
- `cargo fmt-stacks --check`

Regression coverage includes direct tuple payloads, nested tuples, nested `print` calls, sibling tuples after a print, and tuples enclosing a print.

---

### Checklist

- [x] Tests added in this PR

### DIFF
--- a/components/clarity-repl/src/analysis/lints/unnecessary_tuple.rs
+++ b/components/clarity-repl/src/analysis/lints/unnecessary_tuple.rs
@@ -27,6 +27,10 @@ pub struct UnnecessaryTuple<'a> {
     /// Pointers to `SymbolicExpression` nodes that are direct arguments to
     /// `merge` — single-field tuples are required there, so we skip warnings.
     merge_args: HashSet<*const SymbolicExpression>,
+    /// Whether the current expression is part of a value passed to `print`.
+    /// Printed values are an off-chain interface whose tuple shape may be
+    /// prescribed by a protocol or consumer.
+    inside_print: bool,
 }
 
 impl<'a> UnnecessaryTuple<'a> {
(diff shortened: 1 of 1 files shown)


jbencin-stacks (approved):
Nice work @Jainakin! Thanks for the contribution!

PR #2541, opened 2026-09-14
Perf/lsp didsave session cache

hugo-stacks:
## Speed up `textDocument/didSave`

Every save rebuilt the LSP's analysis state from scratch, including the parts that cannot change between saves. On a 53-contract project this cut `didSave` from **332.5 ms to 234.5 ms in wasm (~29%)**, measured end-to-end through the extension's real `LspVscodeBridge`.

### The problem

`build_state` runs on every save and, for each check environment, booted a brand-new `Session`: genesis accounts, first chain tip, and the epoch's boot contracts. None of that depends on contract source — it is identical save over save — yet it was redone each time. Booting also deploys ~36 MB of boot contract sources, ASTs and analyses, and `Session` is `Clone`, so every clone deep-copied all of it.

### The approach

- **Cache the booted session** (`BaseSessionCache` in `clarinet-deployments`). The base state is keyed on exactly what produces it — session settings, genesis spec, first batch epoch — so a key match is a sound reuse rather than a guess. Contract execution then resumes from the cached base.
- **Make the pairing unrepresentable.** Resumption skips the first batch's tip and epoch advance because `prepare` already performed it, so there are two ways to get it wrong: an unprepared session leaves the plan with no boot contracts, and a session prepared for a *different* plan runs this one at the other's epoch and genesis state. Both are silent, since nothing about a `Session` says which plan prepared it. `BaseSessionCache::run_deployment_plan` is the only way in — it names the deployment once and drives both halves itself — so neither mismatch can be written. `PreparedSession` and both halves are private.
- **Share the boot set** behind an `Rc` instead of deep-copying per clone. `update_epoch` forks via `Rc::make_mut`, guarded so a no-op transition doesn't fork the map away from sessions sharing it.
- **Carve out remote-data sessions.** `Session::new` resolves `initial_height: None` against the live chain tip and seeds its datastore from the network, and neither input reaches the key — the setting still reads `None` however the tip resolved. A hit would pin whichever tip the first save saw for the rest of the editor session, so these bypass the cache and drop any entry a prior local build left behind. The cost is real: remote-data sessions do amortize work, because `Session::new` skips boot contracts for them but `prepare`'s `update_epoch` installs them anyway (`deploy_boot_contracts_for_range` has no remote guard). Those projects keep the old per-save cost; a live tip is worth more than the save.
- **Bound the cache** to the 2 most recently used manifests (LRU). Nothing else would ever drop an entry — `clear_protocol` runs on every save so it can't double as the eviction hook — and each retained manifest pins its own boot set in a wasm heap that never returns memory.
- **Tighten the `#[env(...)]` pre-filter** so it matches the annotation's shape rather than the substrings `#[` and `env` anywhere in the file. The skip must stay an over-approximation: a source it rejects is never parsed, so it never reports a parse failure, and `remove_env_simnet` is what keeps simnet-only code off an on-chain network.

### Consequence worth flagging

`EditorState` now holds `Session`s, and `Session` holds `Rc`s, so **`EditorState` is no longer `Send` or `Sync`**. Nothing crossed threads before or after — the native bridge still compiles unchanged — but the wasm bridge's `Arc<RwLock<EditorState>>` now trips `clippy::arc_with_non_send_sync`, silenced with a scoped `#[allow]` and a note that wasm is single-threaded. If the native LSP ever needs to move editor state across threads, this is the constraint it will hit.

### Tests

- The base session cache is invalidated by a changed genesis, a changed first-batch epoch, and matches a fresh build on a hit.
- Repeated LSP saves reuse the cached base, asserted on the hit count — a dropped entry is rebuilt within the same save and is otherwise indistinguishable.
- A three-batch plan advances the chain tip exactly once per batch past the first, and a reused base restarts where it was cached rather than where the previous run ended. One batch cannot tell a correct skip from a missing advance.
- `update_epoch` forks the shared boot set for a real epoch transition and leaves it shared for a no-op one.
- Two plans pinning different epochs, run through one cache, each execute at their own — the corruption a mismatched pairing would cause, now guarded by the API shape rather than by the caller getting it right.
- Two saves against a remote-data API each re-resolve the tip, asserted on the mock's request count rather than on the cache alone — reusing the session is exactly what would stop the second request — and the primed local entry is evicted rather than left pinning its boot set.

### DIFF
--- a/components/clarinet-deployments/src/lib.rs
+++ b/components/clarinet-deployments/src/lib.rs
@@ -6,6 +6,7 @@ pub mod onchain;
 pub mod requirements;
 pub mod types;
 
+use std::fmt;
 use std::path::{Path, PathBuf};
 
 use clarinet_defaults::DEFAULT_EPOCH;
@@ -119,8 +120,8 @@ pub fn setup_session_with_deployment(
     }
 }
 
-pub fn initiate_session_from_manifest(manifest: &ProjectManifest) -> Session {
-    let settings = SessionSettings {
+pub fn session_settings_from_manifest(manifest: &ProjectManifest) -> SessionSettings {
+    SessionSettings {
         repl_settings: manifest.repl_settings.clone(),
         disk_cache_enabled: true,
         cache_location: Some(manifest.project.cache_location.clone()),
@@ -131,8 +132,11 @@ pub fn initiate_session_from_manifest(manifest: &ProjectManifest) -> Session {
             .map(|(name, entry)| (name.clone(), entry.source.clone()))
             .collect(),
         ..Default::default()
-    };
-    Session::new(settings)
+    }
+}
+
+pub fn initiate_session_from_manifest(manifest: &ProjectManifest) -> Session {
+    Session::new(session_settings_from_manifest(manifest))
 }
 
 fn update_session_with_genesis_accounts(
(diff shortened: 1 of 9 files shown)


components/clarinet-deployments/src/lib.rs
```diff
@@ -1063,7 +1186,9 @@ pub async fn generate_default_deployment_with_cache(
         contracts.insert(contract_id, contract_spec);
     }
 
-    let session = Session::new(settings);
+    // Only ever used for `interpreter.build_ast`, which parses against `&mut ()`
+    // and reads no datastore state, so the boot contract set would be dead weight.
+    let session = Session::new_without_boot_contracts(settings);
```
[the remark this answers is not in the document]
hugo-stacks: This one doesn't hold — the console never loads this session.

`Terminal::load(artifacts.session)` gets its artifacts from `load_deployment_and_artifacts_or_exit`, not from `generate_default_deployment_with_cache`. All three of that function's success paths return artifacts built by `setup_session_with_deployment` (`cli.rs:1748`, `:1766`, `:1794`), which calls `initiate_session_from_manifest` -> `Session::new` — boot contracts included. The session from the generate path is only ever read for `asts`/`diags`/`success` and then dropped; I checked every consumer of `artifacts.session` across the CLI, DAP, SDK and LSP and none of them evaluate against it. The DAP does the same thing explicitly at `dap.rs:37` and `dap.rs:116`.

The only path that returns the parse-only artifacts is the `!ast_artifacts.success` branch at `cli.rs:1781`, and the console exits on failed diagnostics before reaching `Terminal::load`.

The underlying readability concern is fair, though: `DeploymentGenerationArtifacts.session` means different things depending on which constructor produced it, and nothing said so. Adding a doc comment on the field in a follow-up.

components/clarity-repl/src/repl/session.rs
```diff
@@ -248,7 +249,9 @@ pub enum CallKind {
 #[derive(Clone)]
 pub struct Session {
     pub settings: SessionSettings,
-    pub boot_contracts: ExecutionResultMap,
+    /// Shared, not owned: ~36 MB of sources, ASTs and analyses that only
+    /// `update_epoch` mutates, against a clone per LSP save.
+    pub boot_contracts: Rc<ExecutionResultMap>,
```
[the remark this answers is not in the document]
hugo-stacks: Breaking is accurate, and it's the intended trade rather than an oversight — sharing the boot set is the whole point of the change. `Session` is `Clone` and the LSP clones one per save; the map holds ~36 MB of boot contract sources, ASTs and analyses that only `update_epoch` ever mutates, so the deep copy was pure waste.

On the migration ask: every in-repo use is a read and deref-coerces unchanged (`core.rs:532`, `:762`, `:1397`, plus the tests). The one mutation site is `update_epoch`, which forks via `Rc::make_mut` guarded so a no-op epoch transition doesn't fork the map away from the sessions sharing it.

The suggested alternative doesn't actually avoid the break — swapping a public field for an accessor is itself source-breaking for the same callers, so it trades one break for another. And `clarity-repl`'s last crates.io release is 2.0.0 from Nov 2023 against a workspace at 3.24.0, so there's no external consumer tracking this API to migrate.

components/clarity-lsp/src/common/backend.rs
```diff
@@ -53,17 +53,24 @@ where
     // `mem::take` would be cheaper, but then a `build_state` error would
     // require manually restoring the cache. We eat the clone for safety.
     let cached_asts = Some(editor_state.try_read(|es| es.ast_cache.clone())?);
+    // Moved out, not cloned like the AST cache — a `Session` is too expensive —
+    // and put back below on every path.
+    let mut base_sessions = editor_state.try_write(|es| std::mem::take(&mut es.base_sessions))?;
```
jbencin-stacks: I remember explicitly not using `std::mem::take` and reverting to `clone()` due to problems restoring it in every possible error path.

Caching in general is hard and has a lot of edge cases. I'll have to dig into this more to make sure it's safe

hugo-stacks: My understanding is that for the AST cache the commit happens at the end of `build_and_commit`, after the Err early return, so a `mem::take` there leaks on the error path.

`base_sessions` is different because the put back is above the early return, and `build_state(...).await` doesn't have the `?` so nothing returns between the take and the restore.

I can add a unit test and improve the comment here.

What do you think?

jbencin-stacks: In theory, assigning `es.base_sessions` on line 71 could clobber the existing value, if it was written to from another function while this one was awaiting `build_state()`

Claude thinks this can't currently happen in practice, but we may want to tighten it up to be safe against future features/refactors. I will continue looking this over, see if I can find any further issues or make a reccomendation

hugo-stacks: the added test should do it

components/clarity-lsp/src/common/backend.rs
```diff
@@ -53,22 +53,32 @@ where
...
-    {
+    .await;
+
+    let new_cache_entries = match built {
         Ok(entries) => entries,
         Err(e) => return Ok(LspNotificationResponse::error(&e)),
     };
 
     editor_state.try_write(|es| {
+        es.base_sessions = base_sessions;
```
jbencin-stacks: This has the same potential problem, and can still clobber cache, only now it's only in the non-error case.

Codex tells me this won't happen in practice, that this function shouldn't be called multiple times concurrently, but that's a protocol-level detail, and there's nothing in the code stopping it.

Since this shouldn't really happen, I'd settle for a `TODO` or `FIXME` comment documenting it, but I think the proper fix would still be an async mutex around `base_sessions` (which would allow `mem::take` also).

This is I'll approve either way

components/clarity-lsp/src/common/backend.rs
```diff
@@ -1481,6 +1500,108 @@ mod lsp_tests {
         );
     }
 
+    /// `ContractSaved` clears protocol state and `build_and_commit` moves
+    /// `base_sessions` out and back — either can drop the cache unnoticed, since
```
jbencin-stacks: This comment doesn't match the latest changes

hugo-stacks: fixed

components/clarity-lsp/src/common/backend.rs
hugo-stacks: Fixed; `build_state` now takes `&mut BaseSessionCache` and the commit goes into its own slot `es.base_sessions.commit(&manifest_location, base_session)`. It can't name another manifest's entry, so there's nothing to serialize and no protocol-level assumption left in the code.
It should be cheaper too

jbencin-stacks (approved):

PR #2544, opened 2026-09-15
perf(clarity-vscode): coalesce queued didChange notifications per document

hugo-stacks:
### Description

Follow up to #2534. A perf change, not a correctness fix — which is why it was left out of that PR.

Document sync is `FULL`, so every `textDocument/didChange` carries the whole document. A burst of keystrokes while the bridge is busy (`build_state` takes ~100-240ms) queues N full snapshots and re-analyzes all N, even though only the last one is observable.

A new `didChange` now overwrites an already-queued `didChange` for the same document instead of appending. It never touches entry 0 (in flight), never merges across a same-document `didOpen`/`didSave`/`didClose`, and skips entries for other documents.

Dropping the intermediate snapshots is safe: the Rust side never reads `version`, and `ContractChanged` only replaces the source — it publishes no diagnostics.

#### Breaking change?

No.

### Example

Typing 4 characters while `build_state` is busy: 4 analyses before, 2 now.

---

### Checklist

Six cases in `server/tests/notifications.test.ts`. Each guard was checked by sabotaging the implementation — disabling the merge, merging into the in-flight entry, and scanning forwards each fail 3-4 tests.


hugo-stacks:
draft for now, re-assess after #2541

PR #2547, opened 2026-09-16
fix: restore the pox-5 testnet admin rewrite

hugo-stacks:
## Problem

Since #2510, simnet has booted pox-5 with a **mainnet-version** admin principal. `make_pox_5_testnet` is supposed to rewrite that literal to its testnet twin; it silently stopped doing so, and the simnet pox-5 body has been byte-identical to the mainnet one ever since

The visible effect is on the protocol-bond surface. `setup-bond`, `set-bond-admin` and the pause controls all gate on `(is-eq contract-caller (var-get bond-admin))`, and `bond-admin` holds `SP72DMR3MJKS7RVBY33JVV7EEJSQ1PYDVKDP10FX` — a mainnet-encoded address that no simnet account matches. Every ordinary call returns `ERR_UNAUTHORIZED`.

It was not strictly unrecoverable. Simnet's only validation of `sender` is that it parses as a standard principal (`components/clarinet-sdk-wasm/src/core.rs:895`), which a mainnet-version address does — so a test that passed the *mainnet* admin as `sender` would have satisfied the guard. Nobody arrives at that except by reading the boot source, and it leaves a mainnet principal sitting in testnet state, but it means the surface was awkward rather than sealed.

## Cause

#2510 moved `bond-admin` and `pause-admin` to a new address and left the Rust constant behind:

```diff
-(define-data-var bond-admin principal 'SP000000000000000000002Q6VF78)
+(define-data-var bond-admin principal 'SP72DMR3MJKS7RVBY33JVV7EEJSQ1PYDVKDP10FX)
```

`str::replace` on a needle that is no longer present is a silent no-op.

Only simnet is affected. Real networks run their own on-chain pox-5, not this boot body. Devnet does use it, but the orchestrator additionally writes `pox_5_bond_admin = "<deployer>"` into the node config (`components/stacks-network/src/orchestrator.rs:956`), pointing the role at an account the user actually holds a key for. Simnet has no equivalent injection — the boot-body rewrite is all it has, which is why the no-op surfaces there and nowhere else.

## Fix

Repoint the constant, and **derive** the testnet address by re-encoding the hash160 with the testnet version byte rather than hardcoding a second literal, so the pair cannot drift apart again.

To be precise about what this restores: `ST72DMR3MJKS7RVBY33JVV7EEJSQ1PYDVHE5T2XC` is not a key anyone holds either, and this does not hand the admin role to a simnet wallet. What it fixes is consistency — a testnet contract whose principals are testnet-encoded, which is what `pox-5.clar`'s own comments say should happen, and what the rewrite did before #2510 (burn address → burn address, equally unkeyed). Tests still have to call as the admin explicitly; they can now do so with a principal that is well-formed for the network it is deployed on.

> [!WARNING]
> The restored admin is `ST72DMR3MJKS7RVBY33JVV7EEJSQ1PYDVHE5T2XC`, **not** the old `ST000000000000000000002AMW42H` — same hash160, different checksum, so it is not a prefix swap. Downstream suites pinning the old address will break; read it off `getContractSource` instead of hardcoding it.


components/clarity-repl/src/repl/boot/mod.rs
```diff
@@ -113,16 +113,38 @@ static BOOT_CODE_POX_3_TESTNET: LazyLock<String> =
...
+/// `stacks_common::address` is not in this crate's dependency graph.
+fn to_testnet_address(mainnet_address: &str) -> String {
+    let (_, bytes) = PrincipalData::parse_standard_principal(mainnet_address)
+        .expect("a hardcoded boot principal must parse")
+        .destruct();
+
+    StandardPrincipalData::new(BOOT_TESTNET_PRINCIPAL.version(), bytes)
+        .expect("the testnet version byte is in range")
+        .to_address()
+}
```
jbencin-stacks, line 145: This doesn't work for arbitrary addresses. If `BOOT_TESTNET_PRINCIPAL` is singlesig, like it is now, it will only work for singlesig addresses.

Contrary to the comment above, we can use `stacks-common` since it's already a transitive dependency through `clarity`. We can rewrite this as:

```suggestion
  fn to_testnet_address(mainnet_address: &str) -> String {
      let (version, bytes) =
          PrincipalData::parse_standard_principal(mainnet_address)
              .expect("hardcoded mainnet principal must parse")
              .destruct();

      let testnet_version = match version {
          C32_ADDRESS_VERSION_MAINNET_SINGLESIG => {
              C32_ADDRESS_VERSION_TESTNET_SINGLESIG
          }
          C32_ADDRESS_VERSION_MAINNET_MULTISIG => {
              C32_ADDRESS_VERSION_TESTNET_MULTISIG
          }
          _ => panic!("expected a mainnet address"),
      };

      StandardPrincipalData::new(testnet_version, bytes)
          .expect("testnet version byte is in range")
          .to_address()
  }
```

So it handles all cases

hugo-stacks: good catch, fixed

jbencin-stacks (approved):

PR #2549, opened 2026-09-17
chore: harness-neutral agent tooling for review and simplify passes

hugo-stacks:
### Description

Adds agent tooling under `.agents/`, kept independent of any one coding agent. No product code changes.

- **`clarinet-gates.sh`** — resolves a review target (branch, commit, range, PR, uncommitted work) into the diff, works out which build surfaces it reaches, and prints only the CI gates that apply.
- **`clarinet-peer-review.sh`** — adversarial review by a *different* model vendor than the host: Codex when run from Claude Code, Claude when run from Codex. The repo's agent instructions are suppressed for the peer, so it is not primed by the same context. Non-blocking by contract — a missing CLI, login or timeout logs a reason and writes nothing.
- **Three skills** — `clarinet-review` (report-only), `clarinet-simplify` (applies fixes), `rust-conventions` (reference).
- **`AGENTS.md`** is now the instructions file, with `CLAUDE.md` importing it via `@AGENTS.md`. Codex previously read none of this repo's instructions.
- **Rust conventions** move out of the instructions file into `rust-conventions`, which both harnesses read. `.claude/settings.json` adds a `PreToolUse` hook that reminds on `.rs` edits.

`.agents/skills/` is Codex's own convention, so it discovers these with no config; `.claude/skills` is a symlink to it for Claude Code.

#### Breaking change?

No.

### Example

```bash
.agents/scripts/clarinet-gates.sh 4bb1877d
```

```
SURFACES:
  native      — clarinet-cli and the native crates
  wasm-sdk    — crate is in the clarinet-sdk-wasm graph
  wasm-lsp    — crate is in the clarity-lsp graph
  vscode      — VSCode extension surface

GATES:
# Run in order, cheapest first. Stop at the first failure.
cargo fmt-stacks --check
cargo clippy --workspace --tests --exclude clarinet-sdk-wasm
cargo clippy --package clarinet-sdk-wasm --target wasm32-unknown-unknown
# stdout macros are silently discarded on wasm32 — see lints/wasm/clippy.toml
CLIPPY_CONF_DIR=$PWD/lints/wasm cargo clippy --no-default-features --lib --target wasm32-unknown-unknown -p clarinet-sdk-wasm -- -A clippy::all -D clippy::disallowed_macros
cargo clippy --target=wasm32-unknown-unknown --no-default-features --package=clarity-lsp
cargo tst
# slow (wasm-pack build), and CI runs it for the whole sdk-wasm graph
pnpm run build:sdk-wasm && pnpm --filter ./components/clarinet-sdk/node run test
wasm-pack test --node components/clarinet-sdk-wasm
# components/clarity-vscode is its own pnpm workspace
pnpm --dir components/clarity-vscode run lint
pnpm --dir components/clarity-vscode run test:server
```

### How to use it

Both skills take a branch, a commit, a PR number, or nothing (defaults to uncommitted work):

```
/clarinet-review 4bb1877d
/clarinet-simplify
```

`clarinet-review` reports and stops; `clarinet-simplify` applies fixes to the working tree. Same names work in Codex, which finds them in `.agents/skills/` with no setup.

The gate resolver is useful on its own, before pushing:

```bash
.agents/scripts/clarinet-gates.sh          # what CI will run for your current changes
```

**Peer review is optional.** It needs the *other* vendor's CLI installed and logged in — `codex` if you are on Claude Code, `claude` if you are on Codex. Without it the review still runs and simply reports that no peer was available; the script never fails its caller.

### What the gate list does and does not promise

CI's workflows filter on nothing but `**/CHANGELOG.md`, so **every job runs on every PR**. The list above is selected by path to save time locally, which makes the implication one-directional: a gate that fails here fails in CI, but a clean local run is not a prediction of a green CI. Both skills say so, and `clarinet-simplify` says it loudest, since it is the one that has already written the changes the claim would be covering.

Within that, the list is meant to be a faithful subset. A self-review of this PR found four places it was not, all fixed here: `wasm-pack test --node` was missing entirely, so nothing ran `clarinet-sdk-wasm`'s tests; a change to a shared crate like `clarity-repl` never pulled in the SDK gates CI runs against it; the manifest-schema flag missed the `clarity-repl` types that are in that schema; and the VSCode surface linted without running the server tests.

`.agents/scripts/*.sh` now selects a gate of its own — `shellcheck` plus the peer script's `--dry-run` — because no CI job lints these, and they decide which gates everything else runs.


.claude/skills/clarinet-review
```diff
@@ -0,0 +1 @@
+../../.agents/skills/clarinet-review
```
jbencin-stacks: Why symlink the individual files instead of the directory?

AGENTS.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Key Dependencies
+
+Clarity interpreter and Stacks libraries come from `stacks-network/stacks-core` git dependency (see `Cargo.toml` for current revision).
+
+## Conventions
+
+- Follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) for commit messages
+- PRs merge via "squash and merge"
+- Rust style, the helpers to reuse, and the wasm32 platform rules live in
```
jbencin-stacks: Do you have to tell it this?

hugo-stacks: Hmm, Claude initially added a 4 lines paragraph about it, that i removed. Looks like it added it instead, I missed it (part of the last minute changes I mentioned). I'll double check what happens when we remvove it

I'll also want remove the whole "Conventions" section in a follow up PR, and add a skill for Git conventions (similar to the rust-conventions one)

.claude/skills/clarinet-review
hugo-stacks: Tested and you're right, that works as well, will do

AGENTS.md
hugo-stacks: removed, the skill is still triggered without it

.agents/skills/clarinet-review/references/clarinet-risk-map.md
```diff
@@ -0,0 +1,45 @@
+# Clarinet risk map
+
+**Only what survives a green CI.**
+
+This repo's CI gates compile every surface, run the tests, and lint the banned stdout macros. Whatever they catch is theirs to report — don't re-hunt it by hand, and don't list it here. This file exists for defects that a clean gate run still lets through.
```
jbencin-stacks, line 5: This line seems kinda repetitive. Maybe give this PR a pass with Codex to clean up some of Claude's extra words

hugo-stacks: @jbencin-stacks This PR, like many others, was reviewed by Claude and Codex

.agents/scripts/clarinet-gates.sh
```diff
@@ -0,0 +1,345 @@
...
+  echo "pnpm --dir components/clarity-vscode run test:server"
+fi
+if $rust_deps; then
+  echo "cargo audit"
+fi
+if $js_deps; then
+  echo "# pnpm-workspace.yaml sets minimumReleaseAge: any new npm dep must be >= 5 days old"
+  # A manifest edited without regenerating the lockfile fails CI here and
+  # nowhere else, so the comment above needed a command under it.
+  echo "pnpm install --frozen-lockfile"
```
jbencin-stacks: We may want to be more specific here:

```bash
pnpm --dir components/clarity-vscode install --frozen-lockfile
```

.agents/scripts/clarinet-gates.sh
```diff
@@ -0,0 +1,345 @@
...
+rust=false wasm=false lsp=false sdk_ts=false vscode=false manifest=false
+rust_deps=false js_deps=false agent_sh=false
+
+while IFS= read -r f; do
+  [[ -n $f ]] || continue
+  case $f in
+    # .clar boot sources are include_str!'d into clarity-repl, and a manifest
+    # or lockfile change can break the build on its own. Both need the compile
+    # and test gates, not just the audit.
+    Cargo.toml | Cargo.lock | components/*/Cargo.toml)
```
jbencin-stacks, line 202: Changes to the root Cargo files can affect everything, so we should have something like:

```bash
  Cargo.toml | Cargo.lock)
    rust=true
    rust_deps=true
    wasm=true
    lsp=true
    ;;
```

.agents/README.md
```diff
@@ -0,0 +1,137 @@
...
+
+`AGENTS.md` deliberately does **not** point at it. Both harnesses load it from
+its `description` alone: given only a Rust editing task, each one invoked it
+unprompted and applied rules that exist nowhere but the skill file. Claude Code
+has two further routes — that description, plus the `PreToolUse` hook on `.rs`
+edits in `.claude/settings.json` — and Codex needs no `allow_implicit_invocation`
+to do the same. A pointer in `AGENTS.md` would cost every session the tokens and
+earn nothing.
+
+## Why `.agents/` and not `.claude/`
```
jbencin-stacks, line 30: Do we need this section? It seems kind of obvious

.agents/skills/clarinet-review/references/clarinet-risk-map.md
jbencin-stacks: I've also reviewed it with both, and neither one said anything

While Codex doesn't have the same tendency towards excess verbosity as Claude, it doesn't complain about or change it unless you specifically ask it to

.agents/scripts/clarinet-gates.sh
hugo-stacks: good catch

hugo-stacks: Fixed in 959fda52 by resolving the workspace from the changed file: it walks up to
the nearest `pnpm-workspace.yaml` and emits one install per workspace the diff
actually touches.

PR #2551, opened 2026-09-18
refactor: drop nestable_block_on in favor of explicit tokio runtimes

hugo-stacks:
## Problem

`hiro_system_kit::nestable_block_on` promised to block on a future whether or not a runtime was already present. It could not do that.

```rust
pub fn nestable_block_on<F: Future>(future: F) -> F::Output {
    let (handle, _rt) = match tokio::runtime::Handle::try_current() {
        Ok(h) => (h, None),
        Err(_) => {
            let rt = tokio::runtime::Runtime::new().unwrap();
            (rt.handle().clone(), Some(rt))
        }
    };
    handle.block_on(future)
}
```

`Handle::block_on` panics when called from an async context, so the `Ok(h)` branch — the entire point of the name — would abort with *"Cannot block the current thread from within a runtime"* if it were ever reached.

It was never reached. All 10 call sites ran either on `fn main` or on a freshly spawned OS thread, where `Handle::try_current()` always errs. Every call therefore fell through to `Runtime::new()`, quietly building a **multi-threaded** runtime (one worker thread per core) that no call site had asked for or needed — none of them spawn a task.

The sharpest edge was the bitcoin mining runloop, which built and tore down that runtime on *every mined block* for the life of a devnet session.

## Fix

Delete `nestable_block_on` and make each call site say what runtime it wants.

| Call site | Now |
|---|---|
| `stacks-network/src/main.rs` | `#[tokio::main(flavor = "current_thread")]` |
| Bitcoin mining runloop | one runtime built **outside** the loop |
| 5× LSP bridge thread spawns | one `native_bridge::spawn_language_server()` |
| devnet start, deployment generation, coordinator/orchestrator threads | `create_basic_runtime().block_on(…)` |

`create_basic_runtime` stays. Unlike `nestable_block_on` it does not reimplement a tokio concept — it names a *configuration* (current-thread, io + time, capped blocking pool) shared across eight sites. `frontend/telemetry.rs` had that same builder copy-pasted inline and now calls the helper.

Also removed along the way:

- `stacks_network::block_on` — dead `pub` API, zero callers.
- The vestigial `result` / `is_err()` binding in `run_lsp`; `start_language_server` is no longer `pub`, since `spawn_language_server` is now its only caller.
- `features = ["tokio_helpers"]` on clarity-repl's `hiro-system-kit` dependency. clarity-repl only ever used the `red!`/`yellow!` macros. To be precise about where this lands: it is on the `cfg(not(target_arch = "wasm32"))` entry, and `dap` — a default feature — declares its own `tokio`, so a default native build is unchanged. The edges that actually get lighter are the crates taking clarity-repl with `default-features = false`: `clarity-lsp`, `clarinet-deployments`, `clarinet-files` and `clarity-events` no longer drag `tokio/full` in behind it.

## This is not a speedup

The removed per-block overhead was measured, not estimated — 200 iterations each, 14-core arm64:

| Operation | Per call |
|---|---|
| `Runtime::new()` + drop (multi-thread, 14 workers) | 327 µs |
| `create_basic_runtime()` + drop (current-thread) | 8.9 µs |

At 327 µs per mined block, the overhead was 0.001% of the 30s default block interval and still only 0.03% at an aggressive 1s. **No block-timing or throughput change should be expected at any interval.**

What the change is actually worth is twofold. It stops creating and joining 14 OS threads per block indefinitely — invisible in timing, visible in thread-limit headroom on constrained CI containers. And it closes a silent-failure path: `Runtime::new().unwrap()` panicked inside a thread whose handle is discarded (`let _ = thread_named(…).spawn(…)`), so a spawn failure under resource pressure would have stopped block production with no diagnostic, and in dashboard mode the panic goes to a stderr the TUI has taken over. That window is now startup-only.

## Reviewer note

One line looks inconsistent with the rest of the diff and is deliberate: **`clarinet-cli/src/devnet/start.rs` keeps a multi-threaded runtime.**

Everywhere else the current-thread flavor is safe, because the futures involved spawn no tasks and their HTTP clients are one-shot (`reqwest::blocking` in `stacks-rpc-client`, `reqwest::get` in `requirements.rs`). This site is different: `prepare_local_network()` opens Docker connections on that runtime, and hyper spawns a driver task per pooled connection. The bollard client is then handed to the orchestrator thread — a *different* runtime — while the original thread blocks forever in the dashboard/event loop. On a current-thread runtime those drivers freeze, and the orchestrator's first reuse of a pooled connection would stall until bollard's 120s timeout.

It is pinned to `worker_threads(1)` rather than the `Runtime::new()` default of one-per-core: `stacks-network` contains no `tokio::spawn` at all, so keeping that idle connection pool alive is the only work the runtime has left.

## Out of scope

Two call sites match the pattern this PR removes and are deliberately untouched, to keep the diff reviewable:

- `observer/src/event_handler/mod.rs:170` — `Runtime::new().unwrap()` under a raw `std::thread::Builder`.
- `stacks-network/src/orchestrator.rs:2265` — a hand-rolled equivalent of `create_basic_runtime()`.

Worth a separate look: `do_run_devnet` is an `async fn` with zero `.await` in its body. It spawns OS threads and then blocks its caller's runtime thread forever, which is the underlying reason `devnet/start.rs` needs the special case above. Making it a plain `fn` would remove the constraint rather than document it.

## Testing

`cargo clippy --workspace --exclude clarinet-sdk-wasm --all-targets`, the `clarinet-sdk-wasm` wasm-target clippy job, and `cargo fmt-stacks --check` are clean. 54 tests pass across `clarinet-cli`, `stacks-network` and `hiro-system-kit`, including all five LSP tests — two of which now call `spawn_language_server` from inside a live multi-threaded runtime.

CI cannot boot a devnet, which is the path this PR touches most, so it was run manually against this branch:

- **Boot to completion** — network plus all 7 containers created, contracts deployed, `Local Devnet network ready`. This exercises `prepare_local_network()` on the new `worker_threads(1)` runtime, the risk described above; no stall.
- **Sustained mining** — bitcoin blocks at a steady ~60s across five consecutive iterations, with no drift. This is the runloop whose runtime is now hoisted out of the loop, reused across 60s gaps in which it is never polled.
- **Container log streaming** (`--save-container-logs`) — `start_container_logs_streaming` clones the bollard client onto a *third* runtime, so all three share one connection pool. Streamed 5k+ lines and kept growing; no pool interaction observed.
- **Graceful shutdown** — SIGINT, log stream ended cleanly, all containers removed, process exited.
- **Snapshot boot path** — the default path (no `--from-genesis`) extracts the embedded snapshot, resumes at the snapshot heights, mines and shuts down cleanly.

> [!NOTE]
> One coverage gap remains: the **TUI dashboard** path was not run, only `--no-dashboard`. Both end up in `do_run_devnet`'s blocking tail, but via different branches.


jbencin-stacks (approved):
It's not necessary to address in this PR, but we should consider dropping `hiro_system_kit` and using `tokio` directly. The extra layer doesn't add much, but leads to more confusing and non-standard code

hugo-stacks:
yes that was my initial goal. The remaining `create_basic_runtime` does provide some useful abstraction and was a bigger lift to get rid off. But we could definitely do it if we think it's worth it

PR #2554, opened 2026-09-22
chore: add simplify, review and ship agent skills

hugo-stacks:
### Description

A smaller take on #2549, following the discussion in #labs-devtools: start with a small base and add to it as our needs sharpen.

- **Three skills** in `.agents/skills/`:
  - `clarinet-simplify` applies quality fixes.
  - `clarinet-review` is read-only and reports correctness findings.
  - `clarinet-ship` covers the checks per surface, git conventions and opening the PR.
- **`AGENTS.md`** is the instructions file, and `CLAUDE.md` is a symlink to it. Git conventions moved into `clarinet-ship`.
- **Rust conventions** moved to `.agents/rust-conventions.md`. `AGENTS.md` points to it before editing Rust, simplify applies it, and review reports violations as minor.
- **Harness-neutral.** Codex and Pi discover `.agents/skills/` natively. Claude Code reads it through the `.claude/skills` symlink.
- **Risk map** has only the two findings that passed CI and were caught in review, with the PR as evidence: `::encode` missing the boot remap (#2518) and a hardcoded epoch (#2513).
- **Review output** goes to `.agents/reviews/<PR-n|branch>.md` (gitignored). Each run updates that file, so a review can continue across runs and across harnesses.

Compared with #2549, this leaves out the scripts (gates and peer review) and the automatic cross-model review. Models work out what to run from the diff and the checks listed in `clarinet-ship`.

#### Breaking change?

No.

### Example

```
/clarinet-simplify
/clarinet-review 2549
/clarinet-ship
```

In Codex, use `$clarinet-review`. In Pi, use `/skill:clarinet-review`. All three also trigger from plain requests like "review my branch".

---

### Checklist


.agents/skills/clarinet-ship/SKILL.md
```diff
@@ -0,0 +1,60 @@
...
+pnpm --dir components/clarity-vscode exec playwright install chromium && pnpm --dir components/clarity-vscode test
+```
+
+`components/clarity-vscode` is its own pnpm workspace, so it needs `pnpm --dir components/clarity-vscode install --frozen-lockfile` first.
+
+Report which checks ran and which didn't. A skipped check is not a passing one.
+
+## Git conventions
+
+- Commit messages and PR titles follow [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary): `fix:`, `feat:`, `chore:`, `refactor:`, `docs:`, `test:`, `perf:`.
```
jbencin-stacks, line 53: We don't need to enforce this on every individual commit, just on PR titles and the final squash merge commit

hugo-stacks: Right, I also thought about back when we did the CLAUDE.md and even the root READE.md back then. 
But my assumptions has been, for both external and even now for agents, that it's simpler to always enforce it instead of conditionally doing it?

Unless there are some benefits I don't see in not doing it?

.agents/rust-conventions.md
```diff
@@ -0,0 +1,29 @@
+# Rust conventions
```
jbencin-stacks, line 1: Should this be formatted as a skill and under `.agents/skills`?

.agents/rust-conventions.md
```diff
@@ -0,0 +1,29 @@
...
+- `?` for error propagation instead of manual `match`/`unwrap`
+- Destructuring in function arguments and match arms
+- Meaningful type aliases and newtypes where they aid readability
+- `impl Into<T>` / `AsRef<T>` parameters for flexible function signatures when appropriate
+- `From` / `TryFrom` / `FromStr` impls for conversions, not standalone `to_x` / `parse_x` helpers; implement `From`, not `Into`
+- Derive macros (`Clone`, `Debug`, `Default`, etc.) rather than manual implementations
+- `Display` for rendering a type, not a `to_string`-style helper
+- Inline variables in format strings: `format!("{name}")`, not `format!("{}", name)`
+- `indoc!` / `formatdoc!` for multi-line strings
+- `strum` derives (`EnumString`, `Display`, `EnumIter`) instead of hand-written enum boilerplate
```
jbencin-stacks, line 16: Maybe modern LLMs understand this by default, but the most important thing in Rust IMO is to make full use of the type system:

- Convert between types with by implementing `From`, `Into`, etc. (already covered)
- Never use sentinel values, always `Option`/`Result`
- "Make invalid states unrepresentable", or use types to guarantee correctness

When I did the coding interview for Hiro, this was the major thing to look for to determine how "fluent" a candidate was in Rust

.agents/rust-conventions.md
```diff
@@ -0,0 +1,29 @@
...
+
+Reuse what exists instead of re-deriving it:
+
+- Default epoch and Clarity version: `clarinet_defaults::{DEFAULT_EPOCH, DEFAULT_CLARITY_VERSION}`
+- A boot contract's epoch: `clarity_repl::repl::boot::get_boot_contract_epoch_and_clarity_version`
+- Formatting a Clarity `Value`: `clarity_repl::repl::clarity_values::value_to_string`
```
jbencin-stacks, line 22: Could also add a note about not duplicating work from stacks-core. Pull stuff from there if possible, or suggest a refactor upstream if it's not properly exposed. I've dealt with this a lot lately

.agents/skills/clarinet-review/SKILL.md
```diff
@@ -0,0 +1,35 @@
+---
+name: clarinet-review
+description: Review a change in the Clarinet repo for correctness — a branch, a commit, uncommitted work or a pull request. Use when asked to review changes, review a PR, or check work before marking a PR ready. Reports findings without editing code.
+---
+
+# Review a Clarinet change
+
+Find defects in the change. Read-only: don't edit code unless the user then asks for fixes. Never post to GitHub.
```
jbencin-stacks: I sometimes tell it that it that it can use a temporary branch to confirm findings, but don't make any lasting changes

.agents/skills/clarinet-review/SKILL.md
```diff
@@ -0,0 +1,35 @@
...
+
+- **A rule added at one entry point misses the others.** The same session is reached through the console's `::` commands, `simnet.execute`, `runSnippet`, the DAP, the SDK and deployment plans. When a change adds a rule (remap an id, rewrite a source, validate an argument), list every entry point and check each one. The one that was missed is unchanged code, so it won't be in the diff. Missed in `::encode` (#2518).
+- **A hardcoded epoch goes stale.** Epoch and Clarity version literals change with Stacks releases. Derive them (see `.agents/rust-conventions.md`). A warning with a baked-in epoch (#2513).
+
+## Report
+
+Group the findings: **blocking**, then **minor**, then **resolved** since the last review. For each, give `file:line`, the defect in one sentence, and the input or state that triggers it. If there are no findings, say so.
+
+Write the report to `.agents/reviews/<name>.md` (gitignored), with `<name>` set to `PR-<number>` or the branch name with `/` replaced by `-`. If the file already exists, update it: move fixed findings to resolved and add new ones. That way the next run, in any harness, picks up where this one stopped.
```
jbencin-stacks, line 35: Maybe specify naming conventions for non-PR reviews also

jbencin-stacks (dismissed):

jbencin-stacks:
I left a few comments, address if you think they make sense, or in a follow-up PR, but I think this is good enough to be merge

EDIT: Don't merge just yet, give @brady-stacks a chance to review also

.agents/rust-conventions.md
hugo-stacks: It's externalized in its own file so that it can be referenced from other places like [AGENTS.md](https://github.com/stx-labs/clarinet/pull/2554/changes#diff-a54ff182c7e8acf56acfd6e4b9c3ff41e2c41a31c9b211b2deb9df75d9a478f9R5), [simplify](https://github.com/stx-labs/clarinet/pull/2554/changes#diff-89750b30e54a66ad9efd0e87ba7c84779092ff8051f66c3dcd8fb65f978e6367R20) and [review](https://github.com/stx-labs/clarinet/pull/2554/changes#diff-e98031c0cf4f1626b9cb0864613c6dbfc94c14fc12a707adf7ccc81dfcb6934bR20)

.agents/skills/clarinet-review/SKILL.md
```diff
@@ -5,7 +5,7 @@ description: Review a change in the Clarinet repo for correctness — a branch,
 
 # Review a Clarinet change
 
-Find defects in the change. Read-only: don't edit code unless the user then asks for fixes. Never post to GitHub.
+Find defects in the change. Read-only: don't edit code unless the user then asks for fixes. To confirm a finding, you may write a throwaway test on a temporary branch or worktree, but remove it and leave no lasting changes. Never post to GitHub.
```
jbencin-stacks, line 8: I think it's fine to leave the temporary branch around after the review, as I'll often end up using the tests and making them pass, but we also don't want a bunch of leftover temp branches cluttering up our local repos, so I'm not sure what to do here

